PR #5, opened 2015-11-19
Add dependency on React pod

keeth:
With use_frameworks! enabled in my PodFile, I get a linker error:

Undefined symbols for architecture x86_64:
  "_RCTRegisterModule", referenced from:
      +[RNDeviceInfo load] in RNDeviceInfo.o
ld: symbol(s) not found for architecture x86_64

This is fixed by adding React as a dependency in the podspec


rebeccahughes:
Thanks! I've pushed this in 0.5.1 to npm just now :)

PR #8, opened 2015-11-29
Add user defined deviceName field

donald-jackson:
I needed to get the user defined device name for my project so I made these changes to allow for it.

They have the unfortunate side effect of requiring the Bluetooth permission on Android so I don't think this pull request should necessarily be accepted but its here if someone needs/wants it.


rebeccahughes:
Thanks :) I'll test it to see if it will throw an exception if you don't have the bluetooth permission, if it's caught then it can be included in the project and just warn the user in the console if they don't have the permission.

PR #10, opened 2015-12-20
Set lint warning for InvalidPackage, instead of error

lifuzu:
Reference:
https://github.com/square/okio/issues/58


rebeccahughes:
Thanks :)

Issue #14, opened 2016-01-06
Upgrade deviceLocale on Android


Bhullnatik:
Hi,

First off thanks for your module!
I just noticed that your last version now gets the device locale which is great. I wrote a module to do it myself but would be happy to use yours.

My problem is that [`locale.toString()`](http://developer.android.com/reference/java/util/Locale.html#toString()) is... weird on Android as it can return unpredictable results and even the correct result is different from the iOS one (`en_US` vs `en-US`). [`locale.toLanguageTag()`](http://developer.android.com/reference/java/util/Locale.html#toLanguageTag()) seems more robust but only available on API 21 (Lollipop), but using it with a good (enough) fallback seems the good solution to me.

In reference here is the code I'm using now:
```java
private String getCurrentLanguage() {
    Locale current = getReactApplicationContext().getResources().getConfiguration().locale;
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
        return current.toLanguageTag();
    } else {
        StringBuilder builder = new StringBuilder();
        builder.append(current.getLanguage());
        if (current.getCountry() != null) {
            builder.append("-");
            builder.append(current.getCountry());
        }
        return builder.toString();
    }
}
```

I would be happy to look into that and submit a PR.

rebeccahughes:
Thanks :) that looks like a good solution, if you want to submit a PR that would be great!

PR #15, opened 2016-01-06
Use getLanguageTag to retrieve device locale

Bhullnatik:
As discussed in #14.

I didn't change much code, tell me if there is something more I can do!


rebeccahughes:
Thanks :)

PR #26, opened 2016-02-25
Avoid destructuring to support Node.js

martnu:
Running into an issue when unit testing RN from the CLI. Since [Node doesn't support it yet](https://kangax.github.io/compat-table/es6) it will break. 

```
var { RNDeviceInfo } = require('react-native').NativeModules;
    ^

SyntaxError: Unexpected token {
```

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -2,7 +2,7 @@
  * @providesModule react-native-device-info
  */
 
-var { RNDeviceInfo } = require('react-native').NativeModules;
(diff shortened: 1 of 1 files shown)


martnu:
@rebeccahughes What do you think about this change?

rebeccahughes:
Sorry, i'll push it to npm now

martnu:
Thanks for quick action and a great plugin!

PR #27, opened 2016-03-05
Update device names to respect Apple Branding

marekhrabe:
Apple uses lowercase `s` for all devices since 5s model


lucasfeliciano:
:+1:

PR #30, opened 2016-03-17
Add support for device country

mikefowler:
Provides the ISO country string derived from the application's locale.

@rebeccahughes: wasn't sure if this was the best key name; please advise if you'd prefer something different.


mikefowler:
:pray: :fireworks:

PR #33, opened 2016-04-06
Update RNDeviceInfo.podspec

aforty:
Resolved a hard error (missing authors)
Resolved warning (git repo should end in .git)
Updated version to match package.json


aforty:
By the way, you have to keep the version up-to-date in the podspec file going forward. Otherwise it does no one any good. :+1:

PR #36, opened 2016-04-14
Add build version

chetstone:
Returns full build version on iOS, use with [xcode-git-version.sh](https://gist.github.com/karlvr/c93a98d7000ecb163895) to print complete git info.
(Not sure what it'll do on Android)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,8 @@ console.log("Bundle Id", DeviceInfo.getBundleId());  // e.g. com.learnium.mobile
 
 console.log("Build Number", DeviceInfo.getBuildNumber());  // e.g. 89
 
+console.log("Build Version", DeviceInfo.getBuildVersion());  // on iOS, full build info, e.g. v0.5.9-51-g71ec8a0+
+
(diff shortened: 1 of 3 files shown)


chetstone:
This worked when building for the simulator but when building for a device my git-version script fails with `Set: Entry, ":CFBundleBuildVersion", Does Not Exist`. 

This may be related to the warning:
```
(null): Object file (/Users/chet/prj/malas/firebase/ios/build/Debug-iphoneos/libRNDeviceInfo.a(RNDeviceInfo.o)) was built for newer iOS version (8.0) than being linked (7.0)
```
Oddly, I cannot seem to change the iOS deployment target for RNDeviceInfo.xcodeproj, or rather, I can change it to 7.0 but xcode ignores it and builds 8.0 instead!!

So closing, this needs more work.

chetstone:
Well there wasn't anything wrong with the PR, the script error was due to a ham-fisted typo in the script (which is not part of this PR) , and I managed to change the iOS deployment target by editing the project file. For some reason changing it in the xcode gui wasn't doing it.

However Android is still not supported.

PR #40, opened 2016-04-27
Fix to list creation

LukeAvery92:
Compile time error fix. Array List with no type in template. Error occurred on react-native 0.21.0 when running android version.


LukeAvery92:
I have now discovered this to be an issue with my Java version being 6 and not 7.

PR #41, opened 2016-04-29
Add timezone to information provided by module.

juankiz:


juankiz:
Hey @rebeccahughes any comments on the pr?

GantMan:
@juankiz  - Can you please implement this on Android as well?  I think this library is keen on keeping parity between iOS and Android.

Also, you might be able to do this with a 100% javascript solution right?   
`let offset = new Date().getTimezoneOffset();`

mehcode:
Added a java impl on https://github.com/rebeccahughes/react-native-device-info/pull/59

@GantMan Not possible to get an Olson formatted timezone form javascript.

GantMan:
Closed via #59

PR #42, opened 2016-05-09
get android device id by TelephonyManager

zhuo87:
If there is any question, please commit it. Thank you!


martincik:
There's even smaller solution see here: http://stackoverflow.com/questions/2785485/is-there-a-unique-android-device-id

peteclark3:
Note that this isn't a viable solution for a device id for tablets... (any non-phone android device).. hence implemented this instead: https://github.com/rebeccahughes/react-native-device-info/pull/46

GantMan:
Closed via #47

PR #47, opened 2016-05-25
Adding support for android's Instance ID

peteclark3:
Adding support for android's Instance ID

I apologize, I tried to squash my commits by rebasing, but I'm new to it and clearly did something wrong... ended up adding more commits :-(.  If you can't merge due to this, no worries..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,21 @@ DerivedData
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-ignore-the-pods-directory-in-source-control
 #
 #Pods/
+
+.idea/.name
+
+.idea/compiler.xml
+
+.idea/copyright/profiles_settings.xml
+
+.idea/encodings.xml
+
+.idea/misc.xml
+
+.idea/modules.xml
+
+.idea/react-native-device-info.iml
+
+.idea/vcs.xml
+
+.idea/workspace.xml
--- a/README.md
+++ b/README.md
@@ -125,4 +125,7 @@ console.log("User Agent", DeviceInfo.getUserAgent()); // e.g. Dalvik/2.1.0 (Linu
 console.log("Device Locale", DeviceInfo.getDeviceLocale()); // e.g en-US
 
 console.log("Device Country", DeviceInfo.getDeviceCountry()); // e.g US
+
+console.log("App Instance ID", DeviceInfo.getInstanceID()); // ANDROID ONLY - see https://developers.google.com/instance-id/
+
 ```
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,5 +20,5 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.11.+'
-    compile 'com.google.android.gms:play-services-identity:8.3.0'
+    compile 'com.google.android.gms:play-services-gcm:8.3.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,5 +20,5 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.11.+'
-    compile 'com.google.android.gms:play-services-location:8.3.0'
+    compile 'com.google.android.gms:play-services:8.3.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,4 +20,5 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.11.+'
+    compile 'com.google.android.gms:play-services-location:8.3.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,5 +20,5 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.11.+'
-    compile 'com.google.android.gms:play-services-base:8.3.0'
+    compile 'com.google.android.gms:play-services-identity:8.3.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -19,6 +19,5 @@ android {
 }
 
 dependencies {
-    compile 'com.google.android.gms:play-services-gcm:8.3.0',
     compile 'com.facebook.react:react-native:0.11.+'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -20,5 +20,5 @@ android {
 
 dependencies {
     compile 'com.facebook.react:react-native:0.11.+'
-    compile 'com.google.android.gms:play-services:8.3.0'
+    compile 'com.google.android.gms:play-services-base:8.3.0'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -19,5 +19,6 @@ android {
 }
 
 dependencies {
+    compile 'com.google.android.gms:play-services-gcm:8.3.0',
     compile 'com.facebook.react:react-native:0.11.+'
 }
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -6,6 +6,8 @@ import android.content.pm.PackageManager;
 import android.os.Build;
 import android.provider.Settings.Secure;
 
+import com.google.android.gms.iid.InstanceID;
+
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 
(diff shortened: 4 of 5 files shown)


romansklenar:
👍

GantMan:
I personally would like to see this squashed.  Please open a new branch, and just render this in 1 to 2 commits please.   Anyone tracking through your history will be very sad :(

romansklenar:
Repo maintainer can [squash commits in PR when merging](https://help.github.com/articles/about-pull-request-merge-squashing/).

mehcode:
:-1: This isn't cross implemented (or possible) in iOS and can be replicated cross-platform through very few lines of javascript (generate a UUID on app open and store in async storage).

peteclark3:
Correct.  It isn't possible to get an android instance id on iOS.. you get a gold star. 

I did a bunch of research on this and UUID definitely *is not* the best practice way to identify a device on android.  RTFM: https://developer.android.com/training/articles/user-data-ids.html

@rebeccahughes just either close this PR without merging or merge it so we can put this to bed. 

Happy 4th!

GantMan:
@peteclark3  - I remember this quite clearly from my Android dev days :)

I guess the question at hand was, "What should this do on iOS? if someone leaves it in?"   As of right now it will error out.   In my opinion, that's what it should do.  We can't deny the difference between supported platforms.  This will only grow as React adds UWP, too!

My only suggestion is that you add a friendlier error message on iOS side.  It should still throw, but with a very direct message.

I'm going to squash and merge this monster PR, but please do consider adding that friendly message in a new PR.

Cheers!

mehcode:
@peteclark3 

No need to be defensive. I apologize if I sounded harsh.

I've done my research as well. Even the link you reference (and tell me to RTFM which is quite rude to be honest when talking to a fellow engineer who we can hopefully assume has) states "Use an Instance ID or a privately stored GUID" (which the second bit is what I said) each time it mentions the instance ID. I fail to see the benefit of the instance ID. I'd appreciate your reasoning beyond "it's better". Not because I'm trying to attack you or this merge but because I'm genuinely curious.

---

@GantMan 

At least open an issue ( #60 ) to improve the support so we don't forget. This kind of library ( in my opinion ) should not support something that isn't implementable on all platforms. It should at least make it clear (eg.. `deviceInfo.android.getInstanceId()`).

GantMan:
@mehcode  - I agree it should be clear.
We're going to find issues like this on the corners.   I really like the idea of an android chain, or perhaps adding android to the function name?

We're never going to get away from the devices being a little different, because of course they are.  And not all concepts will be able to straddle them.   I think you're right, let's come up with a standard to make it clear.

Do you think the `android` chained in is better than keeping shape and adding `Android` to the name?   Seems React Native usually puts `IOS` on things it knew to be one platform only.   So following React Native, maybe `getInstanceIdAndroid()` ?

mehcode:
@GantMan Agree :+1:

If we are to have a method that is unique (we don't have one yet) it should be `<name>Android` or `<name>IOS` just like react native (didn't think of that)

peteclark3:
@mehcode 

At that link it states very simply "we recommend: instance id" - going further to only recommend GUID *if instanceid is insufficient* 
 
I pointed you to the docs in hoping you would see that and understand that my reasoning came from Google itself, not just me saying "it's better."

peteclark3:
@mehcode 

Frankly, your initial "feedback" was annoying and came off as basic mud-slinging because:

- You cited no documentation to the contrary of my initial claim (which actually did include documentation) - https://github.com/rebeccahughes/react-native-device-info/pull/46#issue-156856487 ... and gave no indication that you had read *anything* on the topic.. 
- Instead of providing an actual coded alternative, you simply indicated "very few lines of javascript."  I appreciate feedback I can actually *learn from* to provide better solutions.  You cited no documentation, gave no code, offered nothing.  Well, except an emoji.  👍 
- Your comment missed the point of the PR, and, IMO, the module in general.   This module is for **device info** ... i.e. getting information from the device/os.  Not writing javascript for a cross-platform unique id.  There are plenty other libraries that try to do that. **I was adding support for android's instance ID, plain and simple.**  

Per your later comment, I never left it at "mine's better."  I cited Google's android documentation originally, and have continued to.. where they always mention Instance ID first, GUID second, once indicating the usage of an app-generated GUID "only when Instance ID is not sufficient."  And again, I wasn't implementing a GUID in the first place.    

... so, yeah.  To me, your comment read like basic mud-slinging without any actual usable feedback. Re: your seeing no benefit to using instance ID, if Google's docs can't convince you, I'm definitely not going to try ;-) 

Incidentally, if you do get an iOS identifier working that persists across installs *and* doesn't use the IDFA, I'm all about it...

mehcode:
@peteclark3 

To put the record straight... I never meant to be offensive. I don't always have the time to write out a full response. I'm not attacking you and never was.

--- 

I'm not being stubborn, rude, etc. 

I seriously question the purpose of the SDK-provided Instance ID (over a locally generated UUID). The documentation doesn't do much beyond state it exists in the same paragraph it says "or GUID".

I understand _what_ it does. That's not what we're debating. I do not understand the _benefit_ over a locally-generated UUID. And no, google does not state Instance ID is better unless you're reading into how it says "In cases where an Instance ID isn't practical, custom globally unique IDs [...]" (of course it never says why the Instance ID wouldn't be practical or go into detail about why it'd be preferable when it is practical). Google's documentation is downright garbage most of the time. 

I assumed as you were the one writing this PR that you had some "front-line knowledge" (of actual use and why you're using that instead of UUID). 

---

> Not writing javascript for a cross-platform unique id. There are plenty other libraries that try to do that.

I don't think you understand what I meant. I never said "cross-platform unique id". And no libraries exist (that I know of) that do what I did say, which was "cross-platform instance id".

---

> Your comment missed the point of the PR, and, IMO, the module in general. This module is for device info [...]

I'm going to ignore that this is a bit offensive and point out that this module does a disservice if it is not cross-platform across iOS and Android (and now there is a big PR to make it work on UWP as well). Each platform-unique case should needs special attention from maintainers and documentation to point this out (so someone on a team that's making a marketable app [works on iOS and Android] and wonders why the app is dead until they come across `deviceInfo.getInstanceID` and scream at their team member for using android-only code).

---

> Instead of providing an actual coded alternative [...]

Now this is what I meant about time constraints. I work a lot (probably far too much) as a CTO of a software company. I love contributing on github and poking around (as you can see from my profile). I just don't always have the time to go into detail (I've submitted far too many PRs or Issues with no description at all) and hope the person on the other end gets the idea.

Here is your code, by the way (really just a few lines):

```js
  async getInstanceID() {
    // Check for an existing ID
    let id = await AsyncStorage.getItem("instance-id");
    if (id != null) return id;

    // Generate a new UUID (react-native-uuid)
    id = uuid.v4();

    // Save the ID
    await AsyncStorage.setItem("instance-id", id);

    return id;
  }
```

peteclark3:
@mehcode 

We're fellow CTOs then, who both work too much.  We have that in common for sure.  

Frankly I was frustrated by your initial comment b/c I was trying to do the right thing and not just fork/modify/use this repo without contributing something back.  I learned early on in college from a dear professor of mine ('round 1995, ugh.. I'm old) that for all we take from the Internet, it's good to give something back whenever we can.  

So, when I read your comment, again, I was frustrated and had the general feeling of "why did I bother..."  because without supporting doco, it did feel like it was just a quick "meh, this sucks." (Your username had something to do with it for sure, "meh" doesn't come off as someone who "cares".. know what I mean? ... so that biased my response a bit.)   It's clear however that you are someone who cares, just by virtue of the fact that you've taken so much time to respond back to me despite my noticeable frustration. 

Back to the code... the Instance ID did serve my purpose (I'd read other SO posts that suggested Instance ID was what I needed for my use case...  which I didn't bother referencing b/c I didn't want to go hunting for them again - you can relate to why.. time constraints), and really, regardless if you or I or anyone find Instance ID "useful" ... it exists, and I just was adding the capability to this module.  Plus, I did, from the get go, have it in my documentation, calling it out as android only... scroll to the very bottom here: https://github.com/rebeccahughes/react-native-device-info ... so yes, when your comment mentioned "not cross-platform" I had the reaction of "umm yeah, I *said* that.... " 

I've *since* gone back to the page I linked and I see that Instance ID is something one can get on iOS too.   I'd **swear** it wasn't there before (I would have implemented if it was, or I'd seen it)... but Google claims that page hasn't been updated since Sept 2015, so maybe I'm just going senile.  

Overall, this has netted positive for me, and I do feel I've learned something and have benefited from our interactions here.  Perhaps another PR is in my future implementing Instance ID for iOS.  :-)

PR #49, opened 2016-06-09
device name issue on emulator

ankitpatial:
BluetoothAdapter does not work well on emulator. for simplicity we can make use of MANUFACTURER and MODEL.

Best thing can be to make use of https://github.com/jaredrummler/AndroidDeviceNames

Here is reference discussion on stackoverflow http://stackoverflow.com/questions/7071281/get-android-device-name


mehcode:
@ankitpatial This doesn't have the same result. It just concatenates the `Manufacturer` and `Model` parameters that we already capture separately. 

Just live with getting `Unknown` on emulators. It works on a handful of real devices I've tested.

PR #51, opened 2016-06-11
Install with rnpm

dominhhai:
Add instruction of installing via `rnpm`


GantMan:
Since `rnpm` has been merged into React Native CLI, can you update this?

GantMan:
I guess `rnpm` instructions are better than no instructions.   We need to revisit this when people are completely off of `rnpm`

PR #54, opened 2016-06-20
plenty of maintainers now

GantMan:
should be good now, we can lower this flag.


GantMan:
as mentioned in https://github.com/rebeccahughes/react-native-device-info/issues/37  we're good.

PR #59, opened 2016-07-03
Add timezone

mehcode:
I apologize if the iOS version doesn't "just work"; my Objective C knowledge is very minimal right now and I'm blind coding.

### DIFF
--- a/RNDeviceInfo/RNDeviceInfo.m
+++ b/RNDeviceInfo/RNDeviceInfo.m
@@ -141,6 +141,12 @@ RCT_EXPORT_MODULE()
   return country;
 }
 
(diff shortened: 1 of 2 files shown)


RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -141,6 +141,12 @@ - (NSString*) deviceCountry
   return country;
 }
 
+- (NSString*) timezone
+{
+  NSTimeZone *currentTimeZone = [NSTimeZone localTimeZone];
+  return currentTimeZone.name;
```
GantMan: good call on going with localTimeZone, I think that might be the best fit.

PR #61, opened 2016-07-06
Fix crash Android in virtual Method

jerson:


mehcode:
@jerson It's supposed to crash. That's why it's in the `try .. catch` block and then ignores the exception.

jerson:
no, for some reason, the try catch does not work, probably because it is a virtual method

jerson:
Crash in android 4.4 or lower in 5 or 6 its ok

jerson:
Tested also on a real Device with Android 4.4

mehcode:
@jerson Wait.. the try..catch doesn't work? I don't a lot about java so maybe @GantMan can help me here.

Perhaps you're mistaking the stack trace printout to be a crash (in the catch)?

jerson:
try catch, catch the error, but the application is closed anyway

mehcode:
That does seem weird. I'll give it a poke later on some devices. We just added this package to an app we have in production and haven't made an upgrade yet so I'll definitely take a look at this soon.

Maybe its better to rewrite it so it doesn't have the `try .. catch` at all? With what you added we shouldn't need the `try .. catch`.

jerson:
if it sounds weird, but that prevents the application to crash, the try catch think it is only for that possible NullPointerException so maybe we could remove it and just leave if !=null

jerson:
This is the mistake that captures the try catch, but the application error is closed
```
E/BluetoothAdapter( 6165): Bluetooth binder is null
E/LALA    ( 6165): java.lang.NullPointerException
E/LALA    ( 6165): 	at com.learnium.RNDeviceInfo.RNDeviceModule.getConstants(RNDeviceModule.java:78)
E/LALA    ( 6165): 	at com.facebook.react.bridge.BaseJavaModule.writeConstantsField(BaseJavaModule.java:452)
E/LALA    ( 6165): 	at com.facebook.react.bridge.NativeModuleRegistry.writeModuleDescriptions(NativeModuleRegistry.java:82)
E/LALA    ( 6165): 	at com.facebook.react.bridge.CatalystInstanceImpl.buildModulesConfigJSONProperty(CatalystInstanceImpl.java:368)
E/LALA    ( 6165): 	at com.facebook.react.bridge.CatalystInstanceImpl.initializeBridge(CatalystInstanceImpl.java:126)
E/LALA    ( 6165): 	at com.facebook.react.bridge.CatalystInstanceImpl.access$200(CatalystInstanceImpl.java:39)
E/LALA    ( 6165): 	at com.facebook.react.bridge.CatalystInstanceImpl$1.call(CatalystInstanceImpl.java:97)
E/LALA    ( 6165): 	at com.facebook.react.bridge.CatalystInstanceImpl$1.call(CatalystInstanceImpl.java:92)
E/LALA    ( 6165): 	at com.facebook.react.bridge.queue.MessageQueueThreadImpl$1.run(MessageQueueThreadImpl.java:73)
E/LALA    ( 6165): 	at android.os.Handler.handleCallback(Handler.java:733)
E/LALA    ( 6165): 	at android.os.Handler.dispatchMessage(Handler.java:95)
E/LALA    ( 6165): 	at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:31)
E/LALA    ( 6165): 	at android.os.Looper.loop(Looper.java:136)
E/LALA    ( 6165): 	at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:193)
E/LALA    ( 6165): 	at java.lang.Thread.run(Thread.java:841)
D/dalvikvm( 6165): GC_FOR_ALLOC freed 855K, 12% free 8141K/9176K, paused 65ms, total 65ms
I/ReactNativeJS( 6165): Running application "Setbeat" with appParams: {"initialProps":{},"rootTag":1}. __DEV__ === true, development-level warning are ON, performance optimizations are OFF

```

mehcode:
> yes, it is very strange, but that prevents the application to crash, the try catch think it is only for that possible NullPointerException so maybe we could remove it and just leave if !=null

It'd probably read a bit better this way as well.

ssancheza227:
@jerson You solution worked for me :) :+1: :heart:

jerson:
:/ can you just merge it ?

mehcode:
@jerson I'm not a collaborator but didn't we agree on removing the `try .. catch`?

//cc @GantMan

jerson:
I think the try catch if it is useful, but maybe for other errors

ssancheza227:
I think the _**try catch**_ trying of catch the exception however this combierte in error and can not be controlled by the _**catch try**_ and the application closes immediately on the device.  I also think this error should be treated with an _**if**_, and support the _**try catch**_. //cc @jerson :)

ptmt:
Same annoying error, Android 6.0 Nexus 5x emulator. @jerson fix works.

Jagadeesh-Govindaraj:
remove try catch and update the code like this
```
BluetoothAdapter mBluetoothAdapter = BluetoothAdapter.getDefaultAdapter();
if (mBluetoothAdapter == null) {
    // Device does not support Bluetooth
} else {
    if (!mBluetoothAdapter.isEnabled()) {
        // Bluetooth is not enable :)
    }else{
//permission enabled get whatever you want
}
}
```

sibelius:
this works fine

PR #64, opened 2016-07-07
attempt to fix android build issue

gregthewhite:
override RNDeviceModule#canOverrideExistingModule() to return true


gregthewhite:
this is the fix that the logcat told me to try. it remains to be seen if that is the only fix necessary

kevinjasker:
👍

PR #66, opened 2016-07-13
correct Device Name to System Name

emmby:
conflicted with Device Name further down in the list


antoinerousseau:
this can be closed now

PR #69, opened 2016-07-20
Fix missing `TimeZone` import.

kompot:


davidstoker:
+1 Can we get this merged?

PR #71, opened 2016-07-25
Avoid crash when country code is nil

### LINKED ISSUE #70: Crash when country code is nil
`[[NSLocale currentLocale] objectForKey:NSLocaleCountryCode]` [returns nil](http://stackoverflow.com/questions/15202454/nslocalecountrycode-returns-nil) in some cases. This causes a crash when the dictionary is constructed in the `constantsToExport` method.

In my case, I was running UI tests on the simulator.

This field should probably either be omitted or set to `[NSNull null]` when the country code is nil.

bjornorri:
Fixes #70


steinnes:
+1 on this, we're currently installing @bjornorri's version via package.json to avoid this crash when running UITests.

antoinerousseau:
@GantMan and this

GantMan:
Thanks @antoinerousseau - been a helluva week.   :D

PR #75, opened 2016-08-05
Update README.md

alexHlebnikov:
Update Android installation instruction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,23 @@ Run your project (Cmd+R)
 
 ### Installation (Android)
 
-* Add Gradle configuration changes
+* In `android/setting.gradle`
 
-Run `react-native link react-native-device-info` in your project root.
+```gradle
+...
+include ':RNDeviceInfo', ':app'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,11 +48,23 @@ Run your project (Cmd+R)
 
 ### Installation (Android)
 
-* Add Gradle configuration changes
+* In `android/setting.gradle`
 
-Run `react-native link react-native-device-info` in your project root.
```
GantMan: Not sure you're modifying the correct part here.  Using `react-native link` should work.

I thought you were going to remove the directions for `rnpm`

alexHlebnikov: ```react-native link react-native-device-info``` does for me the same thing as ```rnpm install react-native-device-info``` - installs package only on iOS.

GantMan: That's odd, `react-native link` should properly link to MainApplication.java, that's actually the PR I wrote to react native.

alexHlebnikov:
The problem still occurs on RN 0.33.0.RC.0. `react-native link` install package only on iOS.

PR #76, opened 2016-08-05
Use System for System.

joshuapinter:
You had two Device Names in here, which was a little confusing. Changed to use System Name and System Version.


joshuapinter:
👍 

For another README PR, check out #133.

PR #81, opened 2016-08-17
add getBrand() on both android and ios

zxcpoiu:
android is using value from `Build.BRAND`
ios is using fixed value `"Apple"`


GantMan:
Adding to the API requires proper documentation upgrades int he README.md

zxcpoiu:
sure, updated

PR #85, opened 2016-08-18
updating release notes

jeveloper:


jeveloper:
i realized my commit wasn't as great as another person who added changelog :)

PR #92, opened 2016-09-05
added getIsTablet method to detect iPad or Android Tablet screen sizes

iotashan:
Handy method so you can switch layouts or styles for Tablets. Blatantly based on the Appcelerator Titanium / Alloy isTablet property code.

PR #107, opened 2016-10-29
added isTablet method for iOS and Android

tsemerad:
Detects whether you're on a tablet device when using iOS or Android. Inspired by PR #92, but addresses @GantMan's concerns with it.


GantMan:
Looks great!

PR #129, opened 2016-12-07
Update build.gradle

mcz:
Update buildToolsVersion to support faster builds with 23.0.3 build tools


jaggs6:
👍

jaggs6:
@mcz this is not needed any more

PR #155, opened 2017-03-07
Update buildToolsVersion to 25.0.1

jaggs6:
Need to update this in order to use with the latest build tools

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "25.0.1"
+    buildToolsVersion "25.0.2"
 
     defaultConfig {
         minSdkVersion 16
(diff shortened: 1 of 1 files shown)


mquander:
+1, I had to do this and it would be nice to have in upstream.

jaggs6:
Thanks for merging this. Would be great if we can update the npm version as well.

GantMan:
there's one more PR coming that I would want.  Give me a min to wait for that one then release.

huston007:
That's good to keep latest versions, but React Native itself [still uses 23.0.1](http://facebook.github.io/react-native/releases/0.42/docs/getting-started.html#android-development-environment) version. So now I should have both versions installed just to get compiled one of dependencies. 
I don't think that's really straightforward.

mquander:
It's true that it says that in the docs, but in your React Native application it's easy to set the build tools version to whatever you like in your build.gradle.

It's possible that there is some reason they preferred 23.0.1 by default though -- I have no idea.

PR #156, opened 2017-03-07
Add PIN/Fingerprint Check to Device Info

kacynMedallia:


kacynMedallia:
@7ynk3r opened https://github.com/rebeccahughes/react-native-device-info/issues/153 a few days ago.  This is the pull request to go along with that.

@passabilities @rebeccahughes @GantMan

GantMan:
very nice!  @kacynMedallia - do you have a second to answer a few questions regarding this?

Can you tweet me?  @gantlaborde ?

kacynMedallia:
@GantMan Done!

RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -162,6 +163,12 @@ - (bool) isTablet
   return [[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad;
 }
 
+- (bool) isPinOrFingerprintSet
+{
+  LAContext *context = [[LAContext alloc] init];
+  return ([context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:nil]);
```
GantMan: just because Biometrics fails, doesn't mean pin should be false, right?

kacynMedallia: Good catch.  LAPolicyDeviceOwnerAuthentication will behave as intended, but only for iOS 9.0+.  In the case of iOS <9.0, I can fall back on LAPolicyDeviceOwnerAuthenticationWithBiometrics and write a note about this in the README.

kacynMedallia:
@GantMan Fixed the iOS code and README

GantMan:
LGTM! 👍

7ynk3r:
@GantMan we found that `isPinOrFingerprintSet` value might change while the app is running. If that it's the case, beacuse the values are exported as constants, the only way to reload the values is to re-open the application. 

Any advice on how to deal with this problem? Maybe export the function with a differnt method name so it doesn't overlap with the contants? Remove it from the contant list? 

Thanks!
JM

GantMan:
So the best thing we can do is emit events.   So if it changes and our native code can fire, we can emit an event, and so apps that really care about catching this information when it's fresh, can then subscribe for those events and update accordingly in their own state.

I'd be willing to have something like that in here if you're interested.  But I'm unfamiliar with the specific event you're looking for.  Examples of emitting events are easy to come by.

kacynMedallia:
@GantMan Thanks for getting back to us.  I think it's a simpler solution to use callbacks in this case, so apps using this don't have to set up listeners.  I tested this method and it seems to work well.

Please see my PR here for the fix using callback functions: https://github.com/rebeccahughes/react-native-device-info/pull/157

Let us know what you think!

cc @7ynk3r

PR #157, opened 2017-03-10
Change PIN/Fingerprint check from constant to callback

kacynMedallia:


GantMan:
you're right, that is simpler.  You'll need to document this in the Readme.

kacynMedallia:
@GantMan Updated README

GantMan:
published 10.1

Issue #142, opened 2017-01-18
ios: getUserAgent not same as real user agent !


atteia:
hello
I get user agent 
`Mozilla/5.0 (iPhone; CPU iPhone OS 10_1 like Mac OS X) AppleWebKit/602.2.14 (KHTML, like Gecko) Mobile/14B72`
when using your **DeviceInfo.getUserAgent()**

but the real useragent which posted to Apache is 
`AppName/1 CFNetwork/808.1.4 Darwin/16.3.0`

can I get the real one ?

mikehardy:
I believe this is related to how we get the user agent. Whether it's using native networking primitives within react-native, or using webviews. We get the WebView, this is covered elsewhere but if you want the react-native primitive response you'd need a new API. Either way, resolving this is very stale and it's covered elsewhere

PR #189, opened 2017-06-26
Actual WebView User Agent support for Android.

Naoto-Ida:
I have noticed that the current `userAgent` implementation for Android returns the device's user agent, not the WebView's user agent, as suggested in posts such as [this](https://stackoverflow.com/a/19795789/4776002) on Stack Overflow.

Exposing the WebView's `userAgent` as `webViewUserAgent` will benefit many developers who want to use it in their Android implementation, i.e. append something to the preexisting user agent.

The only problem I have is that since `userAgent` on the iOS-side already returns the expected WebView user agent, the implementation of `webViewUserAgent` for iOS will merely return the same value as `userAgent`.


Naoto-Ida:
Forgot to update Readme, closing for now.

Naoto-Ida:
I understand that there hasn't been activity lately in regards to merging requests,
so if anybody is interested in these additions:

```
"dependencies": {
  "react-native-device-info": "git+https://github.com/Naoto-Ida/react-native-device-info.git#webview-ua"
}
```

Cheers.

leethree:
Would be great if this PR is merged. 👍

Naoto-Ida:
I hope so too...

Naoto-Ida:
Any updates on this? I can fix the conflicts if there's any plans of merging this.

GantMan:
will merge, ping me when done with conflicts.

leethree:
Actually the iOS `userAgent` returns the WebView user agent, instead of the default HTTP user agent like this:
```
<product-name>/<build-number> CFNetwork/548.0.3 Darwin/11.2.0
```

which is the opposite behavior for Android clients. For backwards compatibility, I suggest deprecate the existing `DeviceInfo.getUserAgent()` method and add
- `DeviceInfo.getWebViewUserAgent()`
- `DeviceInfo.getHTTPUserAgent()`

See also: #142

machour:
@Naoto-Ida what's the status on this PR?
@leethree comment seems like the good way to go if there is indeed an inconsistency.

DracoBlue:
Good idea!

DracoBlue:
The android WebView is already the real android webview useragent, since 6 months (in my understanding) https://github.com/rebeccahughes/react-native-device-info/commit/41d0ca9a88429bf7057661a1be878a772c72600b

Naoto-Ida:
@DracoBlue Yes I believe you can close this.

PR #196, opened 2017-08-02
Remove deprecated `createJSModules`

ground5hark:
Since react-native 0.47


brunsy:
There are already 2 PR's for this. See #196 and #186

PR #217, opened 2017-09-14
Add DeviceUID.h to public headers

crafterm:
Hi @rebeccahughes 

Hope all is going well.

This PR adds the DeviceUID.h file to the 'copy files' build phase, making it available to import and use at the Objective-C/Swift level in your app. This makes it possible to reuse the same code used at the react native level,  in the rest of your project, for example in extensions or other native code.

Cheers,

Marcus


GantMan:
thanks for the PR

PR #221, opened 2017-09-16
fix(Types): TS type definitions didn't use the correct export mechanism

itchingpixels:
Reverting #12c93cb that has broken TS support.

based on @aarondail's issue: 
https://github.com/rebeccahughes/react-native-device-info/issues/208

We've bumped into the same problem.


GantMan:
YES!  Will always merge awesome typescript!

PR #222, opened 2017-09-18
First and Last Install Time

rafaeltakiyama:
Add firstInstallTime and lastUpdateTime from Android PackageInfo

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java
@@ -1,15 +1,14 @@
 package com.learnium.RNDeviceInfo;
 
-import java.util.Arrays;
-import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java
```diff
@@ -1,15 +1,14 @@
...
 import com.facebook.react.ReactPackage;
+import com.facebook.react.bridge.JavaScriptModule;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.uimanager.ViewManager;
-import com.facebook.react.bridge.JavaScriptModule;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
```
GantMan: Looks like you removed `java.util.Arrays` was that on purpose?  Was it unused?

PR #224, opened 2017-09-20
Added device detection even when in an iOS emulator.

douglasjunior:
With the release of the iPhone X it is essential to know the device model to change the behavior of View, even when using an iOS Simulator. So I implemented the function to detect the device model even inside the simulator.

Changes:
- Added device detection even when in an iOS emulator.
- Additionally, added device types for iPhone 8, iPhone 8 Plus and iPhone X.


GantMan:
Fantastic work!

PR #247, opened 2017-10-11
Check for null with getCurrentActivity()

rizzomichaelg:
getCurrentActivity() occasionally returns null


GantMan:
thanks!  can you add this to the changelog too?

rizzomichaelg:
will do!

GantMan:
thanks!

PR #257, opened 2017-10-27
jail

fighting300:


GantMan:
This seems pretty specific and it changes a lot.

I'd recommend instead trying a different library like my other one here: https://github.com/GantMan/jail-monkey

It's best not to mix opinionated items into this particular lib.

PR #278, opened 2017-12-11
feat: add getFontScale()

SteffeyDev:
It is common to change the layout based on the font size when adapting an app for accessibility setting.  This allows developers to fetch the font scale easily so that they can determine if they want to use a layout that looks better with larger texts (change number of columns, amount of padding, etc)

The font scale attribute is built in to Android, and I adapted it to iOS by using the font size details from https://developer.apple.com/ios/human-interface-guidelines/visual-design/typography/

I am open to changes in the explanation I added in the README, not sure if I put too much detail or anything.  Also open to discussing implementation details, though I believe this is a fairly basic change.


SteffeyDev:
@rebeccahughes I'd love to hear back on this when you get the chance.

GantMan:
Rebecca doesn't really maintain this repo anymore.

It's a bit of a democracy now.  If you can get other people saying this is a good add, then it will get added.

SteffeyDev:
@GantMan Good to know, so if I don't know people in the community I can't get things merged?  I do worry about projects without a POC because it gets stalemated and everyone just uses their own forks (case in point: react-native-popover).

machour:
@SteffeyDev I'm a new maintainer trying to clear the issues/pr queue, and this does look like a neat feature to me! Thank you for this.

I'll try to test your PR soon and get it merged for 0.15.0

RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -200,6 +200,27 @@ - (bool) isTablet
   return [[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad;
 }
 
+- (NSNumber*) fontScale
```
machour: Please add a comment linking to https://developer.apple.com/ios/human-interface-guidelines/visual-design/typography/

SteffeyDev:
It reloads on Android on 7.1.1 in my app, so it must be a configuration thing.  Are you running normally, with expo, or in create-react-native-app?

SteffeyDev:
And do you have anything under `activity` > `android:configChanges` in your manifest?

SteffeyDev:
> android:configChanges
Lists configuration changes that the activity will handle itself. When a configuration change occurs at runtime, the activity is shut down and restarted by default, but declaring a configuration with this attribute will prevent the activity from being restarted. Instead, the activity remains running and its onConfigurationChanged() method is called.

Source: https://developer.android.com/guide/topics/manifest/activity-element.html#config

SteffeyDev:
Because react native apps only use one activity, it will be reloaded for any configuration change not exempted by including it in the `configChanges`

SteffeyDev:
Regarding the rounding of numbers on Android, I would lean toward not, though you certainly could for your test app on the output.  Rounding is best done immediately before use, and should be unnecessary for most use cases (greater-than, less-than tests).

machour:
I'm using this ejected CRNA app to test the library: https://github.com/machour/react-native-device-info-test, with this in configChanges:

```
android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
```

so I guess I'm missing `fontScale` to get this automatically reloaded. Might be worth noting in the documentation.

As for rounding, agreed, let's leave the choice to the user!

README.md
```diff
@@ -389,8 +390,30 @@ const firstInstallTime = DeviceInfo.getFirstInstallTime();
 // Android: 1517681764528
 ```
 
+**Notes**
```
machour: Can you review (update if needed) and move this section under `getFontScale()`? Seems like my first coffee didn't completely kick in when I was resolving conflict.

machour:
@SteffeyDev could you please take care of the two review comments I left you? I'll merge this PR as soon as it's done. (sorry for messing up the docs 😆)

SteffeyDev:
@machour Actually, if you read the documentation I copied in, it looks like the app will be reloaded if you *do not* have `fontscale` in `android:configChanges`.  This means that it should be working in your app.  In the app I have it working in, I don't have it in the list.  This means that something about your setup isn't working, or something else about my setup allows it to work.

SteffeyDev:
I pushed the changes with the README updated per my last comment, but I would love to figure out why it's not working on your app.

machour:
I'd love that too. Could you clone my repo and give it a shot?

SteffeyDev:
I tried downloading and running your test app on my android Moto E, but all I get is a white screen (after it pulls the index file).  Do you have to do anything special to get it working?

SteffeyDev:
Ok, for me your test app does reload when I change the font scale, so it is probably an issue with your device.  Do you have another android device you can try or maybe test on a simulator?

machour:
I was testing with a Nexus 5 Simulator.
Now I'm testing on my OnePlus 5T (Android 7.1), same thing.

Here's what I do so that we're on the same page:

- Display getFontScale() result in the app
- Load the App on my phone, fontScale shows **1**
- Go to settings > display > font size, make it wider
- Switch back to the app, **1** is still displayed, same font size
- Close the App and start it again from the home menu
- fontScale show 1.299 and fonts are indeed bigger

Is this how you're testing on your side?

machour:
item 4, I meant bigger font size, sorry. The new font size is indeed applied, but fontScale still displays the old value.

SteffeyDev:
Ok, that makes much more sense.  I couldn’t actually figure out how to get your test app to display the fontScale, I tried adding `getFontScale` to the `methods` array, but all it showed in the test app was `undefined getFontScale():` with no value after.  I did swap in my branch of `react-native-device-info` in the node modules.  I thought the issue was with the automatically reload, not with fetching the wrong font size after that reload.  Can you share your branch of the test app that includes the font scale so I can test with that?

SteffeyDev:
On my side, I say the font size change in the test app, but because I couldn’t get it to show the font scale I didn’t see whether or not the value was changing.

machour:
did you run `react-native link` before running the app again? 

On my side, I just took your PR [patch](https://github.com/rebeccahughes/react-native-device-info/pull/278.diff) and applied it inside `node_modules/react-native-device-info/` using `patch -p1 < 278.diff`
Then I ran the `react-native link && react-native run-android`

machour:
I think that to achieve what we want, we'd need to be able to register a listener, something like this:
   https://github.com/yamill/react-native-orientation/blob/master/android/src/main/java/com/github/yamill/orientation/OrientationModule.java#L48

Come to think about it, other methods in react-native-device-info could benefit from an OnConfigurationChanged listener (like is24hour() for example).

Maybe we could just drop the extra **Notes** section from the README.md for now, and come up with a global implementation in the future.

SteffeyDev:
What's really funny is that I already created a version with an `onConfigurationChange` listener, but later ditched that because there were no other listeners in the library and I wanted to keep to the nature of the library.  I still have the branch however, should I start looking into it again?

machour:
Excellent 😄Definitely! (in another PR of course)

SteffeyDev:
Ok, I'll work on that.  Would you prefer to have a `getInitialFontScale` and then a separate `listenForFontScaleChanges` methods, or do you want one `listenForFontScale` method that always passes the initial font size to the callback?

machour:
I really didn't put that much thought into this yet... I think we should keep `getFontScale()` as it is, and have something like `DeviceInfo.addListener('fontScale', function(newScale) { /* ... */ });` ... not sure yet.

machour:
By the way, I think I might publish 0.15.0 tomorrow. 
Could you remove the Notes in README so that I can merge this PR and ship it? 🚢

SteffeyDev:
Ok, done 👍

machour:
Thank you so much for bearing with me @SteffeyDev 🕺

machour:
0.15.0 is out with your changes, enjoy!

PR #289, opened 2018-01-04
Added memory functions

IjzerenHein:
Hi, this  PR adds support for two functions to query information about the memory:

## getTotalMemory (iOS & android)
Returns the total amount of RAM installed on the device. On iOS this is the physical amount of memory on the device. On Android this is the physical amount of memory, minus some critical memory reserved for below kernel fixed allocations, like DMA buffers, RAM for the baseband CPU, etc.
https://developer.apple.com/documentation/foundation/nsprocessinfo/1408211-physicalmemory?language=objc
https://developer.android.com/reference/android/app/ActivityManager.MemoryInfo.html#totalMem

## getMaxMemory (android only)
Returns the maximum amount of memory available to the process. This depends mostly on the total amount of RAM on the device, but also on Android version and the vendor.
A process can typically not exceed this limit, so it is useful to know if you are experiencing memory issues with react-native. This limit can be increased by setting `largeHeap=true` in the android manifest.
https://developer.android.com/reference/android/app/ActivityManager.html#getLargeMemoryClass()
https://stackoverflow.com/questions/2630158/detect-application-heap-size-in-android


GantMan:
can you please add your change to the changelog?

IjzerenHein:
Of course, will do

IjzerenHein:
Done

GantMan:
👍

PR #291, opened 2018-01-07
Adding Getters for Initial Screen Size

SteffeyDev:
With the dawn of split-screen capabilities, the `Dimensions` API no longer accurately tells you the size of your app container; it only tells you the screen size.  Many times, you want to know the width and height your app has on first load, before any `onLayout` callbacks can return it for you.  This would add two methods `getInitialWidth` and `getInitialHeight` that return the initial container size for your app.

Common scenarios would be detecting when your iPad app is launched in split-screen mode over another app, and loading the iPhone layout instead of a full-width iPad layout. Knowing the initial size right away can prevent that 1-frame stutter caused by using the `onLayout` callback.

This is also useful if you use a `TabNavigator` from `react-navigation`, which can take an `initialLayout` in order to prevent the same 1-frame delay in rendering the tabs correctly.


machour:
👋

before reviewing changes, I'm wondering if it wouldn't be better to provide a single function called `getInitialScreenSize()` that would return a `{ width: number, height: number }` structure (similar to `Dimensions.get('window')`)

this would help keep the public API slim, and would avoid the repetitions in the native implementation..

what do you think?

SteffeyDev:
That sounds like a good idea, I’ll get on that

SteffeyDev:
Ok, new version implemented and tested.

RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -200,6 +200,20 @@ - (bool) isTablet
   return [[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad;
 }
 
+- (CGFloat) width
```
machour: This name looks a bit too generic to me.

Additionally, would you consider merging `height` and `width` into `getInitialScreenSize` on the native side as well? I think returning a `NSDictionary` should be the way to go

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -94,6 +94,16 @@ private Boolean isTablet() {
...
+  private float width() {
+    Configuration config = getReactApplicationContext().getResources().getConfiguration();
+    return config.screenWidthDp;
+  }
+
+  private float height() {
+    Configuration config = getReactApplicationContext().getResources().getConfiguration();
+    return config.screenHeightDp;
+  }
+
```
machour: Same remarks for Android (names are too generic + let's try to merge this internally)

deviceinfo.d.ts
```diff
@@ -21,6 +21,7 @@ export function getTimezone(): string;
 export function getInstanceID(): string;
 export function isEmulator(): boolean;
 export function isTablet(): boolean;
+export function getInitialScreenSize(): object;
```
machour: Is it possible to define the returned object structure a bit more?
Not a typescript expert, but I think something like this should do the trick:

```typescript
interface IScreenSize {
   width: number;
   height: number;
}

// ...

export function  getInitialScreenSize(): IScreenSize;
```

SteffeyDev:
Ok, I made all the requested changes and tested.  I also added to `deviceinfo.js.flow`, but I'm not sure if I did it right, so let me know.

mikehardy:
@SteffeyDev  This is very stale unfortunately - apologies - however, have you considered alternate solutions like https://www.npmjs.com/package/react-native-extra-dimensions-android ?

They might be interested in initial layout information before screen splits or similar - it just seems like a library much more focused on layout and dimensions might handle this better.

At the same time, I can see how the real official screen size (which is what this would return, I believe?) is a natural fit for device info

Before I spend time working through this to shape it for possible merge, let me know what you think

PR #297, opened 2018-01-21
Add documentation about is24Hour function

pterolex:


Noitidart:
Is there a event listener on this? So we can detect change that happens during run time?

Issue #311, opened 2018-02-06
getIPAddress & getMACAddress errors


elliotfleming:
In latest version, the following methods are called in the assignment unlike the rest of the methods... I tested by removing the '()' for each and they both worked as expected.

* `getIPAddress()`
* `getMACAddress()`

elliotfleming:
Fixed and submited pull request #312

machour:
Hi @elliotfleming, the two methods are called in the assignment because unlike the rest of the methods, they return a Promise:

```js
  componentWillMount() {
    this.setState({ ip: false });
    DeviceInfo.getIPAddress().then(ip => this.setState({ ip }));
  }

  render() {
    const { ip } = this.state;
    return (
      <View style={styles.container}>
        <Text>IP: {ip}</Text>
      </View>
    );
  }
```

<img width="326" alt="screen shot 2018-02-06 at 9 17 44 am" src="https://user-images.githubusercontent.com/304450/35848495-a10d7992-0b1e-11e8-8c25-6a99789ca752.png">

If I remove the parenthesis as proposed in your PR, here's the result:

<img width="293" alt="screen shot 2018-02-06 at 9 15 24 am" src="https://user-images.githubusercontent.com/304450/35848509-a89e4984-0b1e-11e8-8b7a-467dd9ba041f.png">

Please re-open if you have a working example without parenthesis, or if I missed something obvious!

PR #312, opened 2018-02-06
Update deviceinfo.js

elliotfleming:
Remove function calls for getIpAddress and getMacAddress


machour:
Closed as per my comment in #311 😉

PR #313, opened 2018-02-06
fix: enhance isTablet() on Android

mikpalm:
Changed the nature of IsTablet for Android, since newer devices with higher dp and screensizes returns as tablet, needs further testing.

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -100,8 +100,6 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
               || metrics.densityDpi == DisplayMetrics.DENSITY_TV
               || metrics.densityDpi == DisplayMetrics.DENSITY_XHIGH) {
         return true;
-      } else {
-        return false;
       }
     }
     return false;
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -18,6 +18,7 @@ import android.webkit.WebSettings;
 import android.telephony.TelephonyManager;
 import android.text.format.Formatter;
 import android.app.ActivityManager;
+import android.util.DisplayMetrics;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -90,8 +90,21 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   private Boolean isTablet() {
-    int layout = getReactApplicationContext().getResources().getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK;
-    return layout == Configuration.SCREENLAYOUT_SIZE_LARGE || layout == Configuration.SCREENLAYOUT_SIZE_XLARGE;
+    boolean layout = (getReactApplicationContext().getResources().getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK) >= Configuration.SCREENLAYOUT_SIZE_LARGE;
+    if (layout) {
+      DisplayMetrics metrics = new DisplayMetrics();
+      getCurrentActivity().getWindowManager().getDefaultDisplay().getMetrics(metrics);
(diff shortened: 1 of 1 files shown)


Ingibjorg:
@mikpalm  and @machour how is this coming along? The issue is tagged with "help wanted". How can I help?

machour:
@Ingibjorg we need more testing as mentioned by Mikael.
Do you have a failing isTablet() test that is fixed by this PR?

mikpalmen:
Can be tested on almost any newer android. Change the resolution from standard to high. Then logout the response from isTablet. it will give you true. when this fix is not applied. I have only tried it on 2 android phones but on Emulators for tablets and more phones.

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -90,8 +90,21 @@ private Boolean isEmulator() {
...
+    if (layout) {
+      DisplayMetrics metrics = new DisplayMetrics();
+      getCurrentActivity().getWindowManager().getDefaultDisplay().getMetrics(metrics);
+      if (metrics.densityDpi == DisplayMetrics.DENSITY_DEFAULT
+              || metrics.densityDpi == DisplayMetrics.DENSITY_HIGH
+              || metrics.densityDpi == DisplayMetrics.DENSITY_MEDIUM
+              || metrics.densityDpi == DisplayMetrics.DENSITY_TV
+              || metrics.densityDpi == DisplayMetrics.DENSITY_XHIGH) {
+        return true;
+      } else {
```
machour: The else can be dropped, `false` is returned at the end of the function

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -90,8 +90,21 @@ private Boolean isEmulator() {
   }
 
   private Boolean isTablet() {
-    int layout = getReactApplicationContext().getResources().getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK;
-    return layout == Configuration.SCREENLAYOUT_SIZE_LARGE || layout == Configuration.SCREENLAYOUT_SIZE_XLARGE;
+    boolean layout = (getReactApplicationContext().getResources().getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK) >= Configuration.SCREENLAYOUT_SIZE_LARGE;
```
machour: Even if `Configuration.SCREENLAYOUT_SIZE_LARGE` is 3 and `SCREENLAYOUT_SIZE_XLARGE;` is 4, I wouldn't rely on that as we never know what may be added in the future..

Can you switch back to testing both values? 🙏

machour:
Ew, sorry I thought that `needs further testing` was referring to this PR :)

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mikpalmen: Removed the unnecessary else. What do you meen by switching back to test both. Im doing bigger than or equal?
```
 >= Configuration.SCREENLAYOUT_SIZE_LARGE;
```

machour:
```
int layout = getReactApplicationContext().getResources().getConfiguration().screenLayout & Configuration.SCREENLAYOUT_SIZE_MASK;

if (layout != Configuration.SCREENLAYOUT_SIZE_LARGE && layout != Configuration.SCREENLAYOUT_SIZE_XLARGE) {
  return false;
}

// rest of your code, without the if (layout)
```

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -90,8 +90,19 @@ private Boolean isEmulator() {
...
+    if (layout) {
+      DisplayMetrics metrics = new DisplayMetrics();
```
Ingibjorg: I got a build error on this line: `error: cannot find symbol DisplayMetrics`. Did you forget to import android.util.DisplayMetrics?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mikpalmen: done

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
machour: Nice catch!

mikpalmen: Yes fixing.

machour:
Thank you both 🙌

I'll test this PR tomorrow and will ship it in 0.17.2

Issue #39, opened 2016-04-16
Crash introduced in 0.9.2


dudeinthemirror:
After upgrading to `0.9.2`, I have this crash upon starting the app:

```
2016-04-15 17:22:54.850 MyApp.STAGING[6639:2558933] Loading JS bundle from file:///Users/dudeinthemirror/Library/Developer/CoreSimulator/Devices/DD3C4AE0-310F-4E3B-86F6-1AF87F33309F/data/Containers/Bundle/Application/D5E5316E-BAC7-4D92-9334-8B136FCA8327/MyApp.STAGING.app/main.jsbundle
2016-04-15 17:22:55.314 MyApp.STAGING[6639:2558933] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[11]'
*** First throw call stack:
(
    0   CoreFoundation                   0x0000000110d47d85 __exceptionPreprocess + 165
    1   libobjc.A.dylib                  0x000000011078fdeb objc_exception_throw + 48
    2   CoreFoundation                   0x0000000110c4423e -[__NSPlaceholderDictionary initWithObjects:forKeys:count:] + 318
    3   CoreFoundation                   0x0000000110c565bb +[NSDictionary dictionaryWithObjects:forKeys:count:] + 59
    4   MyApp.STAGING                    0x000000010b6362c3 -[RNDeviceInfo constantsToExport] + 924
    5   MyApp.STAGING                    0x000000010b5dfbad __32-[RCTModuleData gatherConstants]_block_invoke + 41
    6   MyApp.STAGING                    0x000000010b61d676 RCTExecuteOnMainThread + 53
    7   MyApp.STAGING                    0x000000010b5dfb74 -[RCTModuleData gatherConstants] + 156
    8   libdispatch.dylib                0x000000011197dd9d 
```

By changing `package.json` 
from

```
    "react-native-device-info": "^0.9.1",
```

to

```
    "react-native-device-info": "0.9.1",
```

the crash goes away.

jerson:
the same here :/

rebeccahughes:
Can you try with 0.9.3 please?

dudeinthemirror:
Yes, 0.9.3 fixes it. 
Thank you

jerson:
👍  thanks

cnjon:
+1

[...]

Issue #89, opened 2016-08-29
Runtime failure - crash on UIWebView


pospetur:
Followed the instructions for installing 0.9.4 and tried both automatic and manual install. But it will crash when I run the app on my iPhone with these messages =>
(will you help?)

**\* Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '**\* -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[10]'
**\* First throw call stack:
(0x18246adb0 0x181acff80 0x18235377c 0x182353614 0x10058f45c 0x1004447c8 0x10045eb24 0x1004446a8 0x1004361e8 0x100cbda7c 0x100cbda3c 0x100cc34e4 0x182420d50 0x18241ebb8 0x182348c50 0x183c30088 0x187632088 0x100034f94 0x181ee68b8)
libc++abi.dylib: terminating with uncaught exception of type NSException

M1chaelTran:
i'm getting the same error.

would be good if someone could help us out.

machour:
Is this still happening with 0.14.0 ?

patmisch:
I'm still seeing this with 0.14.0

machour:
@patmisch how did you do the linking ? what is the call that is provoking the runtime error ? Thanks!

simonheys:
@machour The crash is this line: https://github.com/rebeccahughes/react-native-device-info/blob/1ec397f306317751fe0754a93c804da6756c9280/RNDeviceInfo/RNDeviceInfo.m#L175

UIWebView does not initialise (not clear why), so userAgent function returns nil, which causes an attempt to insert nil into the dictionary here: https://github.com/rebeccahughes/react-native-device-info/blob/1ec397f306317751fe0754a93c804da6756c9280/RNDeviceInfo/RNDeviceInfo.m#L262

...which results in the error

machour:
Thank you for the feedback.. any clues on how to fix this one? 🤔

machour:
Should be fixed in 0.15.3 🎉

ntphong1990:
I facing it in 0.21.5
I back to 0.15.3 and it's gone

machour:
Please open a new issue.

PR #328, opened 2018-02-15
fix: prevent insertion of nil values in the dictionary

machour:
## Description

Fixed issues #39 and #89. 

The problem was that `nil` was potentially inserted into the returned NSPlaceholderDictionary.


ahfarmer:
@machour looks good to me!

machour:
Thank you so much. Just tested this in an app, looks fine.

PR #329, opened 2018-02-19
Fix Android Crashes in Versions <= 4.2.2 (Api 18) 

ErickMaeda:
## Description

The Android Version 4.2.2 or fewer was getting crash when importing the react-native-device-info.

The uses of method *getTotalDiskCapacity* and *getFreeDiskStorage* was getting crash because was using the method *getAvailableBlocksLong* which was added only in Android 4.3 

## StackTrace

```
Fatal Exception: com.facebook.react.common.JavascriptException: C++ Exception in 'NativeModules': java.lang.NoSuchMethodError: android.os.StatFs.getBlockCountLong
```

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅❌     |

## Checklist

* [X] I have implemented an verification of Android Version to get the correct method compatilble

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -127,25 +127,23 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
-  public Integer getTotalDiskCapacity() {
-    try {
-      StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
+  public long getTotalDiskCapacity() {
+    StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
+    if (android.os.Build.VERSION.SDK_INT <= android.os.Build.VERSION_CODES.JELLY_BEAN_MR1){
       return root.getBlockCount() * root.getBlockSize();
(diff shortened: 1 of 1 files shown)


machour:
Hi @ErickMaeda and thank you so much for opening this PR.

I'm a bit puzzled, `getAvailableBlocksLong` usage was introduced in 0.15.0 and removed in 0.15.1
Did you notice the crash with 0.15.1+, or were you using 0.15.0?

Anyway, your fix looks cleaner than mine, I'll review and release it shortly!

ErickMaeda:
I Noticed this error by Fabric Crashlytics, and was just happening on Android Versions 4.2.2 or Fewer.

So I found on the Android Developers Documentation [https://developer.android.com/reference/android/os/StatFs.html](this) Summary which shows that ```StatFs.getBlockCountLong()``` and ```StatFs.getAvailableBlocksLong()``` were just added on api (**18+**) or (**Android Version 4.3+**)

machour:
I was asking about the version of `react-native-device-info` your app shipped when getting the problem ;)

ErickMaeda:
Yes, I noticed on 0.15.0

mikehardy:
This was resolved and now uses BigIntegers and some math to get the integer-overflowing block size but without using the API-4.x-dangerous APIs. I've tested it on Android down to API16 and it works, so this is superceded/obsolete

PR #333, opened 2018-02-21
feat: Google play services versioning

kingbin:
## Description

Added a new Gradle build config that pulls the root-project ext variable  `googlePlayServicesVersion` if it exists and sets the module's play-services-gcm version. It defaults to `+` if it doesn't exist mimicking the current functionality. 

## Compatibility

Gradle Build Configuration

| Android |    ✅     |

## Checklist

* ✅ I have tested this on a device/simulator for each compatible OS
* ✅ I added the documentation in `README.md`.
* ✅ I mentioned this change in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### next
 
+* Add Gradle build config that allows a consumer to set Play Services Version from a root-project ext variable.
+
 ### 0.15.3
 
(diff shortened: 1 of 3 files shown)


machour:
Neat stuff @kingbin! Would that solve the problem #334 is trying to fix?

kingbin:
I wrote this exactly for that reason. I'm using this packing along side react-native-push-notification, FCM and react-native-maps.

Issue #307, opened 2018-02-04
Ideas for version 3


machour:
At some point, we are going to have to release a BC breaking version in order to fix some inconsistencies in this library. While we're at it, we might want to rethink the implementation a bit.

Here are some of my findings after having spending some time triaging issues, reviewing pull requests and reading the source code:

### Return type inconsistency

Findings:
* getBuildNumber() returns a string on iOS, and a number on Android (#214). Should be a number IMHO
* The windows implementation sometimes returns "not available" when a feature is missing
* Some methods are not implemented at all on a target and causes exceptions when called. Example: getIPAddress()

Proposal:
* No method should throw an exception if not implemented.
* Exception should only be thrown for Android when a permission is missing
* Unimplemented methods should *always* return `undefined`.

### Namespacing

Findings:
The API is getting bigger and deals with a lot of mixed stuff (networking, telephony, calendar, ..).

Proposal:
I think that introducing some king of namespacing could help keeping it tidy:

```js
DeviceInfo.telephony.getPhoneNumber();
DeviceInfo.network.getMacAddress();
DeviceInfo.network.getIPAddress();
// ...
``` 

We would also split the native code files by namespace

### Naming

Naming is inconsistent: `getDeviceName()` vs `getBrand()`.

We should take advantage of namespaces introduction to tidy up names.

### Multi SIM

Currently, telephony related functions returns data for the first SIM card. Might be good to introduce multi SIM card support.

Same goes for network informations (#259)

---

Feel free to join the conversation and suggest other or better ideas!

machour:
### Promises vs Callback

isPinOrFingerprintSet() returns a callback while getIPAddress and getMACAddress uses promises. 

Proposal: 
Stick to promises

machour:
As mentioned in #314, we should be using either `Id` or `ID` consistently.

machour:
Introduce listeners for configurations changes? https://github.com/rebeccahughes/react-native-device-info/pull/278#issuecomment-363507963

mvanroon:
shouldn't methods that start with is (like isEmulator) return a boolean?

machour:
@mvanroon thank you for jumping in, I was feeling kind of lonely here 😆 

All the `isXXXX` method normally already return a boolean, did you have a counter example?
(if a method is unsupported on a platform, it should return `undefined`)

mvanroon:
poor guy!

I removed my initial comment because I though I misread the API.
[...]

PR #334, opened 2018-02-21
Removes getInstanceID and Google Play Services dependency

mikemorris:
## Description

Removes `getInstanceID()` which allows for removal of Google Play Services dependency. There are [numerous open issues](https://github.com/rebeccahughes/react-native-device-info/search?q=gcm&type=Issues&utf8=%E2%9C%93) related to versioning conflicts or missing/failed dependency resolution for, IMHO, marginal utility gained from this requirement. The change in https://github.com/rebeccahughes/react-native-device-info/pull/226 to make this dependency optional doesn't alleviate versioning conflicts if another dependency (like https://github.com/zo0r/react-native-push-notification) requires `play-services-gcm`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |
| Windows |    ✅    |

## Checklist

* [x] I mentionned this change in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### next
 
+* Removes `getInstanceID` and Google Play Services dependency (https://github.com/rebeccahughes/react-native-device-info/pull/334)
+
 ### 0.15.3
 
 * Fix crash on iOS: prevent insertion of nil values in the dictionary (https://github.com/rebeccahughes/react-native-device-info/pull/328)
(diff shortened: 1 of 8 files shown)


machour:
Hi @mikemorris and thank you for the proposal. 
I understand your concern, but your PR introduce a huge BC break for people using getInstanceId..
Wouldn't be possible instead to just drop the dependency from the Gradle file and instruct people to add it in their project if needed ?

machour:
Also, check the solution implemented in #333

machour:
Just merged #333 which solves the problem without breaking backward compatibility.
Please let me know if you're still facing an issue with versioning, and we'll get that sorted.

mikemorris:
Given that [GCM is being deprecated in favor of FCM for API 26+](https://developers.google.com/cloud-messaging/android/client), I'd still strongly consider removing this dependency in an upcoming semver major release.

machour:
Will definitely consider it for #307

Issue #323, opened 2018-02-12
Add support for web


peacechen:
## Description

I'm using [react-native-web](https://github.com/necolas/react-native-web) to build a RN project for the browser.  React-native-device-info tries to call native methods which of course don't exist in the browser. 

## Use Case

I've started creating a `web` folder that simply returns an object with the data/methods that react-native-device-info supports.  Most of it is empty since the browser doesn't have much of the information (the few exceptions include userAgent).  `deviceinfo.js` checks whether the Platform.OS is web and conditionally requires the web module or the native module.

Would a PR be accepted?

machour:
Hey @peacechen, I heard of react-native-web, and I'm really curious about seeing a real-life example where it's used. Is your project available for iOS, Android AND web? (if it's a public project, I'd love to check it out).

This being said, go ahead and open the PR, even if it's still a work in progress. We'll figuring things out from there.

peacechen:
We have an iOS app live:
https://itunes.apple.com/us/app/onceconnect/id1308143592?ls=1&mt=8

Android will be released eventually.  Current priority is a web release.  This is a full production app with lots of pieces.  Getting it all sorted out is a bit daunting.  react-native-device-info is one of a number dependencies that I'll need to sort out.

peacechen:
Created PR #324 .  Let me know if you need any more information.

machour:
After reviewing the PR, I must say I'm really doubtful about merging this PR ..

Wouldn't it be possible to find a solution within your app codebase instead?
  Check this article part: https://medium.com/@mitchoneill/react-native-for-web-16c5c96f7165#994a

peacechen:
It may be necessary to split the app up into native and non-native pieces.  Bundling the RN JS for web is problematic imports (and CJS require) are done statically.  Even though the RNDeviceInfo implementation is require'd conditionally, the bundler grabs both which then pukes on the native references.

It would be ideal if RNDeviceInfo could handle this internally, but I don't have a solution for that.

machour:
What are the errors being thrown? 🤔 

If anything we could check if `Platform.OS === "web"` and bail out in `deviceinfo.js` as a courtesy, but adding an almost empty & inaccurate implementation definitely sounds like a big no-no. 
(especially since it would add ~25% of red crosses in a matrix we'd like to keep as green as possible 😄 )

peacechen:
[...]

PR #339, opened 2018-02-25
feat: add empty polyfill for react-native-web

machour:
## Description

Implemented support for react-native-web (#323). We now provide an empty polyfill.


peacechen:
It would be handy if userAgent returned accurate information since it's applicable for web.
```
userAgent: window.navigator.userAgent
```

Issue #331, opened 2018-02-21
getFreeDiskStorage returns negative number


L3V147H4N:
## Summary
Happens on multiple devices
![image](https://user-images.githubusercontent.com/5846479/36487176-0dba5fa6-16ff-11e8-9967-f5811c5a1635.png)

Windows 10 PRO

## Current behavior

Returns negative number for Free Disk Storage

## Expected behavior

Returns number of bytes for Free Disk Storage

machour:
Weird, here's the code again for this: https://github.com/rebeccahughes/react-native-device-info/blob/master/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L144

I'm wondering if we're not overflowing the Integer here.. what's the actual free disk space on the affected devices?

L3V147H4N:
sadly I don't have access to those devices, they are used by employees of my client and we are not in the same city so I would not be able to tell, I have a few devices of my own I'll test to see if the same bug appears in one of them

L3V147H4N:
Perhaps tomorrow I could get some more information

DracoBlue:
I think a long should be ok, too.

DracoBlue:
A possible solution could be to add a new function, which returns the storage in mb with everything after the `.` beeing the kb.

So we get rid of the long/int and have float.

E.g: 1200000 becomes 1.2

winadiw:
i also got that problem when getting `getTotalDiskCapacity()`, how to solve this problem?

jamninetyfive:
any new？getTotalDiskCapacity thx@machour

hannigand:
Any progress on this? `getFreeDiskStorage` and `getTotalDiskCapacity` are both returning negative numbers.

Device: OnePlus 3
OS: OxygenOS Version 5.0.5
Android Version: 8.0.0

katieisnell:
I am also having this issue with both `getFreeDiskStorage` and `getTotalDiskCapacity` on emulators and my phone

Device: Pixel 2
Android version: 8.1.0

and also on

Device: OnePlus 3T
OS: OxygenOS 4.5.1
Android vers: 7.1.1

My Pixel is 1GB with 100MB external storage but `getFreeDiskStorage` gives me 576454656
and `getTotalDiskCapacity` is -1694711808.

jamninetyfive:
> 任何新的？thx @machour

jamninetyfive:
@knot123 thx

Issue #346, opened 2018-03-05
getTotalDiskCapacity() do not get the same value as ”System Setting“ In android


jcliang1212:
In Android, getTotalDiskCapacity() do not get the right value。 So I change the Java code as follow:
```java
public Long getTotalDiskCapacity() {
    try {
      StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
      return root.getBlockCountLong() * root.getBlockSizeLong();
    } catch (Exception e) {
      e.printStackTrace();
    }
    return null;
  }
``` 

But the function returns 4.3GB, but the System Setting show 8.0GB. The Android device is "SM-T719C - 6.0.1". Can anyone tell me why? Thanks! @crafterm @doochik @DracoBlue @stevemoser @jeveloper

machour:
Should be fixed by #597

mikehardy:
@machour this one is actually not fixed - I fixed integer overflow with a very focused change, and this one will need a separate change to use different native APIs in a thoughtful way. I didn't attempt it in the integer overflow fix

mikehardy:
If someone wants to submit a PR, that would be great. A good example of the task is in this stack overflow thread: https://stackoverflow.com/a/30401010

Basically, you have to get *all* the external storage directories and get all of their sizes. You should probably also get the internal storage, take care to de-duplicate all of them and that will be your total capacity (and total free space if you call that native API) as well

goislimat:
_I know this issue is opened for a while but I have a possible solution to present._

I'm working on an app that does a check on the total disk capacity. This function works fine for iOS, but on Android, I'm getting values that doesn't seem right.

This function is currently implemented as follows:

```java
  @ReactMethod
  public BigInteger getTotalDiskCapacity() {
    try {
      StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
      return BigInteger.valueOf(root.getBlockCount()).multiply(BigInteger.valueOf(root.getBlockSize()));
    } catch (Exception e) {
      e.printStackTrace();
    }
    return null;
  }
```

When I execute the `getTotalDiskCapacity()` function on my Android (**Xiaomi Mi A2 Lite with 32 GB of disk**) I get 2,600,857,600 bytes (~ 2,6 GB) as result.

[...]

Issue #338, opened 2018-02-24
getFreeDiskStorage does not return the same value as system setting in iOS


jcliang1212:
when in iOS, use the express “getFreeDiskStorage()/1000000000” get 9.94GB, the "System Setting" show 9.73GB, and  use the express “getTotalDiskCapacity()/1000000000” get 13.23GB, the "System Setting" show 13.23GB. So, why getFreeDiskStorage() does not match the value of "System Setting" in iOS? Please help me, thanks. @crafterm @doochik @DracoBlue @stevemoser @jeveloper

DracoBlue:
Hey! The value is a integer in java and a floating point „number“ in JavaScript. Since this is the case, you don’t get 100% accuracy for each digit.

A possible solution could be to add a new function, which returns the storage in kb or mb with everything after the `.` beeing the kb/bytes.

E.g: 1200000 becomes 1.2

jcliang1212:
@DracoBlue , thanks for your answer. But why the getTotalDiskCapacity() matches the value of "System Setting" in iOS? And getFreeDiskStorage() is bigger than the value of "System Setting" in iOS in some iPad and is smaller than the value of "System Setting" in iOS on other iPad. I do not think it is the problem of different data type between JS and other languages.

jcliang1212:
@DracoBlue , this problem does not happen in iOS simulator.

DracoBlue:
This seems to be an issue with the way how iOS retrieves the amount of free disk space.

Can you try one of the alternative ways in this stackoverflow question: https://stackoverflow.com/questions/11717304/why-does-the-value-from-nsfilesystemfreesize-differ-from-the-free-size-reported ?

jcliang1212:
@DracoBlue ，I agree with you, thanks! But I still do not find the solution. Here is another question: In Android, getTotalDiskCapacity()  do not get the right value, because "Integer" is not big enough, so I change the returned type of getTotalDiskCapacity() to Long, but It still not that right. The system setting show the total disk is 8.0GB, but I get 4.30 GB.

PR #350, opened 2018-03-06
adapt iOS 11 and Android 19+ to get total space

genglei01:
## Description
1.Implemented to get the disk free space for iOS11.
2.Implemented to get the disk total space for Android 19+.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -81,32 +81,28 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   private Boolean isEmulator() {
-    return Build.FINGERPRINT.startsWith("generic")
-        || Build.FINGERPRINT.startsWith("unknown")
-        || Build.MODEL.contains("google_sdk")
-        || Build.MODEL.contains("Emulator")
-        || Build.MODEL.contains("Android SDK built for x86")
-        || Build.MANUFACTURER.contains("Genymotion")
+    return Build.FINGERPRINT.startsWith("generic") || Build.FINGERPRINT.startsWith("unknown")
+        || Build.MODEL.contains("google_sdk") || Build.MODEL.contains("Emulator")
+        || Build.MODEL.contains("Android SDK built for x86") || Build.MANUFACTURER.contains("Genymotion")
         || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
         || "google_sdk".equals(Build.PRODUCT);
   }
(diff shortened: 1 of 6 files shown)


machour:
Thank you so much for this! Does it fix #331 #338 #346 ?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -140,21 +149,35 @@ public String getCarrier() {
...
+        File[] files = this.reactContext.getExternalFilesDirs(Environment.MEDIA_MOUNTED);
+        for (File file : files) {
+          totalSize += this.getCapacity(file);
+        }
+      }else {
+        totalSize += this.getCapacity(Environment.getExternalStorageDirectory());
+      }
+      totalSize += this.getCapacity(Environment.getDownloadCacheDirectory());
+      totalSize += this.getCapacity(Environment.getRootDirectory());
+      totalSize += this.getCapacity(Environment.getDataDirectory());
```
machour: Tested this code with my OnePlus5T (64G internal memory, no external memory) this returns `172 612 931 584`, which is incorrect.

My Android storage screen reports 56.55GB.

From Android docs for Environment.getExternalStorageDirectory():
> Note: don't be confused by the word "external" here. This directory can better be thought as media/shared storage. It is a filesystem that can hold a relatively large amount of data and that is shared across all applications (does not enforce permissions). Traditionally this is an SD card, but it may also be implemented as built-in storage in a device that is distinct from the protected internal storage and can be mounted as a filesystem on a computer.

So adding external storage to download & root & data is not always the correct way to go as they are sometime the same. Do you think that we could add a test for the sd card presence and act accordingly?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -90,22 +91,22 @@ private Boolean isEmulator() {
...
+        if (metrics.densityDpi == DisplayMetrics.DENSITY_DEFAULT
             || metrics.densityDpi == DisplayMetrics.DENSITY_HIGH
             || metrics.densityDpi == DisplayMetrics.DENSITY_MEDIUM
             || metrics.densityDpi == DisplayMetrics.DENSITY_TV
             || metrics.densityDpi == DisplayMetrics.DENSITY_XHIGH) {
-      return true;
+            return true;
+        }
+        return false;
     }
```
machour: Please revert this changes

PR #356, opened 2018-03-14
Added support for `getUsedMemory` to get the app memory usage

IjzerenHein:
## Description

Added `getUsedMemory()` that allows for getting the current memory usage of the App. This feature returns the same value as you would otherwise see in the react-native Perf Monitor overlay.

This feature has been battle tested on both Android & iOS in our App for about 5 months now. Works like a charm, no side effects, never saw any problems.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,7 @@ var DeviceInfo = require('react-native-device-info');
 | [getTotalDiskCapacity()](#gettotaldiskcapacity)   | `number`            |  ✅  |   ✅    |   ❌    | 0.15.0 |
 | [getTotalMemory()](#gettotalmemory)               | `number`            |  ✅  |   ✅    |   ❌    | 0.14.0 |
 | [getUniqueID()](#getuniqueid)                     | `string`            |  ✅  |   ✅    |   ✅    | ?      |
+| [getUsedMemory()](#getusedmemory)                 | `Promise<number>`   |  ✅  |   ✅    |   ❌    | 0.18.0 |
 | [getUserAgent()](#getuseragent)                   | `string`            |  ✅  |   ✅    |   ❌    | 0.7.0  |
 | [getVersion()](#getversion)                       | `string`            |  ✅  |   ✅    |   ✅    | ?      |
 | [is24Hour()](#is24hour)                           | `boolean`           |  ✅  |   ✅    |   ✅    | 0.13.0 |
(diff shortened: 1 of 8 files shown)


IjzerenHein:
@machour I've added the web polyfill!

IjzerenHein:
@machour Okay I fixed the merge issues

machour:
Oh thank you, I was currently doing that :)

IjzerenHein:
Oops :)

machour:
I did some testing on Android & iOS:

screen shot 2018-03-14 at 8 02 36 pm

~10 mo on android vs 146 mo on iOS 😮 is that normal?

I also compared the output for Android vs the output of adb : 

screen shot 2018-03-14 at 7 58 12 pm

The total of "Private Dirty" seems to be more or less the number we're getting.. do you know what to compare here?

I also wanted to compare memory usage in Xcode vs what the app reports, but was unsuccessful, could you try on your side and do some comparisons?

README.md
```diff
@@ -739,6 +740,20 @@ const uniqueId = DeviceInfo.getUniqueID();
 
 ---
 
+### getUsedMemory()
+
+Gets the apps memory usage, in bytes.
```
machour: is that app's or applicationS ?

mikehardy:
Hi there! @IjzerenHein I'm willing to patch this up and merge it but I have a question - on Android at least without knowing maximum available heap size, what question does this feature answer? By which I mean, we'll know what we're using but we can't take any actions based on it (to prevent OOM or similar) unless we also get the max heap size with Runtime.maxMemory(). I'm not sure if iOS has a similar API

Is it useful to add this feature if no action can be taken, or do we need the partner maxMemory API as well so we can calculate available memory?

mikehardy:
examine react-native-vitals

IjzerenHein:
@machour  @mikehardy  I added this feature, not specifically to take action on, but for diagnostics. The app I was working on was quite memory intensive (recording thousands of sensor samples per second) and we were seeing issues on certain Android phones. I leave the usability of this metric to the one using it. Obviously, memory usage in VM extended environments, with addition of Garbage Collection, is not so straight forward.
When I did notice back then, that when the mem was low, the app would crash. So this was indeed what was happening on these phones and this metric helped identify that. Eventually it helped us identify that even though phones had maybe 4Gb of memory, the app was only able to use a fraction (256mb) of that on certain phones.

mikehardy:
That's a good defense of the proposed API, thank you @IjzerenHein - I separately (with @cristianoccazinsp ) inquired on react-native-vitals if that repo was indeed vital, as it does not seem to be. Pending results of whether they'll be a good host for mem usage APIs this can go in. I'm leaning towards including it here as that repo does not seem very active

ChrisEelmaa:
+1 to get this in to react-native-device-info :)

We're going to use this metric for two reasons:

1) Attach it to the log entries, having high memory usage is good sign that something is going wrong.

2) It will be available for testers to see. Testers will generally test the app and ensure there are no memory leaks after using app for 20-30 minutes (uploading pictures, etc).

mikehardy:
Enough time has already passed in the related repo between request and any answer that I have mentally marked the repository as "unresponsive", so these APIs will come here as I do 2.0.

I am no fan of endless talk on mythical "next major versions" but it is open source so hopefully people have patience, and I think taking the time to make sure the supported API surface area fits well is as important as eventual code...

mikehardy:
I have this implemented on the v3 branch verbatim now - it was too hard to resolve the conflicts so I copied it over and tested it but the work is pure @IjzerenHein  - thank you!

IjzerenHein:
Awesome @mikehardy :)

Issue #355, opened 2018-03-12
[Discussion] `getUniqueID` for Android should use another implementation


SandroMachado:
`getUniqueID` function for android returns the constant `ANDROID_ID`. Based on an official [post](https://android-developers.googleblog.com/2011/03/identifying-app-installations.html) from Google at the android developers blog, the `ANDROID_ID` is not reliable.

> ANDROID_ID seems a good choice for a unique device identifier. There are downsides: First, it is not 100% reliable on releases of Android prior to 2.2 (“Froyo”). Also, there has been at least one widely-observed bug in a popular handset from a major manufacturer, where every instance has the same ANDROID_ID.

In another [document](https://developer.android.com/training/articles/user-data-ids.html) from google, they suggest to identify the app create a UUID and store it at the internal storage.

> In cases where an Instance ID isn't practical, custom globally unique IDs (GUIDs) can also be used to uniquely identify an app instance. The simplest way to do so is by generating your own GUID using the following code.

They also recommend to not use the `ANDROID_ID`.

> #1: Avoid using hardware identifiers. Hardware identifiers such as SSAID (Android ID) and IMEI can be avoided in most use-cases without limiting required functionality.

My suggestion is change the `getUniqueID` function to return a previously generate `UUID`. If there is no one available, generate it, store it and return that value. Basically is to implement the google's suggestion.

WDYT? I am offering to implement this.

\cc @machour 

Refs: https://github.com/rebeccahughes/react-native-device-info/issues/60

machour:
@SandroMachado thank you so much for bringing this up. In fact, I was thinking of moving getUniqueID() from native to the javascript land and use the same implementation for the three platforms (we could also implement it for react-native-web if needed).

As this is a BC breaking changes, this will have to be shipped in 1.0.0. Have a look at #307 for the rest of the plans and don't hesitate to comment there and give more ideas!

SandroMachado:
I think that we should keep this on native side, to take the advantage of the apple/iOS api, that is really well designed.

On Android, like I said before, I can make a PR to update this API to match the iOS behaviour (removing the vendor logic) since they have an API to generate UUID (`UUID.randomUUID().toString();`). WDYT?

Also, to generate a UUID on JS side, from my search, I didn't find any solution that does not relays on the random string generation method, that is not the most desirable solution since it can lead to collisions and that should not happens.

machour:
I totally agree. Feel free to work on a PR for the implementation whenever you want :)

[...]

Issue #60, opened 2016-07-05
Support instance ID on iOS (this package currently has a java impl for android)


mehcode:
From what I understand the purpose is an ID that is specific to the install of an application.

``` js
  async getInstanceID() {
    // // Check for an existing ID
    let id = await AsyncStorage.getItem("instance-id");
    if (id != null) return id;

    // Generate a new UUID
    id = UUID();

    // Save the ID
    await AsyncStorage.setItem("instance-id", id);

    return id;
  }
```

That should do it across all platforms. Don't have time to make it look nice / work so I'm opening the issue.

danielfx90:
I see that now Google provides support for the InstanceID on iOS as well:
https://developers.google.com/instance-id/guides/ios-implementation

The library could use that in the `getInstanceID()` method.

mikehardy:
As part of the deprecation of GCM #602 I'd like to convert to app-specific/app-local GUIDs as described here with the sample implementation above. Seems preferable to bringing in even a reflective / optional Firebase dependency to replace GCM

PR #358, opened 2018-03-14
Refactor `getUniqueID()` implementation for Android

SandroMachado:
## Description

Fixed issue #355. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |
| Windows |    ❌     |

## Checklist

Before:
screen shot 2018-03-14 at 12 26 57

After:
screen shot 2018-03-14 at 12 36 23

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,14 +698,14 @@ Gets the device unique ID.
 const uniqueId = DeviceInfo.getUniqueID();
 
 // iOS: "FCDBD8EF-62FC-4ECB-B2F5-92C9E79AC7F9"
-// Android: "dd96dec43fb81c97"
+// Android: "8DCFDA40-704E-4693-82E8-82722F45B6EB"
 // Windows: ?
 ```
 
 **Notes**
 
 > * iOS: This is [`IDFV`](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) so it will change if all apps from the current apps vendor have been previously uninstalled.
-> * android: Prior to Oreo, this id ([ANDROID_ID](https://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID)) will always be the same once you set up your phone.
+> * android: This will get a UUID for this installation. If the user removes and installs the app, a new UUID will be returned. We followed the approach recommended by goole [here](https://developer.android.com/training/articles/user-data-ids.html) at `#3`.
 ---
 
 ### getUserAgent()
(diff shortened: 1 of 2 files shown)


ramyyahari:
@machour Do you know when you guys will be releasing this?

machour:
@ramyyahari I've been pretty busy lately with other projects, so this is not really planned. 

We could however introduce this as getUuid() for the current version.. what do you think about it @sandromachado?

SandroMachado:
@machour from my point of view, since the library is not yet in the version `1.0`, we can introduce BC in minor versions (SEMVER `MAJOR.MINOR.PATCH`). We just need to be transparent and include the BC in the `changelog`.

machour:
To me, even a 0.x version should not contain BC breaks. If only to avoid pissing off users and finding myself with a myriad of issues to answer.

Could you update both PRs to introduce new methods instead? :pray:

akiradeveloper:
Yes. I hope this change is merged immediately with different method name as it doesn't break bc. Otherwise applications can't upgrade the library version if they expect the ID is stable across versions. Since this library is starred over 2000, it's not reasonable to ignore existing users even if its version states it is still in dev.

SandroMachado:
@machour to do this, how do you suggest to rename the methods? 

Since the BC is only on android and the method name should be equal in android and iOS.

machour:
@SandroMachado how does getUuid() sound?

vomchik:
@SandroMachado What happens when I reinstall the app? Do I get new id or the same?

vomchik:
@machour or `getUUID()`

paour:
@vomchik I'm also interested in a persistent (best-effort) unique ID that works on Android and iOS.

mangei:
any updates on this? I need this, on order to recognize test devices for Android. the current version of getUniqueId returns an empty string on my device (OnePlus 3)

schumannd:
best practice is to use `getUuid` instead of `getUUID`. Explanation [here](https://stackoverflow.com/questions/15526107/acronyms-in-camelcase)

mikehardy:
Tagging this for #307 v2 but will likely do something pure JS since it's possible, similar to the comment in #60

Husseinhj:
It's not a unique id. It depends on your usage or the goal of a unique id.
What is mean usage or the goal of a unique id?

There are two things for unique id:
1) Uniqueness in the application scope.
2) Uniqueness in device scope.

The application scope just for user application and important to this application check device is unique or not by storing data in `Keystore` (android) and `keychain` (iOS).

The device scope is important for other SDKs and attributions services such as the Adjust, Branch, Appslfayer, and so on.
By using `Advertising id`, you can unique device and understandable string for all applications and SDKs to do special marketing things for this exact device based on user behaviors.

Issue #352, opened 2018-03-09
CI build hangs during pod installation after adding rndi.


IljaDaderko:
## Summary
Hey guys, I added `react-native-device-info` to my project, some really useful data, thank you! I stumbled upon this issue however, this library adds following pod `'RNDeviceInfo', :path => '../node_modules/react-native-device-info'` for some reason it also relies on `React` pod?

When I'm running my build in CI, my `pod install` step hangs at this bit

```
Installing RNDeviceInfo (0.17.4)
Installing React (0.11.0)
```

Any ideas on the fix? Why are we installing React pod by the way? It was deprecated.

|             |     |
| ----------- | --- |
| Version     | ^0.17.4   |
| Affected OS | iOS  |
| OS Version  | 10   |

## Current behavior
pod install hangs.

## Expected behavior
Should install pods without any issues

machour:
Hi @IljaDaderko, never used pods so I'm not sure what to do here.
Would you like to open a PR to remove the React pod if that's the problem?

IljaDaderko:
@machour sorry for late response, I attempted it here
https://github.com/rebeccahughes/react-native-device-info/pull/361

Tested locally seems to be working fine, but same as you my knowledge of cocoapod space is limited, I'm not sure why it was needed as dependency in this projects podspec, but `React` pod was deprecated looooong ago and it is not recommended at all to use it.

rangav:
```
Pod::Spec.new do |s|
  s.name         = "RNDeviceInfo"
  s.version      = "0.21.5"
  s.summary      = "Device Information for react-native"

  s.homepage     = "https://github.com/rebeccahughes/react-native-device-info"

  s.license      = "MIT"
  s.authors      = { "Rebecca Hughes" => "rebecca@learnium.net" }
  s.ios.deployment_target = '8.0'
  s.tvos.deployment_target = '10.0'

  s.source       = { :git => "https://github.com/rebeccahughes/react-native-device-info.git" }

  s.source_files  = "ios/RNDeviceInfo/*.{h,m}"

  s.dependency 'React'
end
```

please remove dependency of react -  still appears - is it not removed yet?

machour:
Hi, please don't comment on closed issues.

Issue #63, opened 2016-07-07
React Pod has been deprecated


emmby:

maciekiks:
workaround:
add to Podfile:
 pod 'React', :path => '../node_modules/react-native'

koenpunt:
Is the Pod even necessary?

jwrubel:
In addition to the pods @koenpunt mentions, pod yoga needs to be lower case:

`pod 'yoga', path: '../node_modules/react-native/ReactCommon/yoga'`

[This page](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html) from React Native may help.

Zephir77167:
Have had a lot of problems since adding this workaround in my podfile. Why exactly does this module has React as a cocoapod dependency?

jwrubel:
React isn't a dependency of this pod, it's a dependency of React Native. I have found problems with a number of different libraries when I started using cocoapods. I saw on another issue list to add the following at  the end of your `Podfile`:

```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    if target.name == "React"
      target.remove_from_project
    end
  end
end
```
Once I added this post_install hook, everything got much more stable.

machour:
@valerybugakov could you provide a PR for this?

machour:
Fixed in 0.21.1

wangwcq:
pod 'React', path: '../node_modules/react-native'
pod 'yoga', path: '../node_modules/react-native/ReactCommon/yoga'

For the pod 'yoga', use lower case package name or pod couldn't find podspec file.

Issue #38, opened 2016-04-16
Naming collision when React Cocoapod is added


joanbarros:
I'm currently having an issue where the same file is being added two times to the project (Xcode).

The issue is with the `polyfills.js` file but it is adding React again. All the files added are already in the project added by `npm`.

```
⋊> ~/D/P/QueueProject on master ⨯ react-native start                              21:25:13
 ┌────────────────────────────────────────────────────────────────────────────┐
 │  Running packager on port 8081.                                            │
 │                                                                            │
 │  Keep this packager running while developing on any JS projects. Feel      │
 │  free to close this tab and run your own packager instance if you          │
 │  prefer.                                                                   │
 │                                                                            │
 │  https://github.com/facebook/react-native                                  │
 │                                                                            │
 └────────────────────────────────────────────────────────────────────────────┘
Looking for JS files in
   /Users/joanbarros/Development/Projects/QueueProject

[10:21:11 PM] <START> Building Dependency Graph
[10:21:11 PM] <START> Crawling File System
[Hot Module Replacement] Server listening on /hot

React packager ready.

[10:21:12 PM] <END>   Crawling File System (1152ms)
[10:21:12 PM] <START> Building in-memory fs for JavaScript
[10:21:12 PM] <END>   Building in-memory fs for JavaScript (640ms)
[10:21:12 PM] <START> Building in-memory fs for Assets
[10:21:13 PM] <END>   Building in-memory fs for Assets (365ms)
[10:21:13 PM] <START> Building Haste Map
[10:21:13 PM] <START> Building (deprecated) Asset Map
[10:21:13 PM] <END>   Building (deprecated) Asset Map (68ms)
Failed to build DependencyGraph: Naming collision detected: /Users/joanbarros/Development/Projects/QueueProject/ios/Pods/React/packager/react-packager/src/DependencyResolver/polyfills/polyfills.js collides with /Users/joanbarros/Development/Projects/QueueProject/node_modules/react-native/packager/react-packager/src/Resolver/polyfills/polyfills.js
Error: Naming collision detected: /Users/joanbarros/Development/Projects/QueueProject/ios/Pods/React/packager/react-packager/src/DependencyResolver/polyfills/polyfills.js collides with /Users/joanbarros/Development/Projects/QueueProject/node_modules/react-native/packager/react-packager/src/Resolver/polyfills/polyfills.js
    at HasteMap._updateHasteMap (HasteMap.js:132:13)
[...]

PR #361, opened 2018-03-17
fix: remove deprecated React pod dependency

IljaDaderko:
## Description

Fixed issue #352

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [x] I added the documentation in `README.md`. (Not needed)

### DIFF
--- a/RNDeviceInfo.podspec
+++ b/RNDeviceInfo.podspec
@@ -13,6 +13,5 @@ Pod::Spec.new do |s|
   s.source       = { :git => "https://github.com/rebeccahughes/react-native-device-info.git" }
 
   s.source_files  = "RNDeviceInfo/*.{h,m}"
-
-  s.dependency 'React'
+  
 end


machour:
Thank you @IljaDaderko !

IljaDaderko:
@machour sick! I will work with it latter today, will let you know if I notice something weird.

IljaDaderko:
@machour I ran into an issue with fresh implementation

`RCTBridgeModule` is not defined and as far as I can see it is checking that it is initially in React pod. Any ideas for fix here? It has fallback to just include this file, but it is not inside pod directory.

screen shot 2018-03-17 at 14 57 31

IljaDaderko:
Going deeper into this, I don't think pod approach is working, I linked everything manually following "ios without cocoapods" guide and it works fine. I think "ios with pods" needs to be updated, especially since it prob wasn't touched for a while? (judging by left in React pod).

Some observations

1) Installing pod doesn't link it to project `Libraries`
2) Pod is installed as `Development Pod`

I think its worth to re-open this issue.

koenpunt:
Since the `React` pod has no version constraint in the podspec, it will rely on the latest version available, which will be the React version added to your project. Removing this dependency will make it unable to resolve the `RCTBridgeModule.h`, like @IljaDaderko [already mentioned](https://github.com/rebeccahughes/react-native-device-info/pull/361#issuecomment-373918914).
I think this change should be reverted.

IljaDaderko:
Agreed, I probably had this cached when I was testing in my project hence, didn't catch it. However, adding React pod back in completely hands pod install (or takes more than 5 min to resolve React pod, I waited this long and gave up).

machour:
@IljaDaderko add --verbose when using `pod install`, this will shade some lights on what's going on. It was a pretty huge download on my side.

@koenpunt I downloaded CocoaPods and tried adding this project on a fresh crna app. From what I'm understanding from the [docs](https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#configuring-cocoapods-dependencies), the reference to React should be made in the application Podfile itself. 

RNDeviceInfo would resolve whatever React is installed by the developer.

Tried this on my end, but I'm facing some unrelated errors from Yoga. Could you confirm / correct me if I'm wrong?

koenpunt:
> RNDeviceInfo would resolve whatever React is installed by the developer.

True, but from my experience; if the dependency is not specified in the podspec, the generated project will not include a header search for the React files, so the dependency declaration is still necessary.

machour:
/summon @GantMan 

Hey Gant, could you shade some lights on this PR and what should be done with pods?

Some people are complaining about having React (deprecated) as a dependency in our podspec. (#38, #63, #352)

I thought that sticking to Facebook recommended way should be fine, but it seems that it would cause problems.

fjtrujy:
Hello,

I'm pretty sure that the React dependency should be kept.
It is mandatory to be used with Cocoapods.
I don't understand why there are people complaining about this dependency, they said that this is deprecated... but I don't find the official documentation where they said about how to make it work with Cocoapod and react-native.

Can you help me with this?

Thanks

machour:
@fjtrujy Here's the documentation I found that lead me to believe that the React pod must be required from the application `Podfile`, not libraries podspecs: https://facebook.github.io/react-native/docs/integration-with-existing-apps.html#configuring-cocoapods-dependencies

I do however understand that dropping this dependency is problematic, and my knowledge is really limited on the subject as I don't use pods at all. 

I think I'll simply rollback this change, as having a warning about deprecation is far less problematic than having problems compiling your app.

PS: for now, you can stick with 0.21.0 until this gets sorted out.

machour:
Really curious about what you have done for the web platform, I was under the impression that we would only be able to provide an empty polyfill, by all means, go ahead! 🙌 

No need to PR the rollback, just waiting to see if @GantMan have the time to have a look before changing that one line.

fjtrujy:
Actually is not just about web, is about native snapshots using jest (that are the end runs over a browser), right now the project is not compatible with this.

Thanks

fjtrujy:
I have created the PR for the snapshots.

Thansk!

https://github.com/rebeccahughes/react-native-device-info/pull/375

EyMaddis:
This rollback issue now installs React (again) in an older version when using it in a project that does not use cocoapods for every dependency.

PR #362, opened 2018-03-18
check to detect is iphone x of not

fattahmuhyiddeen:
## Description

Added `isIphoneX()` that allows to check if user are using iphoneX

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS          |    ✅     |
| Android   |    ✅     |
| Windows |    ✅    |

## Checklist

* [x] I mentionned this change in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Release Notes
 
 ### next
-
+* add method to check if iphoneX
 ### 0.21.1
 
 * Remove deprecated React pod dependency (https://github.com/rebeccahughes/react-native-device-info/pull/361)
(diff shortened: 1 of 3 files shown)


machour:
Hi and thank you for your proposal. 
Have you tried using getModel() and encountered problems?

DeviceInfo.getModel() === 'iPhone X';

fattahmuhyiddeen:
yes I already use it. currently from my forked version

machour:
isIphoneX() is really easy to replicate in your application using:

```
const isIphoneX = () => DeviceInfo.getModel() === 'iPhone X';
```

Adding it to this library doesn't really make sense to me (or else we will end up with isSamsungS8(), isOnePlus5t(), etc..).

fattahmuhyiddeen:
it make sense because iphonex has notch and i really need make my app
header taller just for that iphone to avoid that notch

fattahmuhyiddeen:
starting april, apple force all app to compatible with iphone x ui

machour:
I hear you, but reac-native-device-info is about pulling native information from devices, in the most generic / cross-platform way possible.

Adding isIphoneX() would break that promise, and we already provide `getModel()` that allow you to quickly do it in your app.

fattahmuhyiddeen:
ok

machour:
Please feel free to submit other PRs at any time 🙌

PR #364, opened 2018-03-20
Return error on appVersion if package not found

RalfNieuwenhuizen:
## Description

Instead of the default `not-available`, return `error` when a package can not be found (which should not happen). Better for debugging purposes.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -115,7 +115,6 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     String packageName = this.reactContext.getPackageName();
 
     constants.put("appVersion", "not available");
-    constants.put("buildVersion", "not available");
     constants.put("buildNumber", 0);
 
     try {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -115,7 +115,6 @@ public void getMacAddress(Promise p) {
     String packageName = this.reactContext.getPackageName();
 
     constants.put("appVersion", "not available");
-    constants.put("buildVersion", "not available");
```
machour: Why was this line removed?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -125,6 +124,7 @@ public void getMacAddress(Promise p) {
       constants.put("firstInstallTime", info.firstInstallTime);
       constants.put("lastUpdateTime", info.lastUpdateTime);
     } catch (PackageManager.NameNotFoundException e) {
+      constants.put("appVersion", "error");
```
machour: This sadly causes a BC break for people relying on the string "not available" being returned upon error :/

I'll try to introduce a nicer debug mode for 1.0.0 (#307)

RalfNieuwenhuizen:
@machour you're probably right, the fix is too specific for my use case :)

machour:
@RalfNieuwenhuizen no problems mate :)

If you have any suggestions about how error reporting/hinting could be enhanced for the next BC breaking release, I'm all ears!

Issue #366, opened 2018-03-21
Unable to build - more than one library with package name 'com.google.android.gms.license'


GaudamThiyagarajan:
## Current behavior
react-native run-android

:react-native-device-info:processReleaseManifest
:react-native-device-info:processReleaseResources FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-device-info:processReleaseResources'.
> Error: more than one library with package name 'com.google.android.gms.license'

BUILD FAILED

Total time: 42.342 secs

## Expected behavior

It was working fine the previous day. no changes was made to build.gradle files

## dependecies

```
dependencies {
    compile project(':react-native-splash-screen')
    compile project(':react-native-svg')
    compile project(':react-native-linear-gradient')
    compile 'com.facebook.fresco:fresco:1.5.0'
    compile 'com.facebook.fresco:animated-gif:1.5.0'
    compile project(':react-native-fetch-blob')
    compile group: 'net.lingala.zip4j', name: 'zip4j', version: '1.3.2'
    compile project(':appcenter-crashes')
    compile project(':appcenter-analytics')
    compile project(':appcenter')
    compile project(':react-native-code-push')
    compile project(':react-native-sentry')
    compile 'com.amplitude:android-sdk:2.13.4'
    compile project(':react-native-amplitude-analytics')
    compile project(':react-native-fbsdk')
    compile(project(":react-native-google-signin")){
      exclude group: "com.google.android.gms" // very important
    }
    compile 'com.google.android.gms:play-services-auth:9.2.1' // should be at least 9.0.0
    compile project(':react-native-orientation-locker')
    compile(project(':react-native-device-info')) {
      exclude group: "com.google.android.gms" // very important
    }
    compile project(':react-native-vector-icons')
    compile fileTree(dir: "libs", include: ["*.jar"])
    compile "com.android.support:appcompat-v7:27.0.2"
    compile "com.facebook.react:react-native:+"  // From node_modules
}
```

yangtaufoo:
Please refer to #365

GaudamThiyagarajan:
allprojects {
    repositories {
        configurations.all {
            resolutionStrategy {
                force "com.google.android.gms:play-services-gcm:11.8.0"
            }
        }
        ....
   }
}

adding this to my build.gradle file fixed the issue.

> note: I had to add  force "com.google.android.gms:play-services-auth:11.8.0" as i was using google sign in.

[...]

Issue #365, opened 2018-03-20
Compatibility with react-native-firebase


ricardofavero:
| Version     |  0.21.1  |
| Affected OS | Android |
| OS Version  | 7.0 |

Hi,
I'm trying to use this library and react-native-firebase in my project.
But when I try to compile, I get this error

```
* What went wrong:
Execution failed for task ':react-native-device-info:processReleaseResources'.
> Error: more than one library with package name 'com.google.android.gms.license'
```

If I uninstall any of the 2, then the other one works.

Does anyone know how to fix it?

machour:
Hey, did you check the troubleshooting section in the Readme?

On Mar 20, 2018 22:49, "Ricardo Favero" <notifications@github.com> wrote:

| Version | 0.21.1 |
| Affected OS | Android |
| OS Version | 7.0 |

Hi,
I'm trying to use this library and react-native-firebase in my project.
But when I try to compile, I get this error

* What went wrong:
Execution failed for task
':react-native-device-info:processReleaseResources'.> Error: more than
one library with package name 'com.google.android.gms.license'

Does anyone know how to fix it?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/rebeccahughes/react-native-device-info/issues/365>, or mute
the thread

rgoldiez:
I'm running into the same issue.  I followed the troubleshooting to no avail.  I'm also use react-native-firebase

```
    compile(project(':react-native-device-info')) {
      exclude group: 'com.google.android.gms'    // this did not fix my issue
    }

```

CodyBrouwers:
Seeing the same issue, but am **not** using react-native-firebase. Tried adding the following to no avail:

```java
    compile(project(':react-native-device-info')) {
        exclude group: 'com.google.android.gms'
    }
```

jurajkrivda:
Add googlePlayServicesVersion = "xx.xxx" to your project gradle. For example:

```
ext {
    compileSdkVersion = 26
    buildToolsVersion = '26.0.1'
    googlePlayServicesVersion = "11.6.0"
}
```

CodyBrouwers:
@jurajkrivda Thanks so much for pointing me in the right direction, all I ended up needing to add was this line to my `android/gradle.properties` file:

```
googlePlayServicesVersion=11.0.4
```

I also didn't end up needing to exclude the `com.google.android.gms` group.

bdlee-1:
[...]

PR #368, opened 2018-03-22
compilation error

hirthbrian:
Hello,

I don't know if it's a problem for everyone but there is a compilation error on Android with the updated Google Play services.

Prevent Error: more than one library with package name 'com.google.android.gms.license' with Google Play services 12.0.0

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'com.android.library'
 def DEFAULT_COMPILE_SDK_VERSION             = 23
 def DEFAULT_BUILD_TOOLS_VERSION             = "25.0.2"
 def DEFAULT_TARGET_SDK_VERSION              = 22
-def DEFAULT_GOOGLE_PLAY_SERVICES_VERSION    = "+"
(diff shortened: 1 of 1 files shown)


jcodlingcc:
I can confirm this is a problem for us as well.

machour:
Hi @hirthbrian and thank you for the proposal.

#365 and #366 were opened recently, due to the new version of Google Play services.

I don't know if setting a default value to "11.+" really solves this problem, as some people may want to use "12" for a project, and this is just a default value that can be overridden by configuration.

Would you be up for adding more details about this problem in the README file along with some of the solutions provided in the two issues I linked?

jcodlingcc:
We found a solution:
Set the following in your `android/build.gradle` file:

`ext {
    googlePlayServicesVersion = "11.4.2"
}`

Make sure to use the latest version of `react-native-device-info`.

Therefor this PR is not needed.

hirthbrian:
Yes @jcodlingcc this is the best solution, thanks for the help!

Issue #353, opened 2018-03-12
iOS automatic linking is not working. 


rohitgoyal:
## Summary

Version: 0.17.4 

## Current behavior

RNDeviceInfo automatic linking fails on iOS while it shows success on terminal.
## Expected behavior

RNDeviceInfo should be present in Libraries in Xcode.

machour:
Hi, what's the command you're using?

dougjedwards:
I'm having exactly the same issue.

I'm using:
react-native link react-native-device-info

SandroMachado:
Same here!

machour:
Just tested this on a ejected CRNA, the library gets correctly linked for me:

<img width="275" alt="screen shot 2018-03-12 at 5 42 12 pm" src="https://user-images.githubusercontent.com/304450/37296982-cfa90686-261c-11e8-94df-cd9a8dd2be8a.png">

If however, you're facing the following error : `ld: library not found for -lRNDeviceInfo-tvOS`
You can manually delete libRNDeviceInfo-tvOS.a in Xcode -> [Your iOS build target] -> Build Phrases -> Link Binary with Libraries. It's a react-native link bug.

SandroMachado:
My project uses `lerna` and so on, so, probably can be related to that.

schumannd:
@machour i am getting that exact error. Shouldn't this be handled by `react-native link`?

machour:
@schumannd it currently works for me so I'm not seeing the problem.. could you investigate a bit?

tamhv:
```
1. yarn add react-native-device-info
2. react-native link react-native-device-info
3. react-native run-ios
4. console.log("DEVICE NAME",DeviceInfo.getDeviceName())
```
I got js error `TypeError: undefined is not an object (evaluating 'RNDeviceInfo.deviceName')` 

I notice there is a line `pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'` in my Podfile after i ran `react-native link` 

Do i need to `cd ios && pod install` ?

Actually i've tried to pod install
```
# cd ios
# pod install
Analyzing dependencies
Fetching podspec for `RNDeviceInfo` from `../node_modules/react-native-device-info`
Downloading dependencies
...
Installing RNDeviceInfo (0.21.2)
...
Generating Pods project
Integrating client project
Sending stats
Pod installation complete! There are 7 dependencies from the Podfile and 14 total pods installed.
```
Then xcode can't build, error 
```
[...]

PR #374, opened 2018-03-25
fix: Moved the IOS code to an ios/ directory

rhdeck:
## Description
Fixed issue #353 by moving the IOS code to a more-easily-findable location
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

* [X] I have tested this on a device/simulator for each compatible OS
* [ ] I added the documentation in `README.md`. (bug fix - not go in?)
* [ ] I mentioned this change in `CHANGELOG.md`. (bug fixes go here?)

PR #376, opened 2018-03-26
fix: Move .podspec to root

sdovenor:
## Description

Fixes issues #353 #374
The changes in #374 cause pod install to fail if you currently have 
```
pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'
```
which is most likely the majority of projects and is the current recommendation in the README.  This fix allows that to continue working and still get the fix from #374 for automatic linking.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist


machour:
Thank you, didn't think to check that

PR #382, opened 2018-04-04
Uses this.reactContext instead of getCurrentActivity to check permissions for phoneNumber getter

rcorrie:
Uses this.reactContext instead of getCurrentActivity to check permissions for phoneNumber getter

## Description

getCurrentActivity would return null very often, and would render getPhoneNumber function useless.
Using this.reactContext to check the permissions is more reliable.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Next
  * fix: Add OnePlus A6010 to devicesWithNothc list (https://github.com/react-native-community/react-native-device-info/pull/604)
+ * fix: use reactContext vs getApplicationContext() (https://github.com/react-native-community/react-native-device-info/pull/382)
 
 ### 1.1.0
  * Add `supportedABIs()` (https://github.com/react-native-community/react-native-device-info/pull/598)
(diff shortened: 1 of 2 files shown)


machour:
Thank you! I'll try it out and publish a new version asap

mikehardy:
@rcorrie are you still using this in production? It looks like a clean change and I can test it locally and I could try to get it merged if you are still using it and can say it works for you

mikehardy:
I researched this. I think it's a correct change, getCurrentActivity() has many cases where it might be null: https://stackoverflow.com/questions/45435450/getcurrentactivity-in-reactcontextbasejavamodule-returns-null (with 2 linked github issue comments) and this https://github.com/facebook/react-native/issues/18345

My take-away is that getCurrentActivity() is best avoided if possible, and in this case we just need the context (which we have) so using it directly is safest.

I patched up the changelog to be current (and to practice maintainer skills with git amendments / rebases etc) so I think this looks good now

@machour does this look good to you?

mikehardy:
Thanks for this @rcorrie !

mikehardy:
It's out!

https://www.npmjs.com/package/react-native-device-info/v/1.2.0

rcorrie:
Thank you @mikehardy and yes we have been using this in production.

Issue #396, opened 2018-05-03
getBatteryLevel() always returns -1 using real iOS device


anwarhamr:
## Summary

|             |     |
| ----------- | --- |
| Version     | 21.5   |
| Affected OS | iOS  |
| OS Version  |  11.3 |

when getBatteryLevel promise resolves the value is always -1.

machour:
Waiting for your input on #404

mikehardy:
#404 just merged, but note @hickeyadamc it was mostly a documentation update. You need to add an enabling calling in your AppDelegate, the API docs point out what to do

@machour my favorite -> should close, already merged

PR #404, opened 2018-05-14
Enabled battery monitoring mode to get actual battery data

Elf2707:
## Description

Fixed issue #396

Enabled battery monitoring mode to get actual battery data

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+++ b/ios/RNDeviceInfo.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
--- a/README.md
+++ b/README.md
@@ -283,8 +283,11 @@ DeviceInfo.getBatteryLevel().then(batteryLevel => {
 
 **Notes**
 
+> To be able to get actual battery level enable battery monitoring mode for application
+add this code:
+  [UIDevice currentDevice].batteryMonitoringEnabled = true;
+to AppDelegate.m application:didFinishLaunchingWithOptions:
 > Returns -1 on the iOS Simulator
-
 ---
 
 ### getBrand()
--- a/README.md
+++ b/README.md
@@ -285,9 +285,10 @@ DeviceInfo.getBatteryLevel().then(batteryLevel => {
 
 > To be able to get actual battery level enable battery monitoring mode for application.
 > Add this code:
->
->    [UIDevice currentDevice].batteryMonitoringEnabled = true;
-> 
+
+```objective-c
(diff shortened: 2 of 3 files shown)


Elf2707:
I've been thinking. Maybe a better way not to enable batteryMonitoringMode in each getBatteryLevel call like I did in this PR, but just leave this method as it is right now and just add a hint to documentation to put this line to AppDelegate.m application:didFinishLaunchingWithOptions to get battery level:

[UIDevice currentDevice].batteryMonitoringEnabled = true;

machour:
@Elf2707 Thank you so much.
@anwarhamr could you take a look at this PR, and Elf2707's last comment and give us some feedback?

anwarhamr:
Great that did the trick for and iPhone7.  Could this be something that is handled by the `react-native link` command?
On a side note, how do I see or know when the next release will be available?

machour:
I'm not sure if `link` can handle this, but we would definitely add this to the README. 

@elf2707, as soon as you update the PR accordingly, I'll publish a new release. Thank you again!

machour:
@anwarhamr, you'll definitely get a notification when github when I merge this one, and usually I notify all participants when I publish something they helped fixing. Thank you again for helping out!

Elf2707:
Hi everyone. I changed getBatteryLevel back and added note to getBatteryLevel description. Its a right way to put [UIDevice currentDevice].batteryMonitoringEnabled = true; to AppDelegate. Because it should be called once per application not every getBatteryLevel call. Thanks.

ios/RNDeviceInfo.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
machour: Is this file needed?

Elf2707: Actually according to https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html

>Xcode 9.3 adds a new IDEWorkspaceChecks.plist file to a workspace's shared data, to store >the state of necessary workspace checks. Committing this file to source control will prevent >unnecessary rerunning of those checks for each user opening the workspace. (37293167)

So I could delete it, but as I understand it is useful file. Should I delete it anyway?

hickeyadamc:
I'm also having the issue where getBatteryLevel returns -1. Was this change ever merged in?

ios/RNDeviceInfo.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
mikehardy: Looks harmless and I confirm other high-quality modules have it

mikehardy:
@machour I think this is good to merge. A valid documentation update and the other included change appears fine. You can put a "postlink" (and "postunlink") script in package.json to have link handle this, but altering the internals of a method in AppDelegate seems hard and potentially brittle. I think the documentation pointer is sufficient

machour:
react-native-device-info 0.27.1 published 🎉

mikehardy:
> react-native-device-info 0.27.1 published 🎉

Boom! cleared a lot out in one go, awesome. I'll keep an ear out for issues

PR #408, opened 2018-05-22
Use RCTSharedApplication instead of [UIApplication sharedApplication]

### LINKED ISSUE #347: Make it build with iOS App Extensions
## Summary

|             |     |
| ----------- | --- |
| Version     | 17.3   |
| Affected OS | ios   |
| OS Version  | 11.2   |

## Current behavior

When using `react-native-device-info` in an iOS App Extension it fails to build because sharedApplication is referenced once `RNDeviceInfo.m:207`:

`
'sharedApplication' is unavailable: not available on iOS (App Extension) - Use view controller based solutions where appropriate instead.
`

## Expected behavior

Would love for it to be able to build. And since it's only one line I'm guessing it could be potentially simple. Not an Objective-C pro though - so I don't know how! :)

lnikkila:
## Description

Fixes #347.

Calling `[UIApplication sharedApplication]` [is unsafe in an app extension][0]. This is hidden by `RCTSharedApplication` which lets app extensions include this library through CocoaPods [without compilation errors][1].

[0]: https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW6
[1]: https://github.com/CocoaPods/CocoaPods/pull/3048

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |
| Windows |    N/A     |

## Checklist

  >I don’t think this is necessary for this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Notes
 
 ### 1.4.2:
+ * fix: Use `RCTSharedApplication` so compile works for ios app extensions (https://github.com/react-native-community/react-native-device-info/pull/408)
  * chore: Add 3rd generation iPad pro to device/model list (https://github.com/react-native-community/react-native-device-info/pull/618)
  * feat: Support `getUserAgent()` on old androids (API level <= 16) (https://github.com/react-native-community/react-native-device-info/pull/545)
  * chore: Add Huweai INE-LX1 to devices with notch (https://github.com/react-native-community/react-native-device-info/pull/624)
@@ -119,10 +120,6 @@
 ### 0.22.0
 
 * Add support for `getIpAddress` and `getMacAddress` on iOS (https://github.com/react-native-community/react-native-device-info/commit/41735bd0b2efe1f626afc066604f27073acb9d4c)
-
-* Use `RCTSharedApplication` which hides [UIApplication sharedApplication] so
-  iOS app extensions can include the library through CocoaPods without
-  compilation errors.
 
 ### 0.21.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,10 @@
 
 * Add support for `getIpAddress` and `getMacAddress` on iOS (https://github.com/react-native-community/react-native-device-info/commit/41735bd0b2efe1f626afc066604f27073acb9d4c)
 
+* Use `RCTSharedApplication` which hides [UIApplication sharedApplication] so
+  iOS app extensions can include the library through CocoaPods without
+  compilation errors.
+
 ### 0.21.5
 
 * Rolled back the Pod change made in 0.21.1
--- a/README.md
+++ b/README.md
@@ -478,6 +478,8 @@ Gets the device font scale.
 The font scale is the ratio of the current system font to the "normal" font size, so if normal text is 10pt and the system font is currently 15pt, the font scale would be 1.5
 This can be used to determine if accessability settings has been changed for the device; you may want to re-layout certain views if the font scale is significantly larger ( > 2.0 )
 
+In iOS App Extensions this call always returns 1.0, see #625.
+
 **Examples**
 
 ```js
--- a/ios/RNDeviceInfo/RNDeviceInfo.h
+++ b/ios/RNDeviceInfo/RNDeviceInfo.h
@@ -8,16 +8,9 @@
 
 #import <UIKit/UIKit.h>
 #import <sys/utsname.h>
-
-#if __has_include(<React/RCTAssert.h>)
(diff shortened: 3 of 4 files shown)


dgritsko:
Any update on this? Having the same problem and would love to see this PR merged.

danalloway:
@machour let us know how we can help get this PR merged

alexisbronchart:
RCTSharedApplication() returns nil when ran in an app extension.
I suggest using the following:
`NSString *contentSize = [[UIScreen mainScreen] traitCollection].preferredContentSizeCategory;`

However it would imply dropping iOS 9 support. Or maybe we could just default the contentSize to 1 on that version?

mikehardy:
Hey everyone, I can merge things so I can help. Thanks for your patience!

The change looks clean enough but I'm a novice at iOS / CocoaPods etc, so I'll need some testing confirmation that it works.

Responding to @alexisbronchart - if React Native supports an API, we should support it, so dropping support for iOS 9 is not okay. 

Other than that, if the interested parties here can agree on the best way for it to build and be useful on iOS 9-current, and rebase+ modify the changelog entry to be whatever it is now +1 on the patch level, I can get this in.

lnikkila:
>RCTSharedApplication() returns nil when ran in an app extension.

True, this method checks for that and currently returns 1.0 in an app extension. At the time I wrote this I was only interested in getting the library to compile (we weren’t using this particular method) which is why I didn’t attempt returning the correct value.

>However it would imply dropping iOS 9 support. Or maybe we could just default the contentSize to 1 on that version?

Yeah, using `[UIScreen mainScreen]` on iOS 10+ and returning 1.0 on older versions is a good idea. However I think this should be implemented as a separate PR as I feel getting the library to compile in an app extension is a separate step from achieving 100% feature parity in an app extension.

I understand if people disagree with this point of view, but I’m afraid I don’t currently have the time to update this PR further aside from rebasing it to a mergeable state. If someone volunteers I’d be happy to hand this over.

>The change looks clean enough but I'm a novice at iOS / CocoaPods etc, so I'll need some testing confirmation that it works.

For what it’s worth, we’ve been using an internal fork with this patch in production for about a year across multiple apps, no problems so far.

mikehardy:
Thanks a lot for getting back to me. Let me make sure I understand everything -

> > RCTSharedApplication() returns nil when ran in an app extension.
> 
> True, this method checks for that and currently returns 1.0 in an app extension. At the time I wrote this I was only interested in getting the library to compile (we weren’t using this particular method) which is why I didn’t attempt returning the correct value.

I definitely get that!

> > However it would imply dropping iOS 9 support. Or maybe we could just default the contentSize to 1 on that version?
> 
> Yeah, using `[UIScreen mainScreen]` on iOS 10+ and returning 1.0 on older versions is a good idea. However I think this should be implemented as a separate PR as I feel getting the library to compile in an app extension is a separate step from achieving 100% feature parity in an app extension.

I'm okay with this if you can open a separate issue pointing here, so any interested parties can pick it up easily. An update to this getFontScale() API description in the docs that has a really short warning about how it behaves in App Extensions might save some future person a lot of gray hair too, while being trivial to do

> I understand if people disagree with this point of view, but I’m afraid I don’t currently have the time to update this PR further aside from rebasing it to a mergeable state. If someone volunteers I’d be happy to hand this over.

I totally understand time pressure, plus this is pretty stale through no fault of your own 

> For what it’s worth, we’ve been using an internal fork with this patch in production for about a year across multiple apps, no problems so far.

This is worth a lot! And it would be nice to have another involved user back on the latest release with this in.

So - a quick note in the API docs, rebase, and add a changelog entry with the next patch version and I'll merge this? Thanks again!

lnikkila:
>So - a quick note in the API docs, fresh issue for interested future people, rebase, and add a changelog entry with the next patch version and I'll merge this? Thanks again!

Sure, sounds good. I’ll get these changes in sometime next week.

mikehardy:
@lnikkila - I am almost done with this - I just patched it up and I've got the branch pulled locally for testing before merge + release. If you could test it (I just rebased + pushed) on your end as well that would give me more confidence, but I wanted to get this done this morning :-)

lnikkila:
Ah I noticed, I had forgotten about this. I created #625 and added it into the README. I’ll test this in an app today.

mikehardy:
Okay, after my changes I pulled your branch locally and tested it. I can't vouch for app extensions, but getFontScale() is still correctly working in my emulator after the changes when I twiddle the font setting in the accessibility menu.

Thanks for the contribution!

Issue #410, opened 2018-05-25
getSystemName reports iPhone OS on iOS 9 and older devices


kfoxb:
## Summary

|             |     |
| ----------- | --- |
| Version     | 0.12.0+   |
| Affected OS | iOS   |
| OS Version  | <=9   |

## Current behavior

Calling getSystemName on an iOS 9 or older devices reports "iPhone OS". It would seem that according to this stack overflow answer, https://stackoverflow.com/a/41569148,  iOS 10 devices could also be affected, however, I have not personally experienced that.

## Expected behavior

The README implies that getSystemName will always return "iOS" when running on an iOS device. Either the docs should be changed to show that "iPhone OS" is a possible value that could be returned, or change the method to only return "iOS".

machour:
Hi and thank you for logging this one. 

Provided that Platform.os already exist in React Native, I'd be more prone to leave the method as it is, and update the README. 

Adding ', iPhone OS' after the 'iOS' bit in the example seems sufficient to me to highlight that. 
We could also link to Platform.os for people who are looking for a solid way to identify the platform 

What do you think about it? Do you want to make a PR for this?

PR #411, opened 2018-05-25
docs: clarify iOS return values of getSystemName

kfoxb:
## Description
This commit will clarify in the README that both "iOS" and "iPhone OS" are possible return values from calling getSystemName when running on an iOS device. "iOS" being reported on most iOS 10+ devices and "iPhone OS" reported on iOS 9 and less and some iOS 10 devices.

Fixed issue #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,7 +656,7 @@ Gets the device OS name.
 ```js
 const systemName = DeviceInfo.getSystemName();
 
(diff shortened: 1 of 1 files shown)


machour:
Thank you so much @kfoxb and sorry for the delays :D

PR #435, opened 2018-06-15
Export device info

finnp:
## Description

All the getters fail if there is no `RNDeviceInfo`. I added an export called `data` that exports the `deviceInfo`.

This way people can do checks like: `deviceInfo.data ? deviceInfo.getUserAgent() : 'default-agent'`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -10,6 +10,7 @@ if (!RNDeviceInfo && Platform.OS === 'web') {
 }
 
 module.exports = {
+  data: RNDeviceInfo,
   getUniqueID: function() {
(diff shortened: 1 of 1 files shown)


machour:
Hi @finnp, I'm failing to see the scenario where RNDeviceInfo wouldn't be available, could you elaborate?

finnp:
Never mind, it was a colleague who complained about this. But I don't really remember the context.

PR #344, opened 2018-03-04
feat: add getInstallReferrer() for Android

poison:
## Description

Added `getInstallReferrer()` that returns the install referrer attribution when your android app gets installed (com.android.vending.INSTALL_REFERRER). You can use https://play.google.com/store/apps/details?id=com.myapp&referrer=<your custom referrer string here>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |   ❌     |
| Android |    ✅    |
| Windows |   ❌     |

## Checklist

* [X] I have tested this on a device/simulator for each compatible OS
* [X] I added the documentation in `README.md`.
* [ ] I mentionned this change in `CHANGELOG.md`.

PR #440, opened 2018-06-20
Add android get installer package name

TingC:
## Description

Added `getInstallerPackageName()` that retrieve the package name of the application that installed a package.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### next
 
+* Add `getInstallerPackageName`
+
 ### 0.21.5
 
 * Rolled back the Pod change made in 0.21.1
--- a/README.md
+++ b/README.md
@@ -829,7 +829,7 @@ const isTablet = DeviceInfo.isTablet(); // true
 
 ### getInstallerPackageName()
 
-Retrieve the package name of the application that installed a package..
+Retrieve the package name of the application that installed a package.
 
 **Examples**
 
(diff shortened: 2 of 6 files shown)


mikehardy:
I'd like to stay focused on device information and not worry about packages in this module.

I'm going to close this, however I maintain a different module which can do what you need, here's the exact line even:
https://github.com/mikehardy/react-native-update-apk/blob/master/example/App.js#L170

That module isn't unique either, there are a few that handle all the package details for android

mjgallag:
@mikehardy what's the difference is between including https://github.com/react-native-community/react-native-device-info/pull/344 but not including this?

/cc @TingC

mikehardy:
The difference I was not maintainer then ;-). And what I've seen since taking over the package is that we can try to be all things to all people and not do everything well, or we can stay focused strictly on device info and be excellent at it, while letting other packages that are already doing related work be excellent at that. And I believe there are already packages that handle APK-related things well, like one I also work on, and linked above :man_shrugging: - so, not being obstructive, just trying not to bloat out the package and suffocate under the weight of it

mjgallag:
@mikehardy makes sense to me, thanks for the quick clarification. I just realized #344 was over a year older than this, 2018 not 2019, when I posted I thought it had merged just two weeks before :)

PR #442, opened 2018-06-25
Fixed #260 & changed getUserAgent() return type

### LINKED ISSUE #260: Required dispatch_sync to load constants for RNDeviceInfo. This may lead to deadlocks
I'm getting this warning on 0.49.5
`Required dispatch_sync to load constants for RNDeviceInfo. This may lead to deadlocks`

Everything seems to function fine, but I'm wondering what does this exactly mean?

psegalen:
## Description

Fixes issue #260

/!\ Fix requires to change `getUserAgent()` return type (it was a `string` now it's a `Promise<string>`)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

> I'll let @machour modify the changelog because I don't know how he wants to inform everyone about the breaking change

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -32,7 +32,7 @@ RCT_EXPORT_MODULE(RNDeviceInfo)
 
 + (BOOL)requiresMainQueueSetup
 {
-   return NO;
+   return YES;
 }
 
 - (NSString*) deviceId
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -314,11 +314,17 @@ RCT_EXPORT_METHOD(getUserAgent:(RCTPromiseResolveBlock)resolve rejecter:(RCTProm
         }
         [_webView evaluateJavaScript:@"navigator.userAgent" completionHandler:^(id result, NSError * error) {
             if (error == nil) {
-                if (result != nil) {
-                    resolve([NSString stringWithFormat:@"%@", result]);
+                NSString * returnedValue = @"";
+                if (result != nil && [result isKindOfClass:[NSString class]]) {
+                    returnedValue = result;
                 }
+                dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                    resolve(returnedValue);
+                });
             } else {
-                reject(@"evaluate_js_failed", [NSString stringWithFormat:@"evaluateJavaScript error : %@", error.localizedDescription], error);
+                dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+                    reject(@"evaluate_js_failed", [NSString stringWithFormat:@"evaluateJavaScript error : %@", error.localizedDescription], error);
+                });
             }
         }];
     });
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -8,6 +8,7 @@
 
 #import "RNDeviceInfo.h"
 #import "DeviceUID.h"
+#import <WebKit/WebKit.h>
 #if !(TARGET_OS_TV)
 #import <LocalAuthentication/LocalAuthentication.h>
 #endif
@@ -21,6 +22,9 @@
 #endif
 
 @implementation RNDeviceInfo
+{
+    WKWebView *_webView;
+}
 
 @synthesize isEmulator;
 
@@ -28,9 +32,8 @@ RCT_EXPORT_MODULE(RNDeviceInfo)
 
 + (BOOL)requiresMainQueueSetup
 {
-   return YES;
+   return NO;
 }
-
 
 - (NSString*) deviceId
 {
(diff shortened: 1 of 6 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -314,4 +306,22 @@ - (NSDictionary *)constantsToExport
...
+RCT_EXPORT_METHOD(getUserAgent:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (_webView == nil) {
+            _webView = [[WKWebView alloc] initWithFrame:CGRectZero];
+        }
+        [_webView evaluateJavaScript:@"navigator.userAgent" completionHandler:^(id result, NSError * error) {
+            if (error == nil) {
+                if (result != nil) {
+                    resolve([NSString stringWithFormat:@"%@", result]);
```
maxkomarychev: [docs](https://developer.apple.com/documentation/webkit/wkwebview/1415017-evaluatejavascript) say:

> [...] The completion handler always runs on the main thread.

perhaps it's better to resolve this not on main thread?
can we utilize `self.methodQueue`?

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -314,4 +306,22 @@ - (NSDictionary *)constantsToExport
...
+                }
+            } else {
+                reject(@"evaluate_js_failed", [NSString stringWithFormat:@"evaluateJavaScript error : %@", error.localizedDescription], error);
```
maxkomarychev: same as above

ios/RNDeviceInfo/RNDeviceInfo.m
maxkomarychev: is `result` not a string?

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -314,4 +306,22 @@ - (NSDictionary *)constantsToExport
...
 
+RCT_EXPORT_METHOD(getUserAgent:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (_webView == nil) {
+            _webView = [[WKWebView alloc] initWithFrame:CGRectZero];
+        }
+        [_webView evaluateJavaScript:@"navigator.userAgent" completionHandler:^(id result, NSError * error) {
+            if (error == nil) {
+                if (result != nil) {
```
maxkomarychev: This looks like promise will never be resolved with if `result == nil`.
Can we do better like:

```
if (result) {
  resolve([NSString stringWithFormat:@"%@", result]); // still not sure if this is necessary but I guess docs don't guarantee type of the result...
} else {
  resolve(nil)
}
```
makes sense?

ios/RNDeviceInfo/RNDeviceInfo.m
psegalen: Implicit cast can be dangerous, I'll add a Type check

psegalen: I didn't create a `methodQueue` only for that so I dispatch on global priority default queue like shown in https://facebook.github.io/react-native/docs/native-modules-ios.html#threading

ios/RNDeviceInfo/RNDeviceInfo.m
maxkomarychev: I saw `resolve(nil)` in many libs, but that's not too critical :)

ishabo:
@psegalen can you please resolve the conflicts?

psegalen:
@ishabo done, sorry for the lag!

SudoPlz:
Any news with that?

mikehardy:
Linking this to the #307 v2.0 issue where the main goal will be to Promise-ify the API, since this appears to be tested code.

Worth noting that iOS doesn't crash with master right now, using at least react-native 0.58+ (I did not test 0.57)

rajivshah3:
@psegalen would you mind resolving the conflicts again? Hopefully a maintainer will be able to review this PR soon, as it should improve launch performance on apps that use this library

mikehardy:
The PR is reviewed, it is the direction we will go (so you can feel comfortable using this style), but it is a large breaking change so it hasn't happened yet

web/index.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
   buildNumber: 0,
   appVersion: 0,
   deviceName: '',
-  userAgent: window.navigator.userAgent,
+  userAgent: () => new Promise((resolve, reject) => resolve(window.navigator.userAgent)),
```
LinusU: This could be just `() => Promise.resolve(window.navigator.userAgent)` which avoids creating an extra closure :)

mikehardy:
I'm not too bothered by the conflicts, and this is still the direction I want to go. In the meantime the whole AndroidX/jetifier thing happened though which unfortunately absorbed the time I would have had for the Promise API switch over. It will happen eventually though, and I appreciate the patience. Also if anyone else just couldn't take it anymore and converted everything to promises in a PR that'd be a good way to get it done ;-)

ldco2016:
Any solution to this issue that got locked?
https://github.com/react-native-community/react-native-device-info/issues/260

I get a yellow warning with `Required dispatch_sync to load constants for RNDeviceInfo. This may lead to deadlocks.`

I am using `"react-native-device-info": "2.2.2",` with `"react-native": "0.60.4",`

mikehardy:
No deep solution, currently just ignore. "may lead to deadlocks", but does not in practice. A future version of react-native-device-info could solve it but it requires the conversion to promise etc that has been pending but was delayed by me focusing on the whole RN60/jetifier thing, and no one else submitting PRs 🤷‍♂

ldco2016:
@mikehardy , thanks for clarifying and getting back to me. I understand about the whole jetifier androidx thing, was upgrading a project when all that happened.

mikehardy:
Obsoleted by #757 as I shape up a v3 but this work (esp the java part and docs and js impl) will be informed by it - thank you, and thanks for your patience

PR #453, opened 2018-07-20
Gradle 3.+ support and improve code

KarlosQ:
* Gradle 3.+ support
* RN 0.56.+ compile version
* Improve project-wide properties implementation

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,17 +1,16 @@
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 apply plugin: 'com.android.library'
 
-def DEFAULT_COMPILE_SDK_VERSION             = 23
-def DEFAULT_BUILD_TOOLS_VERSION             = "25.0.2"
-def DEFAULT_TARGET_SDK_VERSION              = 22
-def DEFAULT_GOOGLE_PLAY_SERVICES_VERSION    = "+"
-
 android {
-    compileSdkVersion project.hasProperty('compileSdkVersion') ? project.compileSdkVersion : DEFAULT_COMPILE_SDK_VERSION
-    buildToolsVersion project.hasProperty('buildToolsVersion') ? project.buildToolsVersion : DEFAULT_BUILD_TOOLS_VERSION
+    compileSdkVersion safeExtGet('compileSdkVersion', 26)
+    buildToolsVersion safeExtGet('buildToolsVersion', '26.0.3')
 
-    defaultConfig {
-        minSdkVersion 16
-        targetSdkVersion project.hasProperty('targetSdkVersion') ? project.targetSdkVersion : DEFAULT_TARGET_SDK_VERSION
+  defaultConfig {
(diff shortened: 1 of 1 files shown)


iggyfisk:
If you don't want the Gradle 3 support and general changes, at least consider bumping targetSdk to 26 for RN 0.56 support

machour:
Hi @KarlosQ, just tried this PR on a working RN app, compiling android gives the following error:

```
yarn run v1.5.1
$ react-native run-android
Scanning folders for symlinks in /Users/didou/js/react-native/react-native-device-info-test/node_modules (22ms)
JS server already running.
Building and installing the app on the device (cd android && ./gradlew installDebug)...

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/didou/js/react-native/react-native-device-info-test/node_modules/react-native-device-info/android/build.gradle' line: 26

* What went wrong:
A problem occurred evaluating project ':react-native-device-info'.
> Could not find method implementation() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 4.572 secs
Could not install the app on the device, read the error above for details.
Make sure you have an Android emulator running or a device connected and have
set up your Android development environment:
https://facebook.github.io/react-native/docs/android-setup.html
✨  Done in 6.90s.
```

Could you tell me what the problem is?

andreyslyusar:
@machour @KarlosQ I try this PR and it work in our project in both debug and release mode. This pull request preferred over #454, as it allow to override `compileSdkVersion`, etc from bundle.gradle
```
ext {
    compileSdkVersion   = 26
    buildToolsVersion   = "26.0.2"
    targetSdkVersion    = 26
}
```
and the same patter is used in many other react-native libraries.

fiznool:
@machour The reason for the failure is that the top-level `android/build.gradle` file in the RN project needs to define the following:

```
dependencies {
        classpath 'com.android.tools.build:gradle:3.1.4'
}
```

Currently, a RN 0.56 project scaffolds this as `com.android.tools.build:gradle:2.3.3`. Gradle 3 is required to use `implementation` instead of `compile` when defining dependencies, which is why this PR (which changes `compile` to `implementation`) will fail if using Gradle 2.

It's worth noting that RN 0.57 will [scaffold projects to use Gradle 3 by default](https://github.com/facebook/react-native/commit/6eac2d4e2f5f697043b495002872bfac3e8595cb), so when this is out of RC, users who scaffold a RN 0.57 app and use this library will see a warning:

```
Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
```

It's probably therefore worth getting this merged in when 0.57 lands. Although, of course, it will be a breaking change for users who install this library with 0.56 or below, as they won't be able to compile their project without updating their Gradle to v3.

mikehardy:
@machour this is obsolete - it has a better implementation of safe version fetching but the existing one works, and the rest has either been done (change from compile to implementation) or will be in #548

Issue #403, opened 2018-05-14
eslint-plugin-import error


jay1337:
## Summary

|             |     |
| ----------- | --- |
| Version     | `react-native-device-info@0.21.5`   |
| Affected OS | NA   |
| OS Version  | NA   |

## Current behavior

`eslint-plugin-import@2.11.0` detects an error when importing the module
`import DeviceInfo from 'react-native-device-info';`

## Expected behavior

No error

## Why

I think that `eslint-plugin-import` finds an `import` statement in `deviceinfo.js` so it expects that the module uses `export` statements. `deviceinfo.js` should use only `import` and `export` or only `require(...)` and `module.exports`.

machour:
Hi and thank you for logging this. 
Could you share the actual error message?

jay1337:
This rule is triggered : https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/default.md

`No default export found in module. (import/default)`

machour:
@peacechen could you possibly have a look at this?

peacechen:
Thanks @jay1337 for posting the error details.

Based on the eslint error, changing this line should fix it:
https://github.com/rebeccahughes/react-native-device-info/blob/master/deviceinfo.js#L12

From
```
module.exports = {
```

To
```
export {
```

@jay1337   Would you mind trying that locally?  Edit `node_modules/react-native-device-info/deviceinfo.js` and try building again.

jay1337:
@peacechen you have to use `export default {`, it solves eslint issue.

schumannd:
Will this fixed be merged in the future?

machour:
We were just missing a PR, and here you are :)

machour:
0.22.3 is out with this fix, thank you all!

machour:
This seem to be causing troubles: #468 
/cc @jay1337 @peacechen

schumannd:
I did not encounter these issues. Are they fixed by using `import DeviceInfo from 'react-native-device-info';` instead of other forms of importing this package?

eshikerya:
@schumannd nope. starting from 0.22.2 it is broken. any type of import give an error

schumannd:
Yes, I should have tested alternative ways of importing the module as well. Maybe we can find a way to enable both old and new import styles?

Maybe something like:
```
const obj = {
  ...
}
export default obj;
module.exports = obj;
```
?

What import styles do break right now?

peacechen:
[...]

PR #466, opened 2018-08-07
#403: Fix eslint-plugin-import error

schumannd:
## Description

Fixed issue #403

## Compatibility

N/A

## Checklist

I don't believe this fix needs to be mentioned anywhere

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -9,7 +9,7 @@ if (!RNDeviceInfo && Platform.OS === 'web') {
   RNDeviceInfo = require('./web');
 }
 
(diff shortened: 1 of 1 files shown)


machour:
Thank you so much!

danilobuerger:
@machour @schumannd This breaks importing via Typescript.

Before this PR importing was supposed to be:
```
import * as DeviceInfo from "react-native-device-info";
```
Because of this PR it has to be 
```
import DeviceInfo from "react-native-device-info";
```
which doesn't compile as it doesn't match the typings.

PR #474, opened 2018-08-09
feat: Add getCameraPresence function

Hirurgo:
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ?     |
| Android |    ✅     |
| Windows |    ✅     |


Hirurgo:
This function was made in response to the fact that react-native-camera does not have such functionality, I will try in the future to add an implementation to Android

machour:
Hi ! Would it be possible to distinguish between front and rear camera?

Hirurgo:
@machour Yes. Some part of device have **EnclosureLocation** property, in which the location is indicated (Back, Bottom, Front, Left...), but this method gives an answer of whether there is at least some kind of camera at all.

 
Note that this is the UWP platform and it also works on static computers that can have a usb cameras, without dividing into front and rear. Also there are android devices without a camera but the ability to connect it through the usb port. We have one such. 

P.S. Now I'm working on implementing the same function for android, for our project.
The presence of such a function for ios does not make sense.

Hirurgo:
Hooray! Now for Android 😊

machour:
Awesome, I'll have a look as soon as possible.
On iOS, we should return `true` so that developers don't have to test for the OS

Hirurgo:
https://noncam.com/ - Perhaps it's worth implementing this function for iOS  =)

Hirurgo:
https://stackoverflow.com/questions/3643949/detect-existence-of-camera-in-iphone-app

Hirurgo:
@machour What is the status of this PR? If there are problems, let me know about them

Hirurgo:
@machour If you have a lot of free time, it will be great if you look at this PR =)

mikehardy:
Oh - when I resolved conflicts I pushed right on to your repository. You will likely need to git fetch origin && git rebase (maybe even git rebase --hard) in order to get in sync. That may or may not have broken git-etiquette for you, if so I am sorry but I am trying to help get this merged so apologies in advance :-)

mikehardy:
This is now very stale and I haven't seen a response to my previous questions from @Hirurgo - I just invested some time resolving conflicts and I'm willing to merge this. At this point, as the project has moved forward we need an update illustrating usage for App.js (trivial to add) and an entry for the default/empty polyfill (also trivial) along with the changelog entry. All of these are simple so should not be a barrier but if there's no response I will close it without merge- we need responsive authors more than we need more code, sorry :man_shrugging: - would love to get this in though!

Hirurgo:
I'm here  =)  Now I will try to remember what I did a year ago....

Hirurgo:
@mikehardy  still actual for our project, I'll fix everything now

mikehardy:
:-) fantastic! Let's get this thing merged. It's just what I think of as "mopping up" at this point - all the related bits that aren't the actual code. I'll re-check when you've pushed the extra bits

Hirurgo:
@mikehardy   Add entry in App.js and change log. But I don't understand what means "entry for the default/empty polyfill "

Hirurgo:
@mikehardy done

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -198,6 +201,20 @@ public void getIpAddress(Promise p) {
...
   }
+ 
+  @ReactMethod
+  public void getCameraPresence(Promise p) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      CameraManager manager=(CameraManager)getReactApplicationContext().getSystemService(Context.CAMERA_SERVICE);
+      try {
+        p.resolve(manager.getCameraIdList().length > 0);
+      } catch (CameraAccessException e) {
+        p.reject(e);
```
mikehardy: The API doc you added has the .then() case but does not have a .catch() or indicate it could reject with an exception. Could you add a .catch() that says something like // CameraAccessException or similar ?

Hirurgo: done, but.... how do you think about  "always-resolves" and in this line implement samthing like this   
```
  } catch (CameraAccessException e) {
        p.resolve(undefined);
```

and type 
Promise<?boolean>

mikehardy: Good question - I like it like this personally. An exception is an exception to me. This is a place of opinions though - everyone can have a different one but in the end your code will have to have some sort of branching logic to handle it, whichever style. So I'd leave it - you have this code working, you are using it in an app - that deployed experience is more valuable to me than switching a style

mikehardy:
Great stuff, I'll merge this in and make a final comment when released, then if you are using this in your project you can get back to master instead of maintaining a local fork. :fireworks:  

As an advance warning, we are about to do a major API shift in v2.0 but the plan is not to introduce different behavior except where strongly desired and just blocked by existing API. We will mostly just convert to promise (already implemented that way here) and package the APIs a bit. So when it happens it shouldn't be hard to migrate for you at all - hopefully useful info since you will have *just* moved back to master / npmjs.org releases.

Thanks!

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
Hirurgo: =) yes we use this function, but don't catch error)))

Hirurgo: if reject camera permission, this exception does not throw

mikehardy: interesting - it's documented as doing it so ? I dunno - the code seems obviously correct in reading. I'm running the branch now to test it, if it works, I still think this is the right sort of try/catch for this api...

mikehardy:
Great stuff, I'll test this locally then merge this in, then if you are using this in your project you can get back to master instead of maintaining a local fork. As an advance warning, we are about to do a major API shift in v2.0 but the plan is not to introduce different behavior, just convert to promise and package the APIs a bit, so when it happens it shouldn't be hard to migrate at all - hopefully useful info since you will have *just* moved back to master / npmjs.org releases.

Thanks!

mikehardy:
1.7.0 is out, enjoy :rocket:

Issue #505, opened 2018-10-05
Does not support React Native Windows v0.57 


jmcginty:
## React native windows has increased the minimum target version, and this causes compile time errors if associated projects are targeting a lower version. 

|             |     |
| ----------- | --- |
| Version     | 0.22.6   |
| Affected OS | Windows   |
| OS Version  | 10   |

## Current behavior

React Native Windows has changed it's minimum target version to 10.0.14393, so it will not compile.

## Expected behavior

Can compile the project and run

kgstew:
I'm having the same issue. I was able to work around it by updating `<TargetPlatformMinVersion>10.0.10586.0</TargetPlatformMinVersion>` in `RNDeviceInfo.csproj` to 10.0.14393.

machour:
Fixed in 0.24.3 by @jmcginty

Issue #151, opened 2017-02-16
Does not build when running react-native run-windows


jiankai82:
If you add and link this plugin to your Windows UWP react native app and you run react-native run-windows to build the app, you will get an error saying failed to restore Nuget Package.  This is because the project reference to React Native Windows project has incorrect path.  In react-native-device-info/windows/RNDeviceInfo/RNDeviceInfo.csproj, this line is incorrect: `<ProjectReference Include="..\..\node_modules\react-native-windows\ReactWindows\ReactNative\ReactNative.csproj">`

The correct path should be: `<ProjectReference Include="..\..\..\react-native-windows\ReactWindows\ReactNative\ReactNative.csproj">`

This is not a problem if you build your app by opening your app's solution file in Visual Studio and build using Visual Studio.

Can someone please fix this ASAP?  Thanks.

kevinvangelder:
@jiankai82 Did you by chance try to put screenshots? If so, I can't see them. Can you type out the two paths or try to fix the images?

jiankai82:
Hmm....  I did type the paths in my original post.  I think GitHub might have removed it for security purpose.  I'll try again.

Open react-native-device-info/windows/RNDeviceInfo/RNDeviceInfo.csproj
locate line 116
The Include attribute has the incorrect path which is 
..\..\node_modules\react-native-windows\ReactWindows\ReactNative\ReactNative.csproj
It should be:
..\..\..\react-native-windows\ReactWindows\ReactNative\ReactNative.csproj

Please fix it ASAP.  Thanks.

jonathantorley:
Visual Studio automatically fixes this when you try to build your application (it probably just matches the guid of the project and fixes the reference) but that's obviously no good for a command-line build.

It doesn't make sense to change the project to refer to the new path, as then the project won't compile on its own.

If there is a way to correct the path live Visual Studio does in a pre-build step (maybe Visual Studio uses an msbuild target to do the correction that we can reuse) that would be great, but I haven't got the time to look into it.

In the meantime, I've used the below powershell script to correct the issue.
```powershell
# Get the absolute path of the file
$path = Join-Path -Path $pwd -ChildPath 'node_modules\react-native-device-info\windows\RNDeviceInfo\RNDeviceInfo.csproj'

# Read the xml document
$xml = [xml](Get-Content $path)

# Namespaces, because we need it even for the default (and it can't be xmlns)
[...]

PR #507, opened 2018-10-05
Create a new project to fix RNW version compatability and ReactWindows project reference  error

jmcginty:
## React Native Windows has a minimum target version of 10.0.14393, and react native device info needs to be updated or projects will not compile. 

Also, this fixes another existing bug where the reference to the ReactNative project in the csproj was wrong.

Fixed issue #505 Does not support React Native Windows v0.57
Fixed issue #151 Does not build when running react-native run-windows
## Checklist

* [ X] I mentionned this change in `CHANGELOG.md`.


machour:
Hi and sorry for the delays. Could you fix the merge conflicts ?

jmcginty:
Conflicts fixed !

machour:
Thank you so much. Unfortunately my PC which had a RNW setup died on me a while ago, so I cannot test my old project and see how this PR affects it. 

How will things go with people already using this library on react native windows ? Are there breaking changes, or some adjustments to be made to be able to compile?

jmcginty:
On our project, we were already using react-native-device-info, and only encountered this problem while upgrading to RNW 0.57. After changing the package.json file to refer to the pull request project, it works and no source code adjustments were needed. We were able to delete some workarounds for #151, but no changes other than this.

machour:
And for people on RNW < 0.56, how would that go? :thinking:

jmcginty:
We upgraded from RNW 0.50 to 0.57, so it should go fine.

machour:
All right thanks. I'll merge it when I have access to my laptop :)

jmcginty:
Actually, I did some more testing on an older project, and found that because the .csproj file has moved down one level in the directory structure, it would be better if I moved it back up one level and upgrading would be smoother. Please wait until I make this change before merging.

jmcginty:
The .csproj file is in the same location now, so older projects will just find it where they expect it. The only caveat I guess is that now react-native-device-info has a minimum version of 10.0.14393, the host project will have to have a minimum of this version or higher.

machour:
Thank you so much, just added a quick note in the changelog about the new minimal requirement.

0.24.3 has just been published! 🎉

PR #518, opened 2018-10-11
fixed android getFreeDiskStorage and getTotalDiskCapacity got negative number

jamninetyfive:
fixed android getFreeDiskStorage and getTotalDiskCapacity got negative number

## Description

Fixed issue #331 #346 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -38,7 +38,7 @@ import java.util.Map;
 import java.util.TimeZone;
 import java.lang.Runtime;
 import java.net.NetworkInterface;
-
+import java.io.File;
 import javax.annotation.Nullable;
 
 public class RNDeviceModule extends ReactContextBaseJavaModule {
@@ -190,10 +190,13 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
-  public Integer getTotalDiskCapacity() {
+  public long getTotalDiskCapacity() {
     try {
-      StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
-      return root.getBlockCount() * root.getBlockSize();
+      File sdcardDir = Environment.getExternalStorageDirectory();
+      StatFs statFs = new StatFs(sdcardDir.getPath());
+      long blockSize = statFs.getBlockSizeLong();
+      long totalSize = statFs.getBlockCountLong();
+      return blockSize * totalSize;
     } catch (Exception e) {
       e.printStackTrace();
     }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -190,21 +190,27 @@ public String getCarrier() {
   }
 
   @ReactMethod
-  public Integer getTotalDiskCapacity() {
+  public long getTotalDiskCapacity() {
```
greenbue: I think you should use "Long" instead of "long". I am having the same issue and tried using your fix.
Android is complaining with _"incompatible types: <null> cannot be converted to long"_

iShankarG:
I think this fix is not released yet. I'm using it in production app, any update on merge & release or is there any other temporary fix.

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -190,21 +190,27 @@ public String getCarrier() {
...
     try {
-      StatFs root = new StatFs(Environment.getRootDirectory().getAbsolutePath());
-      return root.getBlockCount() * root.getBlockSize();
+      File sdcardDir = Environment.getExternalStorageDirectory();
+      StatFs statFs = new StatFs(sdcardDir.getPath());
+      long blockSize = statFs.getBlockSizeLong();
+      long totalSize = statFs.getBlockCountLong();
+      return blockSize * totalSize;
```
mikehardy: These StatFS.getXxxLong() APIs are for Android API >= 18, but react-native still supports down to API16. The appropriate way to handle this is more like the approach here with an API guard:

https://github.com/rebeccahughes/react-native-device-info/pull/350/files#diff-d2dc05e092aa2dcf49063f6d1f012068R162

(though I think a long is better than a float as a return value, unlike that link...)

machour:
Thank you for your PR. This have been fixed in #587 and published as version 0.26.5.

Please test it and open a new issue if the problem still exists.

mikehardy:
PR #587 was mine so hopefully this fixes the overflow. I will be the first to say that the results of the module's API call aren't always what you expect due to the android storage archiecture, but there is a separate issue tracking that #346 and it should be pursued as a separate well-thought-out PR

jamninetyfive:
> 谢谢你的公关。这已在[＃587中](https://github.com/rebeccahughes/react-native-device-info/pull/587)修复，并以0.26.5版本发布。
> 
> 如果问题仍然存在，请测试并打开新问题。

thanks

machour:
@jamninetyfive could you write that in English? google translate isn't helping

mikehardy:
it translated for me :-), @machour they were just quoting your reply saying it was fixed, and saying "thanks", so - thanks :-)!

machour:
Thank you both then!

PR #537, opened 2018-11-09
compare hasNotch devices with lowercase

tlow92:
## Description

To have a more robust code we should compare brand and model after using lowercase
Please tell me what you think of it

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [ x ] I have tested this on a device/simulator for each compatible OS

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -207,7 +207,7 @@ export default {
     return RNDeviceInfo.isPinOrFingerprintSet;
   },
   hasNotch: function() {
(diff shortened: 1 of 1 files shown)


johankasperi:
I think this is a good idea. For example on my Huawei P20 Pro `RNDeviceInfo.brand` is `HUAWEI` (uppercase) and thereby `DeviceInfo.hasNotch()` returns `false` since there is no lowercase comparison.

PR #539, opened 2018-11-16
Expose timezone and 24 hour properties

christophermark:
## Description

Exposes functions for getting updated values of `timezone` and `is24Hour`, instead of relying on static values.

Timezone and 24-hour time are both preferences that may change on the OS level as the user changes location or modifies their settings. Exposing these getter methods allows clients to pull the freshest value for these properties, as they may change while the app is still up.

## Checklist

* ~~[ ] I added the documentation in `README.md`.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Notes
 
 ### 0.25
+ * Expose getter methods for timezone and 24Hour time (https://github.com/rebeccahughes/react-native-device-info/pull/539)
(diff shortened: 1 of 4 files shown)


mikehardy:
This is tagged for #307 so should go along with that

mikehardy:
thinking further, there is another react-native-community project that handles localization better and this isn't really device specific, we should remove these completely in device-info 2.0 in favor of this module: https://github.com/react-native-community/react-native-localize

this is a good thing though, as it's less code for you+us and it already works!

PR #541, opened 2018-11-20
Fixes iOS battery level

daniel-constantin:
## Description

On iOS the battery level was always -1, even on a real device.

## Checklist


mikehardy:
@machour should close in favor of #404 - it should be enabled once in AppDelegate not toggled off and on, so it's a documentation update really

PR #545, opened 2018-11-27
Fix getUserAgent for android API level <= 16

mtkopone:
## Description

Changed android code for getting the userAgent. 

The system property "http.agent" is present even on older phones (e.g. 4.1.2), so always use that as a fallback.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |     ✅     |
| Windows |   ❌     |

## Checklist

* [X ] I mentioned this change in `CHANGELOG.md`.


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -299,12 +299,14 @@ public String getInstallReferrer() {
...
-      try {
+    try {
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
         constants.put("userAgent", WebSettings.getDefaultUserAgent(this.reactContext));
-      } catch (RuntimeException e) {
+      } else {
         constants.put("userAgent", System.getProperty("http.agent"));
       }
+    } catch (RuntimeException e) {
+      constants.put("userAgent", System.getProperty("http.agent"));
```
mikehardy: Hmm - from what I read (https://stackoverflow.com/a/5261472/9910298) it looks like the http.agent property is the device user agent and will be different than the WebView user agent, so they should not be conflated, but we already did so with the current implementation. Given that, this seems fine as a strategy.

mikehardy:
@machour I looked at this and it seems fine. If you are okay with it I can fix up the CHANGELOG and merge it in

mikehardy:
will be in the next release, thank you!

Issue #215, opened 2017-09-07
Time zone never updates


testshallpasswork:
I found an issue where if you change the device's timezone and then do `DeviceInfo.getTimezone()` it will result in the time zone it was originally when app initially loaded and not the changed timezone.

react-native-device-info: **0.9.8**
react-native: **0.46.3**
Platform: **iOS and Android**

machour:
This is a know limitation of the current version.
Version 1.0.0 will ship with events (#307)
Closing this issue for now.

testshallpasswork:
For anyone that's curious, my workaround was creating a separate bridge module to query the time zone each time vs. only initially as this package does.

machour:
Thank you @testshallpasswork. Is this work publicly available?

testshallpasswork:
It is not but if interest is high I can share some details or post a gist.

machour:
It definitely is for me :)

testshallpasswork:
Ok, here are the highlights: 

```objective-c
// iOS bridge
//  ....
RCT_EXPORT_METHOD(getTimeZone:(RCTResponseSenderBlock)callback)
{
  callback(@[@{@"timeZone": [self timeZone]}]);
}
- (NSString*)timeZone
{
  NSTimeZone *localTimeZone = [NSTimeZone localTimeZone];
  return localTimeZone.name;
}
// ...
```
```java
// Android bridge
// ...
@ReactMethod
public void getTimeZone(final Callback callback) {
    try {
        WritableMap map = Arguments.createMap();
        map.putString("timeZone", TimeZone.getDefault().getID());
        callback.invoke(map);
    } catch (Exception e) {
        WritableMap map = Arguments.createMap();
        map.putString("error", "Sorry, could not get time zone. Please try again.");
        callback.invoke(map);
    }
}
// ...
```
```javascript
// react-native side
// ...
const TimeZoneManager = NativeModules.TimeZoneManager;
// ...
function getTimeZone(callback) {
  TimeZoneManager.getTimeZone(res => callback(res.timeZone));
}
// ...
```

sam1463:
@testshallpasswork @machour any chance we could get this feature into the library before the next major version number, maybe by adding it as a separate method from the current synchronous `getTimeZone` method? It would be really helpful to have this.

machour:
@sam1463 I wouldn't mind adding it if a PR was submitted :)

sam1463:
[...]

Issue #101, opened 2016-10-02
Timezone is not a function


prashen:
Plugin is not working for me: showing error 
timezone is not function. My package.json dependencies are following:
"dependencies": {
    "react": "15.3.2",
    "react-native": "0.34.1",
    "react-native-device-info": "^0.9.5"
  }

yannickoo:
Duplicate of #98 😇

prashen:
Is there any solution :)

yannickoo:
You can adjust your `package.json` by changing following:

`^0.9.5` ➡️  `git+https://github.com/rebeccahughes/react-native-device-info`

prashen:
But I am using "react-native-device-info": "^0.9.5" already.

yannickoo:
Sorry @prashen but this is not enough. Timezone landed in 92c94a115c9a1a67a7cdff1edc852fbca3655600 which is actually after `0.9.5`. We need to wait for a new version or you simply use `master` in the meanwhile by changing the entry in your `package.json` as described above.

<img width="637" alt="screen shot 2016-10-11 at 22 46 30" src="https://cloud.githubusercontent.com/assets/1086961/19288103/c3c8de0e-9004-11e6-9ee1-4489e6123b23.png">
(the `bump version numbers` commit was the change to `0.9.5`)

GantMan:
I'll try to get us a new version released soon.  Just haven't had the time to test on devices :)  It's not easy.

GantMan:
version 0.9.6 published

krishbhattacharyya:
Hi @prashen , 
You should close this issue ... It is already working.

PR #547, opened 2018-11-30
Feature/dynamic timezone query

justinmakaila:
## Description

Fixed issue #215 and #101 

Added `getTimezone` to native modules that allows for asynchronous fetching of the current timezone, rather than relying on the constant to be up to date.

*This is a breaking change; fetching a timezone is now asynchronous and requires setting up promise observation*

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -213,8 +213,8 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
-  public String getTimezone() {
-    return TimeZone.getDefault().getID();
+  public void getTimezone(Promise p) {
+    return p.resolve(TimeZone.getDefault().getID());
   }
 
   @ReactMethod
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -213,6 +213,11 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   }
 
   @ReactMethod
+  public String getTimezone() {
+    return TimeZone.getDefault().getID();
+  }
+
+  @ReactMethod
   public void getBatteryLevel(Promise p) {
     Intent batteryIntent = this.reactContext.getApplicationContext().registerReceiver(null, new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
     int level = batteryIntent.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
@@ -306,7 +311,6 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
         constants.put("userAgent", System.getProperty("http.agent"));
       }
     }
-    constants.put("timezone", TimeZone.getDefault().getID());
     constants.put("isEmulator", this.isEmulator());
     constants.put("isTablet", this.isTablet());
     constants.put("fontScale", this.fontScale());
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -214,7 +214,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
   @ReactMethod
   public void getTimezone(Promise p) {
-    return p.resolve(TimeZone.getDefault().getID());
+    p.resolve(TimeZone.getDefault().getID());
   }
 
   @ReactMethod
--- a/deviceinfo.d.ts
+++ b/deviceinfo.d.ts
@@ -18,7 +18,7 @@ declare const _default: {
   getUserAgent: () => string;
   getDeviceLocale: () => string;
   getDeviceCountry: () => string;
-  getTimezone: () => string;
+  getTimezone: () => Promise<string>;
   getInstanceID: () => string;
   getInstallReferrer: () => string | null;
   isEmulator: () => boolean;
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -189,7 +189,7 @@ export default {
     return RNDeviceInfo.deviceCountry;
   },
   getTimezone: function() {
-    return RNDeviceInfo.timezone;
+    return RNDeviceInfo.getTimezone();
   },
   getFontScale: function() {
     return RNDeviceInfo.fontScale;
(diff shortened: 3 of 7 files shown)


sam1463:
Would love to see this change get merged some time soon. Do you mind updating the Readme/change log? And can any maintainers of this repo take a look at this PR?

benmgreene:
@sam1463 this looks like a great PR, I'm sorry it hasn't been merged yet. Did you happen to release a forked package on npm with this change?

justinmakaila:
@benmgreene no, but you should be able to use it like this:
```js
"dependencies": {
    "react-native-device-info": "git+https://github.com/wellth-app/react-native-device-info.git#feature/dynamic-timezone-query",
}
```

spodatelev:
Would greatly appreciate this to be merged.

mikehardy:
Hi there! I'm a new maintainer and could merge this. Looks like a nice clean change but it is breaking as it alters an existing api.

We have a couple other things we would like to batch up if we do a breaking change so I see a couoke options. This could be altered to have a new api so it's a minor version and we clean it up later, or we wait a bit more until the 2.x discussion work is implemented then this goes as a part of it. I will probably circle back to do 2.x things in a week or two after my day job calms down

I'll leave that choice up to the impmenter. It would need a change log and Readme entry either way

mikehardy:
Maybe as a temporary workaround that would lots of other use cases, a new oromise api that exposed a refresh of all constants? Locale and languages can change as can many others so that would maybe get them all at once until we had a promise api and/or listeners that emitted events when the dynamics items changes

spodatelev:
@mikehardy the idea w/refreshing constants is a good workaround. For this particular issue I'd prefer something like `getTimezoneRealtime()`.

mikehardy:
Yeah I'm not sure the constants would be refreeshed a cross the bridge anyway, I'd have to read and test that. I think the api you prooose is a stopgap but if it is backwards compatible I can accept it now temporarily on the way to a promise api. Still need Readme and changeling though 😁

mikehardy:
With apologies for both the delay and this resolution, I'm not going to merge this. But! It is because this feature and a pile of others related to localization apparently already have a great home https://github.com/react-native-community/react-native-localize#gettimezone

this is implemented and should be working and ready for you to integrate this instant.

I don't want to duplicate functionality across the umbrella of react-native-community so that package should be the go-to for localization stuff

spodatelev:
@mikehardy thanks for providing clarification on the status.
Just in case: did you check whether `react-native-localize` option was realtime?
The `getTimeZone` function is not async there.

mikehardy:
It's better than that. It is "static" in that they fetch it and store it and return that to you. But! (another good 'But!'), they listen to the device broadcasts for localization changes like I wanted to do here so that it is an efficient implementation, and then they emit an event so you can listen and update in realtime but without having to poll. Very cool

https://github.com/react-native-community/react-native-localize/blob/master/android/src/main/java/com/reactcommunity/rnlocalize/RNLocalizeModule.java#L113

Even better, and why I will not duplicate functionality here, is that they have already thought of and handled only doing the local update while the app is running, just in case.

mikehardy:
in iOS as well - event-based locale change notifications

spodatelev:
Thanks for such a thorough response.

mikehardy:
No problem. I was literally in this area last night on my "day job" app so I had all the tabs open. I also have the tab open with the code you probably want, that dynamically listens to locale changes and updates your view dynamically. It's typescript but if you can create an I18NService singleton, and feed it a ref to your base component you have pure joy.

I18NService in my app. It is initialized in a bootstrapper by the RootView is rendered.

```typescript
import * as RNLocalize from 'react-native-localize';
import * as I18n from 'i18n-js';
//import { memoize } from 'lodash-es';
import { translations } from '../translations';
import { Component } from 'react';

class I18NService {
  private static _instance: I18NService = new I18NService();

  private constructor() {
    // This is a singleton no one else gets to instantiate or the hooks
    // for forcing update when locale updates won't work as the reference won't be bound
    if (I18NService._instance) {
      throw new Error('I18NService is a singleton. Call .getInstance() instead.');
    }
    I18NService._instance = this;
  }

  public static getInstance(): I18NService {
    return I18NService._instance;
  }

  translate: any;

  startup(): SyncTasks.Thenable<void> {
    console.log('I18NService startup called');

    this.setI18NConfig();
    // TODO performance optimization for the future - hard to test with it though
    // this.translate = memoize(
    //   (key: string, config: any) => I18n.t(key, config),
    //   (key, config) => (config ? key + JSON.stringify(config) : key)
    // );
    this.translate = (key: string, config: any) => I18n.t(key, config);

    RNLocalize.addEventListener('change', this.handleLocalizationChange.bind(this));

    return SyncTasks.Resolved<void>();
  }

  setI18NConfig = () => {
    const fallback = { languageTag: 'en', isRTL: false };
    const { languageTag, isRTL } =
      RNLocalize.findBestAvailableLanguage(Object.keys(translations)) || fallback;
    console.log('received best supported language of tag ' + languageTag);

    // clear translation cache
    //this.translate.cache.clear();

    // update layout direction
    I18nManager.forceRTL(isRTL);

    // set i18n-js config
    I18n.translations = {
      [languageTag]: translations[languageTag],
      [fallback.languageTag]: translations[fallback.languageTag],
    };
    I18n.fallbacks = true;
    I18n.defaultLocale = 'en';
    I18n.locale = languageTag;

    //I18n.missingBehavior = 'guess';
    //I18n.missingTranslationPrefix = 'TRANSLATE: ';
    I18n.missingTranslation = (param: any) => {
      return 'I18N: ' + param;
    };
  };

  handleLocalizationChange() {
    console.log('i18NService.handleLocalizationChange called');
    this.setI18NConfig();
    console.log('current best language now: ' + I18n.locale);
    if (this.reactComponent) {
      console.log('we have a component to update! Trying it');
    }
    this.reactComponent && this.reactComponent.forceUpdate();
  }

  reactComponent: Component | null = null;

  setComponentForUpdate(ref: Component | null) {
    this.reactComponent = ref;
  }
}

export default I18NService.getInstance();
```

RootView.tsx in my app
```tsx
mport I18N from '../services/I18NService';
YellowBox.ignoreWarnings(['Warning: Async Storage has been extracted']);

let authScreenMap: Router.NavigationRouteConfigMap = {

  render(): JSX.Element | null {
    return (
      <RX.View style={_styles.root} onLayout={this.props.onLayout}>
      <RX.View
        style={_styles.root}
        onLayout={this.props.onLayout}
        // Allow the I18NService to update the view hierarchy if it detects a locale change
        ref={viewRef => {
          I18N.setComponentForUpdate(viewRef);
        }}
      >
```


Maybe this will help others

spodatelev:
Thanks for sharing, Mike.

PR #548, opened 2018-11-30
Feature/android sdk 28

justinmakaila:
## Description

Updates to android sdk 28

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: 'com.android.library'
 
-def DEFAULT_COMPILE_SDK_VERSION             = 24
-def DEFAULT_BUILD_TOOLS_VERSION             = "25.0.2"
-def DEFAULT_TARGET_SDK_VERSION              = 22
+def DEFAULT_COMPILE_SDK_VERSION             = 28
+def DEFAULT_BUILD_TOOLS_VERSION             = "28.0.0"
+def DEFAULT_TARGET_SDK_VERSION              = 28
 def DEFAULT_GOOGLE_PLAY_SERVICES_VERSION    = "+"
 
 android {


machour:
Apologies for not getting back to you sooner.
Is this safe to ship in 0.27.0 ? Will this break existing apps using this lib ?

puranjayjain:
I think this change will impact the people targeting lower SDK versions as this changes the default target SDK version.

BenWildeman:
you can make sure all your submodules target the same SDK you're using by doing the following in your root build.gradle file:

```
subprojects {
    afterEvaluate {project ->
        if (project.hasProperty("android")) {
            android {
                compileSdkVersion 28
                buildToolsVersion "28.0.3"
            }
        }
    }
}
```

mikehardy:
This change should have small impact (or, a diminishing, user-controllable impact) as the react-native style for android builds has changed such that everyone should be controlling their gradle versions in their own app build.gradle

you can see the versions are checked for external definition here so this change will only effect people that haven't changed to the new style yet https://github.com/rebeccahughes/react-native-device-info/blob/master/android/build.gradle

mikehardy:
@machour I think this should merge, people should have specific versions set in their app/build.gradle if they want to control them.

@justinmakaila if you could also PR a sample app that's just a fresh react-native init with this linked in that would be fantastic. I mention that because for targetSdk API28 people have to be careful of plaintext loading in Android (like this https://github.com/Microsoft/reactxp/pull/1048/commits/5558bddebd1972d1ac044bde9e507f0f01c6d928#diff-6ce28bd386e24468a9ee538ca35dabf8) or their dev build will break, and it will decrease support load if there is an example and we can point to it

machour:
react-native-device-info 0.28.0 published 🎉

(did this one last so that we have 0.28.0 for sdk 28 🤓)

PR #560, opened 2019-01-18
feat(android/ios): Adding getDeviceType function and changing behavior of isTablet

fxdemolisher:
## Description

- getDeviceType will return: Handset, Tablet, Tv, or Unknown
- Supported on iOS and Android.  Always set to Unknown on web. Unsupported on windows
- On iOS/Android, modified isTablet logic to just check if deviceType == Tablet

On iOS, device type detection is done via user interface idiom detection for the device and is thus exact.

On Android, TV platform detection checks for both Android TVs (via UI manager mode) and FireTVs (via system feature check). For non TV devices we attempt to calculate the physical screen size (diagonal) and set some sane cutoffs for when a device is a Handset (3" -> 6.9") or a Tablet (7" -> 18"). If detection fails the deviceType is set to Unknown.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


fxdemolisher:
@machour we would like to use this change in our main app, it would be great to get some feedback on the change. Thanks.

machour:
Thank you a lot for this nice PR!
Just shipped it in react-native-device-info 0.26.0 🎉

PR #563, opened 2019-01-22
added Oneplus 6T

MarianBe:
added OnePlus 6T to the notch devices


machour:
Thank you, shipped in 0.26.1 🎉

Issue #127, opened 2016-12-07
Where is getImei ?


mkjangid:
I dont see a method to retrieve device IMEI.

oximer:
I'm also interested on retrieving the Android IMEI

jinxac:
+1

mkjangid:
Do u guys want a solution? I have some code which does the job. Can post it on stack overflow.

jinxac:
Yup.. you can create a git gist also. Can you please share it?

mkjangid:
Here is it . Find comments at bottom. Wrote a native module for this.

https://gist.github.com/mkjangid/d717f4f7c236c8085c1597e163951612

jinxac:
@mkjangid do we have something similar for IOS too?

mkjangid:
I am not an IOS guy but I think IOS doesn't allow access to device IMEI

rsp8055:
I didn't get it, in link https://gist.github.com/mkjangid/d717f4f7c236c8085c1597e163951612  specifies files 
please tell me how this file will import in MainApplication.java or how can i use it in my react native code. even in RNDeviceModule.java file how to put getImei() code?

Anybody please help me.
Thank you in advanced.

gabsima:
See pull request #250

koswarabilly:
use [react-native-imei](https://github.com/SimenCodes/react-native-imei) instead it works anyway but you might have to edit a line of the code inside node-module, when you first compile if you encounter error, dont panic , see the error and remember which line, delete the line that cause you error and it is done

luisfuertes:
@koswarabilly what line must delete? App crash with debugger active

mikehardy:
For people that need this now https://www.npmjs.com/package/react-native-imei

For Android: you have to be really careful with IMEI.
For iOS: it is simply not available.

It's a unique device identifier that is hard/impossible to change so it is touchy piece of personally identifiable information. Be careful...

mikehardy:
#565 will fix this when it's merged, but it needs some work and hasn't been patched up in response to review

PR #565, opened 2019-01-27
Rndeviceinfo/imeiNumber

bansalayush:
## Description
Fixed issue #127

Added imeiNumber to constants.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


bansalayush:
Hey @machour , please check this .

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -317,6 +317,7 @@ public String getInstallReferrer() {
             getCurrentActivity().checkCallingOrSelfPermission("android.permission.READ_PHONE_NUMBERS") == PackageManager.PERMISSION_GRANTED)) {
       TelephonyManager telMgr = (TelephonyManager) this.reactContext.getApplicationContext().getSystemService(Context.TELEPHONY_SERVICE);
       constants.put("phoneNumber", telMgr.getLine1Number());
+      constants.put("imeiNumber", telMgr.getDeviceId());
```
machour: From the docs:

> Returns the unique device ID of a subscription, for example, the IMEI for GSM and the MEID for CDMA phones. 
> Return null if device ID is not available.
>
> This method was deprecated in API level 26.
> 
> Use (@link getImei} which returns IMEI for GSM
>
> or (@link getMeid} which returns MEID for CDMA.

bansalayush: Sure

mikehardy: I looked into this - it needs more changes than this. There may be 0 or more (up to 3 currently) "phones" on Android, so you need to first get the count: https://developer.android.com/reference/android/telephony/TelephonyManager.html#getPhoneCount()

But that's only available for API23+, so you have to check if it's API23+ and if not to get that info.

Then from 0 to phone count, you need to getPhoneType, then either getMeid or getImei for the slot (for API26+) or getDeviceId for API25- 

I would suggest an API name like getCarrierSubscriberIDs() that returns a string[], containing 0 or more entries, each possibly null, populated like the above.

As a side benefit, people can tell how many cell slots a phone has by doing .length on the return value

I recognize that's a pretty significant change request but I think it's the right thing to do - the results don't necessarily make sense on a lot of devices (including mine, which has one SIM in slot 2) otherwise. I'll approve this if it works like that

mikehardy:
The whole cellular part of this package could use an overhaul - when it started people had one number mostly and not a lot else. The TelephonyManager API on Android is huge now and there's no reason once we've done the permissions / API version checks we are already  doing that we can't fetch it all.

I'm going to tag this for #307 and the main thing to think about is how to handle the data that comes from the system APIs in such widely varying formats across the APIs? If an earlier API doesn't support returning a certain type of data should it be missing from the JSON, null, or an empty string?

mikehardy:
I notice the [react-native-carrier-info package](https://github.com/react-native-webrtc/react-native-carrier-info) and I actually use it myself - Maybe we can work together and pull that package in here to have a single place and one set of infrastructure? @zxcpoiu - what do you think? how do you feel about having the carrier-info stuff here inside the community umbrella?

zxcpoiu:
Sure, given that there may be lots of duplicate works, I was thinking the same thing, too.

@ianlin and I just transferred [react-native-carrier-info](https://github.com/react-native-webrtc/react-native-carrier-info) under newly created orgs [react-native-webrtc](https://github.com/react-native-webrtc) ( with @saghul )

Do you mean transfer [react-native-carrier-info](https://github.com/react-native-webrtc/react-native-carrier-info) to [react-native-community](https://github.com/react-native-community/)?

Or merged into react-native-device-info?

mikehardy:
Hi @zxcpoiu ! :wave: :smile: 

More than anything, I just prefer to see things implemented once, and be well-supported. So that's my goal. 

1- To your questions - react-native-community or not - react-native-community isn't the perfect home for everything so I'm not that bothered if it is in there or not.

2- merged into react-native-device-info - this is a possibility and what I was thinking of doing because the APIs implemented weren't very complicated at the moment, while react-native-community provides us a lot of free tooling (linting, communication, circleci stuff, best practice discussions, etc). And I hadn't seen a lot of activity on carrier-info :man_shrugging: - but if you guys are active, and will accept PRs and release, that works for me - it would be my preference

The key for me is merging and releasing - if you guys are okay with maintaining carrier-info (and ideally expanding it or merging PRs to expand it to cover the TelephonyManager Android API explosion) that would be perfect!

zxcpoiu:
Honstly, I don't have much time to activily expand it, but would help to reviewing PRs and release when free.

It would be great to put carrier info and device info under the same organization since they are relatively similar.

Seems you are already a collaborator of device info, I believe you are the choosen one to be the lead of carrier info as well 😄

If you are ok with it, I would happy to transfer carrier info to the community. To merge into device info or not can be discussed later, obviously there are pros and cons for both options, some may not want to add unused codes...etc.

@saghul is there any one willing to take the mission you can think of? or I prefer to transfer next to device info where @mikehardy shines 😊

mikehardy:
Not a lot of activity on the "is there a better home for the APIs" and I believe that even though my personal time is limited, since react-native-device-info is in react-native-community it makes a very strong home for things...when they make sense. I am leaning towards the idea that it makes sense to keep these APIs here then. This thinking, and the process is intended to be friendly + collaborative so I'm just sharing the thought now with the implementation to follow in a while. Anyone with different ideas, as always feel free to chime in. :+1:

mikehardy:
Given my previous comment then (about keeping the API here and making it as great as possible) my review comments on the PR are still valid then. We need to handle multiple-SIMs well and any motion in this PR to do so would be greatly appreciated

zxcpoiu:
I think it would be fine to transfer to community, seems there is no one who has clear road map to the api and can actively implement them. To me, you really are the chosen one @mikehardy , you have enough experience to the similarity between both `device` and `carrier` repo and might have better considerations when developing new api, especially when it is ambiguous where this api should belongs to or duplication.

mikehardy:
I resolved the conflicts vs the PR that landed a while back where phone permissions were handled in API-version-correct ways.

This is a simple enough change as is though I wish it handled the multi-SIM cases correctly. The previous implementation did not either though so I am provisionally willing to accept this PR.

However, I'd like to see the chores done as part of the PR before I merge - README update for API documentation, the default empty polyfill needs an update, the example App.js needs an example usage, and a CHANGELOG entry. It's not glamorous but it's high quality :sparkles:  :-). Update those @bansalayush and I'll merge this in. If you don't respond in a reasonable time I'll close it. Thanks

mikehardy:
This will have to wait for a v4 breaking change type release where we ingest carrier-info features - I'm doing v3 right now which will be big enough and has some pressure on it from Apple API deprecations

mikehardy:
No response from submitter

PR #571, opened 2019-02-08
Fixed getDeviceType method.

vladrozhnev:


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -145,7 +145,7 @@ private static DeviceType getDeviceType(ReactApplicationContext reactContext) {
 
     // Calculate physical size.
     double widthInches = metrics.widthPixels / (double) metrics.xdpi;
-    double heightInches = metrics.widthPixels / (double) metrics.xdpi;
+    double heightInches = metrics.heightPixels / (double) metrics.xdpi;
```
fxdemolisher: oops, bad copy/paste on my part in the original. I wonder if `metrics.xdpi` in the end of this should be `metrics.ydpi` instead though?

vladrozhnev: @fxdemolisher seems you a right.

machour:
Thank you so much for spotting & fixing this @vladrozhnev !

As always, your reviews are appreciated @benoitdion & @fxdemolisher

PR #573, opened 2019-02-18
Update deviceinfo.js, Fix hasNotch() support for LG phones

beaulieufrancois:
Fix hasNotch() support for LG phones 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅  |

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -70,6 +70,18 @@ const devicesWithNotch = [
     brand: 'Leagoo',
     model: 'S9',
   },
+    {
+    brand: 'LG',
+    model: 'G7',
+  },
+  {
+    brand: 'LG',
+    model: 'G7 ThinQ',
+  },
+  {
+    brand: 'LG',
+    model: 'G7+ ThinQ',
+  },
   {
     brand: 'LG',
     model: 'LM-Q910', //G7 One
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -72,15 +72,15 @@ const devicesWithNotch = [
   },
   {
     brand: 'LG',
-    model: 'G7',
+    model: 'LM-Q910', //G7 One
   },
   {
     brand: 'LG',
-    model: 'G7 ThinQ',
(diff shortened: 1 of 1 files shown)


deviceinfo.js
```diff
@@ -72,15 +72,15 @@ const devicesWithNotch = [
...
   },
   {
     brand: 'LG',
-    model: 'G7 ThinQ',
+    model: 'LM-G710', //G7 ThinQ
   },
   {
     brand: 'LG',
-    model: 'G7+ ThinQ',
+    model: 'LM-V405', //V40 ThinQ
```
machour: Maybe the old values work for other build of those phones.
Can you keep the old one & add new entries instead?

machour:
Merci monsieur!

beaulieufrancois:
Au plaisir!

machour:
Published in 0.26.4 🎉

GiladR1979:
This is still happening on 5.4.0. hasNotch() returns false.
DeviceInfo.getBrand() returns "lge" and not "lg".

GiladR1979:
This is still happening on 5.4.0. 
DeviceInfo.getBrand() returns "lge" and not "lg". => getting "false" in hasNotch() on LG G7 ThinQ

mikehardy:
@GiladR1979 you've mentioned it in 3 places now, which takes the same time as a pull request :-), please make a PR and I'll merge it!

GiladR1979:
Oops, I thought I deleted one response and responded in a wrong place a second time.
Anyway is copying the LG brand to LGE a solution? Its not a pretty one. I don't know why the brand is "LGE" and not "LG" in the first place.

mikehardy:
The entire hasNotch() implementation is a little silly if you think about it. Copy/paste is no better or worse than anything given how it works now. 

What you really want (to know how to display) is handled better by https://github.com/th3rdwave/react-native-safe-area-context

Everyone should really move there but I provide hasNotch here still because it's always been here and doesn't really hurt to maintain as long as the community keeps it working well

PR #583, opened 2019-03-04
[PS] Added isAutoDateAndTime and isAutoTimeZone function

ps1223:
## Description

Added `isAutoDateAndTime()` that allows user to verify whether the automatic time and date option is eneabled.

Added `isAutoTimeZone()` that allows user to verify whether the automatic time zone option is enabled.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

### DIFF
--- a/README.md
+++ b/README.md
@@ -936,7 +936,9 @@ Tells if the Get Automatic date & time setting is enabled in phone.
 **Examples**
 
 ```js
-const isAutoDateAndTime = DeviceInfo.isAutoDateAndTime(); // true
+DeviceInfo.isAutoDateAndTime().then(isAutoDateAndTime => {
+  // true or false
+});
 ```
 
 ### isAutoTimeZone()
@@ -946,7 +948,9 @@ Tells if the Get Automatic time zone setting is enabled in phone.
 **Examples**
 
 ```js
-const isAutoTimeZone = DeviceInfo.isAutoTimeZone(); // true
+DeviceInfo.isAutoTimeZone().then(isAutoTimeZone => {
+ // true or false
+});
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -251,8 +251,8 @@ import DeviceInfo from 'react-native-device-info';
 | [hasNotch()](#hasNotch)                           | `boolean`           |  ✅  |   ✅    |   ✅    | 0.23.0 |
 | [isLandscape()](#isLandscape)                     | `boolean`           |  ✅  |   ✅    |   ✅    | 0.24.0 |
 | [getDeviceType()](#getDeviceType)                 | `string`            |  ✅  |   ✅    |   ❌    | ?      |
-| [isAutoDateAndTime()](#isAutoDateAndTime)         | `string`            |  ❌  |   ✅    |   ❌    | ?      |
-| [isAutoTimeZone()](#isAutoTimeZone)               | `string`            |  ❌  |   ✅    |   ❌    | ?      |
+| [isAutoDateAndTime()](#isAutoDateAndTime)         | `boolean`            |  ❌  |   ✅    |   ❌    | ?      |
+| [isAutoTimeZone()](#isAutoTimeZone)               | `boolean`            |  ❌  |   ✅    |   ❌    | ?      |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -250,6 +250,8 @@ import DeviceInfo from 'react-native-device-info';
 | [hasNotch()](#hasNotch)                           | `boolean`           |  ✅  |   ✅    |   ✅    | 0.23.0 |
 | [isLandscape()](#isLandscape)                     | `boolean`           |  ✅  |   ✅    |   ✅    | 0.24.0 |
 | [getDeviceType()](#getDeviceType)                 | `string`            |  ✅  |   ✅    |   ❌    | ?      |
+| [isAutoDateAndTime()](#isAutoDateAndTime)         | `string`            |  ❌  |   ✅    |   ❌    | ?      |
+| [isAutoTimeZone()](#isAutoTimeZone)               | `string`            |  ❌  |   ✅    |   ❌    | ?      |
 
 ---
 
@@ -901,6 +903,26 @@ Returns the device's type as a string, which will be one of:
 
 ```js
 const deviceType = DeviceInfo.getDeviceType(); // 'Handset'
+```
+
+### isAutoDateAndTime()
+
+Tells if the Get Automatic date & time setting is enabled in phone.
+
+**Examples**
+
+```js
+const isAutoDateAndTime = DeviceInfo.isAutoDateAndTime(); // true
+```
+
+### isAutoTimeZone()
+
+Tells if the Get Automatic time zone setting is enabled in phone.
+
+**Examples**
+
+```js
+const isAutoTimeZone = DeviceInfo.isAutoTimeZone(); // true
 ```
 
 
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -265,6 +265,28 @@ public void isAirPlaneMode(Promise p) {
...
 
+  @ReactMethod
+  public void isAutoDateAndTime(Promise p) {
+    boolean isAutoDateAndTime;
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1) {
+      isAutoDateAndTime = Settings.System.getInt(this.reactContext.getContentResolver(),Settings.System.AUTO_TIME, 0) != 0;
+    } else {
+      isAutoDateAndTime = Settings.System.getInt(this.reactContext.getContentResolver(),Settings.Global.AUTO_TIME, 0) != 0;
+    }
+    p.resolve(isAutoDateAndTime);
```
machour: Could you use the if to just pickup `Settings.System.AUTO_TIME` or `Settings.Global.AUTO_TIME` to make the code more readable?

```
String name;

if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1) {
    name = Settings.System.AUTO_TIME;
} else {
    name = Settings.Global.AUTO_TIME;
}

p.resolve(Settings.System.getInt(this.reactContext.getContentResolver(), name, 0) != 0);
```

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -265,6 +265,28 @@ public void isAirPlaneMode(Promise p) {
...
+  public void isAutoTimeZone(Promise p) {
+    boolean isAutoTimeZone;
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1) {
+      isAutoTimeZone = Settings.Global.getInt(this.reactContext.getContentResolver(),Settings.System.AUTO_TIME_ZONE, 0) != 0;
+    } else {
+      isAutoTimeZone = Settings.Global.getInt(this.reactContext.getContentResolver(),Settings.Global.AUTO_TIME_ZONE, 0) != 0;
+    }
+    p.resolve(isAutoTimeZone);
+  }
+
```
machour: Here too please

ps1223:
Hi,

Updated the changes and README file.

machour:
@mikehardy 👀

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -275,6 +275,28 @@ public void isAirPlaneMode(Promise p) {
...
+  public void isAutoTimeZone(Promise p) {
+    String name;
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1) {
+      name = Settings.System.AUTO_TIME_ZONE;
+    } else {
+      name = Settings.Global.AUTO_TIME_ZONE;
+    }
+    p.resolve(Settings.Global.getInt(this.reactContext.getContentResolver(), name, 0) != 0);
+  }
+
```
mikehardy: This is really really close to being perfect, however, if you get the name from one of Settings.System or Settings.Global, you also need to resolve it from that source. There is an example of what I mean in the airplane mode check just above this one - the whole content resolution is switched on the Build.VERSION_CODE, not just the name resolution: https://github.com/rebeccahughes/react-native-device-info/blob/0de338f2aded9bea71195b9f4565aa1f9505a00f/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L268

README.md
```diff
@@ -251,6 +251,8 @@ import DeviceInfo from 'react-native-device-info';
 | [hasNotch()](#hasNotch)                           | `boolean`           |  ✅  |   ✅    |   ✅    | 0.23.0 |
 | [isLandscape()](#isLandscape)                     | `boolean`           |  ✅  |   ✅    |   ✅    | 0.24.0 |
 | [getDeviceType()](#getDeviceType)                 | `string`            |  ✅  |   ✅    |   ❌    | ?      |
+| [isAutoDateAndTime()](#isAutoDateAndTime)         | `string`            |  ❌  |   ✅    |   ❌    | ?      |
+| [isAutoTimeZone()](#isAutoTimeZone)               | `string`            |  ❌  |   ✅    |   ❌    | ?      |
```
mikehardy: oops - 2 changes actually, this is a boolean return isn't it? same as isAirplaneMode

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
ps1223: @mikehardy Yea, I just didn't notice that part of it. Made this changes after suggestion from @machour. Now I feel I could just revert the commit.

mikehardy: If I recall I think the suggestion from @machour was to protect the API on VERSION.SDK_INT < Build.VERSION_CODES.JELLEY_BEAN_MR1. That part is still vital, as react-native works on those versions, but the API changed during them.

The part I mean is that the *entire call* (including the content resolution part, not just the location of the name) needs protection by API version

Which is to say I don't think a revert is called for, a more thorough API protection / usage is called for

ps1223: @mikehardy Yea, I made a mistake in my initial commit due to which @machour tried to refactor it and I again failed to notice the change. Any thanks for marking it, resolved it now.

README.md
```diff
@@ -927,6 +929,27 @@ Returns the device's type as a string, which will be one of:
...
+```
+
+### isAutoTimeZone()
+
+Tells if the Get Automatic time zone setting is enabled in phone.
+
+**Examples**
+
+```js
+const isAutoTimeZone = DeviceInfo.isAutoTimeZone(); // true
```
mikehardy: I'm really sorry I didn't see this before - the usage example should be for promise - it should end with .then(etc) - there are other examples for other Promise APIs.

I swear this is the last thing :-), this looks good to merge if the example is updated - thanks for the continued effort + patience

ps1223: Not a problem, actually my mistake haven't seen those little things properly. Any how its all sorted now. Thanks for pointing out the mistakes with patience.

mikehardy:
@machour this looks clean and ready for launch - we don't have CI and I haven't pulled it but I have re-read it many times at this point

machour:
react-native-device-info 0.29.0 published 🎉

mikehardy:
@ps1223 hey there! I'm following up on these APIs as I prepare to move them to react-native-localize (a different react-native-community package that seems like a more natural home for them) and I have searched and searched but was unable to find similar APIs for iOS. So maybe you know: when you were implementing this did you attempt to look for iOS APIs to implement it for that platform and could not find them either? Or did you focus on Android only? Thanks!

ps1223:
@mikehardy Hi! Actually I've implemented this only for android, I will check whether any option is available for iOS.

ps1223:
Hi @mikehardy, It seems there is no option available now for this API in iOS

mikehardy:
@ps1223 I came to the same conclusion but am very happy someone else looked to make sure I hadn't missed anything. Thanks for checking!

Issue #581, opened 2019-03-01
hasNotch not working on Pocophone F1


NemanjaManot:
## Summary

**DeviceInfo.hasNotch()** returns false on **PocoPhone F1**

Version: 0.26.4  
Affected OS: Android   
OS Version: 9

edreyyo:
The Poco F1 is in the list of `devicesWithNotch` (see https://github.com/rebeccahughes/react-native-device-info/blob/9d92e55dfe9bc6271887a4637ef0d5b95eb44205/deviceinfo.js#L177) and so should return `true`. What does it show if you call `console.log(DeviceInfo.getModel())` from your Poco F1?

NemanjaManot:
@edreyyo 
'POCOPHONE F1'

edreyyo:
Thanks, I have just submitted a fix @NemanjaManot. It will need to be approved by the repository owner.

machour:
Fixed

PR #584, opened 2019-03-07
hasNotch should return true for Pocophone F1

edreyyo:
Fixed issue #581 

`DeviceInfo.getModel()` returns `'POCOPHONE F1'`, however currently this is getting matched against `'POCO F1'`, which evaluates to `false`. This fixes it.

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -176,6 +176,10 @@ const devicesWithNotch = [
   },
   {
     brand: 'xiaomi',
+    model: 'POCO F1',
+  },
+  {
+    brand: 'xiaomi',
     model: 'POCOPHONE F1',
   },
   {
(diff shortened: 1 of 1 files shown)


mikehardy:
@machour this looks clean and simple, should merge

deviceinfo.js
```diff
@@ -176,7 +176,7 @@ const devicesWithNotch = [
   },
   {
     brand: 'xiaomi',
-    model: 'POCO F1',
+    model: 'POCOPHONE F1',
```
machour: Hi, could you add another line instead of changing this one?

Some phones will report different model names depending on the country/operator/etc.

edreyyo: Hi @machour, thanks for the suggestion. Pull request updated.

machour:
Thank you @edreyyo, react-native-device-info 0.28.1 published 🎉

PR #587, opened 2019-03-10
Use BigInteger to avoid overflow (Fixes #331)

### LINKED ISSUE #331: getFreeDiskStorage returns negative number

mikehardy:
## Description

Fixed issue #331 

There are other PRs floating around for this one but they are all messy or abandoned or both.

This is a very focused change that changes the use of the existing native API calls (which are deprecated but also work back to the react-native supported API16) so that they are computed with protection against integer overflow per this comment from @knot123: https://github.com/rebeccahughes/react-native-device-info/issues/331#issuecomment-428848892

There are valid reasons why you might want to use different native API calls to total up space in Android but you could also argue those are all behavior changes for this module API, so this just fixes the overflow and leaves it at that

## Checklist

* [X] I have tested this on a device/simulator for android (it's an android fix)

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -39,6 +39,7 @@ import java.util.Map;
 import java.util.TimeZone;
 import java.lang.Runtime;
 import java.net.NetworkInterface;
+import java.math.BigInteger;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


mikehardy:
@machour this should be good to go - I'd like for the numbers to be more meaningful but that's different native calls and first we should just stop overflowing...

machour:
Published in 0.26.5, thank you!

PR #589, opened 2019-03-11
Example demonstrating usage

mikehardy:
## Description

Here's an example demonstrating how to use the module for Android and iOS

I implemented it with the included refresh-example.sh bash script on macOS so you can refresh it any time (as react-native upgrades, the scaffolding changes all the time)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


mikehardy:
Everything in the PR is under the "example" sub-directory except the script I used to make it: https://github.com/rebeccahughes/react-native-device-info/pull/589/files#diff-6cfa08d7068deb50cdd88d17ad5c34da

mikehardy:
I don't have a problem scavenging, however my app literally uses every single function so I don't think there is more to test? I even do the pin/password lock callback and the new isBatteryCharging + iOS AppDelegate thing - there is not a single thing I think I can cover. So I think that is not needed

As for npm-ignore and bash script - sure! Had no idea that was even a feature of NPM, I'll read up to learn, and do it.

mikehardy:
it continues below, with the rest of everything (and looks the same on iOS, everything working):

Screen Region 2019-03-11 at 08 10 26

mikehardy:
Actually, I looked at your example and I must admit it is clever - as a programmer I like the API automation. However, I think that's actually an anti-pattern for an example even though it is the most efficient way to implement an example. The reason I say that is that the goal is to demonstrate normal usage, whereas your App.js demonstrates abstract usage. Other than that, the name change will require fiddling any time we need to refresh the example with new react-native version scaffolding and in my experience on other modules it is the example going out of date with regard to scaffolding that is the real pain point vs automating the example itself. So I still like the fully-automated refresh style, and I think my version of the App.js is most useful to example readers, but I like progress more than small differences, so that is totally your call.

I did add the .npmignore though, that's an excellent point...

machour:
I totally agree with your point on the code style for the example, nice work mate!

mikehardy:
Great - and happy to help, this module is important to my app, thank you for keeping it in shape. I'll lay low now for a little and just listen a while as these recent changes settle, but hopefully the example helps users and reduces the support load.

PR #590, opened 2019-03-13
Ignore missing permissions in lint, specify minor support library versions

### LINKED ISSUE #325: Lint Error Android Build
## Summary

when running :react-native-device-info:lint I get an error. Originally it was complaining about RNDeviceModule.java overriding or implementing a deprecated API, but I can't reproduce that error. I have attached the lint results print out.

|             |     |
| ----------- | --- |
| Version     | 0.15.2  |
| Affected OS | ?   |
| OS Version  | ?   |

mikehardy:
Took the advice from a comment on the linked issue: https://github.com/react-native-community/react-native-device-info/issues/325#issuecomment-448498293

## Description

Should fix #325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Release Notes
 
+### 0.29.1
+ * Fix lint errors (library dependences, ignore permissions) (https://github.com/react-native-community/react-native-device-info/pull/590)
+
 ### 0.29.0
  * Add `isAutoDateAndTime()` and `isAutoTimeZone()` (https://github.com/rebeccahughes/react-native-device-info/pull/583)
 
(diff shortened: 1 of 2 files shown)


machour:
Could you add the Changelog entry? We'll be incrementing last digit

mikehardy:
Changelog added, and in further testing I discovered lint gets really mad about support library mismatches, so I had to specify fallback versions for some minor support libraries that matched the main support library version.

That should be backwards compatible as the versions there move in lockstep with the main support library, and we were already relying on either the app to specify it or a version we specified with the previous 28.0.0 version

mikehardy:
Oh, should mention that the test would be to merge the PR (or pull my branch), do a `npm install` in the example directory, then `./gradlew :react-native-device-info:lint`. If you do that in master (before merge) it'll fail, but from my PR branch it passes now.

machour:
react-native-device-info 0.29.1 published 🎉

PR #591, opened 2019-03-13
Getting podspec details from package information

maestor:
For example version is now hardcoded antique 0.21.5. So why not get needed details from package.json and those are always up-to-date.

## Description

Fixed podspec to be more dynamically.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

* [X] I have tested this on a device/simulator for each compatible OS (tested that app with device-info dependency runs and works normally with iOS simulator after change)

### DIFF
--- a/RNDeviceInfo.podspec
+++ b/RNDeviceInfo.podspec
@@ -1,12 +1,16 @@
+require 'json'
+
+package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+
 Pod::Spec.new do |s|
   s.name         = "RNDeviceInfo"
-  s.version      = "0.21.5"
-  s.summary      = "Device Information for react-native"
+  s.version      = package['version']
+  s.summary      = package['description']
 
-  s.homepage     = "https://github.com/rebeccahughes/react-native-device-info"
+  s.homepage     = package['repository']['url']
 
-  s.license      = "MIT"
-  s.authors      = { "Rebecca Hughes" => "rebecca@learnium.net" }
+  s.license      = package['license']
+  s.authors      = package['author']
   s.ios.deployment_target = '8.0'
   s.tvos.deployment_target = '10.0'
 


mikehardy:
I will quickly admit I am very new to iOS development. This looks very clean and simple though. May I ask what steps you took to test that it worked, and what results I should expect before and after your change so I can reproduce that it works as intended?

maestor:
I have project where that library is one dependency. Since pods are default way to go with react native sooner than later (https://github.com/react-native-community/discussions-and-proposals/issues/104, first issue), I have move everything I easily could installed with cocoapods.

In this case:
1. Replace react-native-device-info's podspec from node_modules with that PR one
2. go to project ios folder
3. `rm -rf Pods Podfile.lock` & `pod install` (there is probably better ways to clean Pods project than delete Podfile.lock but it works)
4. Check from installation output that pod get installed with correct version:
```
...
Fetching podspec for `RNDeviceInfo` from `../node_modules/react-native-device-info`
Installing RNDeviceInfo (0.29.1)
...
Pod installation complete! There are 21 dependencies from the Podfile and 33 total pods installed. (numbers from my project)
```
5. Check diff from Podfile.lock, RNDeviceInfo version and checksum should be changed:
```
-  - RNDeviceInfo (0.21.5):
+  - RNDeviceInfo (0.29.1):

-  RNDeviceInfo: e7c5fcde13d40e161d8a27f6c5dc69c638936002
+  RNDeviceInfo: a80e94f282c91e3b3a7d4ae75ed68b11479474b8
```
6. Run app with iOS Simulator, go somewhere where you use react-native-device-info and notice that a) app won't crash, b) used method returns value as expected

I'm not very familiar too with native side (especially iOS) and don't know if this changed really anything with how this library behaviour when you use it, looks like pods have worked with wrong version number too just fine. I have just noticed that if I upgrade react-native-device-info, it won't install Pod with new version like it will do mainly with other dependencies I use, so checked what is wrong here and find that reason is hardcoded version in podspec which won't follow librarys version 🙂

machour:
Awesome, thank you!
This will be shipped in the next release when it's due.

PR #592, opened 2019-03-14
chore: Update url to react-native-community

guhungry:
## Description
Since this project transferred from https://github.com/rebeccahughes/react-native-device-info we should update url accordingly
I check all the url that i changed and verify all links are valid with no direct

## Checklist


machour:
Thank you!

PR #597, opened 2019-03-20
Add Nokia 7.1 to devicesWithNotch list

lasseborly:
Such a rare phone. But my colleague has one. Support the minority 🎉


machour:
Published in 1.0.1, thank you!

Issue #578, opened 2019-02-27
Processor architecture version for Android


hemantasapkota:
## Description

Using the react-native-device-info library, I would like to query for processor architecture version for Android. This can be achieved by invoking ```System.getProperty("os.arch");```. Output example: armv8l armv7l

## Use Case

This property would be useful for telemetry / analytics and for uniquely identifying the device.

mikehardy:
sounds interesting. Can be implemented on iOS as well https://stackoverflow.com/questions/19859388/how-can-i-get-the-ios-device-cpu-architecture-in-runtime

PRs are welcome

hemantasapkota:
I think I might be able to send a PR. What should be the name of this property ? I'm thinking ```osArchitectureVersion```. Any other recommendations ?

mikehardy:
@hemantasapkota that would be awesome. Naming is always opinionated, so this is just opinion. But I'd avoid 'os' because it's a hardware property, not an os-level property. I like the simply idea of just going with supportedABIs to match the Android call https://developer.android.com/reference/android/os/Build.html#SUPPORTED_ABIS and support returning the 1+ array entries sorted by preference right from the device

On iOS it looks like (from the last comment here) we'd just get a single entry https://stackoverflow.com/questions/19859388/how-can-i-get-the-ios-device-cpu-architecture-in-runtime

At the same time, I saw Xamarin didn't return the actual CPU type, they instead tried to make it more useful to the API consumer by instead giving you two flags 32-bit or 64-bit, then arm or x86. Maybe that's interesting as a second couple APIs, but just keeping it easy and going with the array from Build on Android or the single entry from iOS would be cool to me

hemantasapkota:
@mikehardy I'll follow the strategy that you've put down. To summarise:

1. For android, we expose a property called ```supportedABIs`` which returns an array of supported architectures. The first item in the array is the preferred one.

2. For ios, we expose a property called ```cpuType`` as discussed in [1]

3. For windows, perhaps something called ```OSArchitecture``` from [2]

[1] https://stackoverflow.com/questions/19859388/how-can-i-get-the-ios-device-cpu-architecture-in-runtime
[2] https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osarchitecture?view=netframework-4.7.2

mikehardy:
[...]

PR #598, opened 2019-03-21
feat: Added DeviceInfo.supportedABIs()

hemantasapkota:
## Description

Fixed issue #578 

Added `supportedABIs()` which returns a list of supported device cpu types

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ import DeviceInfo from 'react-native-device-info';
 | [getDeviceType()](#getDeviceType)                 | `string`            |  ✅  |   ✅    |   ❌    | ?      |
 | [isAutoDateAndTime()](#isAutoDateAndTime)         | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
 | [isAutoTimeZone()](#isAutoTimeZone)               | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
-| [supportedABIs()](#supportedABIs)                 | `string[]`          |  ✅  |   ✅    |   ❌    | 1.0.1  |
+| [supportedABIs()](#supportedABIs)                 | `string[]`          |  ✅  |   ✅    |   ❌    | 1.1.0  |
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -961,7 +961,7 @@ DeviceInfo.isAutoTimeZone().then(isAutoTimeZone => {
 Returns a list of supported processor architecture version
 
 ```js
-DeviceInfo.supportedABIs(); // [ "x86", "arm64-v8a", "armeabi-v7a", "armeabi" ]
+DeviceInfo.supportedABIs(); // [ "arm64 v8", "Intel x86-64h Haswell", "arm64-v8a", "armeabi-v7a", "armeabi" ]
 ```
 
 ## Troubleshooting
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -253,6 +253,7 @@ import DeviceInfo from 'react-native-device-info';
 | [getDeviceType()](#getDeviceType)                 | `string`            |  ✅  |   ✅    |   ❌    | ?      |
 | [isAutoDateAndTime()](#isAutoDateAndTime)         | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
 | [isAutoTimeZone()](#isAutoTimeZone)               | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
+| [supportedABIs()](#supportedABIs)                 | `string[]`          |  ✅  |   ✅    |   ❌    | 1.0.1  |
```
machour: Will be 1.1.0. Use that to update the Changelog.md file too 🙏

hemantasapkota: done. 🙏

hemantasapkota: sorry, has been fixed now.

example/App.js
```diff
@@ -96,12 +97,12 @@ export default class App extends Component<Props> {
 
   render() {
     return (
-      <View style={styles.container}>
+      <SafeAreaView style={styles.container}>
```
mikehardy: SafeAreaView was new to me! I'm still new at the react-native level of things. This is a nice touch, thanks

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -395,7 +395,9 @@ public String getInstallReferrer() {
     actMgr.getMemoryInfo(memInfo);
     constants.put("totalMemory", memInfo.totalMem);
     constants.put("deviceType", deviceType.getValue());
-
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
mikehardy: I like the API-gate here for the array version of Build.SUPPORTED_ABIS, but this means that in lower APIs (down to API16, React-Native's lower bound) we will not have an entry in constants which will force a logic branch to any consumers of the API, and they won't get any info. I think on the earlier devices you can make an empty array then use Build.CPU_ABI like so https://stackoverflow.com/a/51775541 but add it as [0] in the array

I intend to use this feature almost immediately in my app, down to ABI16 to fetch arch-specific flavors in rn-update-apk so this is important to me, and I'll also serve as a test for it pretty quickly :-). Thanks!

example/App.js
hemantasapkota: glad you like it :)

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
hemantasapkota: fixed.

hemantasapkota: I tested it on a the emulator with API level 16. Fallback seems to work. I'm hoping it'll work as expected on the device too.

mikehardy:
@hemantasapkota this looks great, I'm excited to have this feature in the module. @machour I looked closely and I can't see anything else I'd change in here? I'm not perfect those of course so I'll leave the merge for you in case you want to take another look

hemantasapkota:
@mikehardy thanks for the review. This has been a good experience. looking forward to @machour 's review.

mikehardy:
Whether I got that merge correct or not :pray: - it is my first merge into this repo. Thanks @hemantasapkota for the work and @machour for the help

machour:
👌

react-native-device-info 1.1.0 published 🎉

mikehardy:
I can see in the commit log the way I did it duplicated the message (the title was the same as the commit message) which isn't as tidy as your previous ones. I'll make sure the title is complete/correct next time, but leave the message empty and they'll match up. Not an A+ yet ;-)

Issue #381, opened 2018-04-04
Building without Play Services?


Rudloff:
## Description

It seems that this library can't be built for Android without Play Services:
```
   > A problem occurred configuring project ':react-native-device-info'.
      > You have not accepted the license agreements of the following SDK components:
        [Google Repository].
```

Would it be possible to have an option to build without the `play-services-gcm` dependency? (Even if it means that some functionnalities won't work.)

## Use Case

The Play Services dependency prevents any app using this library from being included in free software repositories like [F-Droid](https://f-droid.org/).

(I see [here](https://github.com/rebeccahughes/react-native-device-info/pull/334) that there was already some discussion about this. The use case back then seemed to be that people wanted to use a different GCM version which is a little bit different.)

Rudloff:
Unfortunately, I don't think I have enough experience with the Play Services to do this without breaking things for people that do use them 😕

Rudloff:
I don't know if it can help you but here is what we used to remove Play services when building an app that uses react-native-device-info 0.14.0:
```bash
sed -i -e '/com.google.android.gms.iid.InstanceID.getInstance/d' node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
sed -i -e '/play-services/d' node_modules/react-native-device-info/android/build.gradle
```

mikehardy:
GCM is only used for the instanceID, #60 should be used instead, during #307 implementation

This is covered in other spots enough I can close this here

jnnkB:
@mikehardy Is this going to be in [v4](https://github.com/react-native-device-info/react-native-device-info/projects/2)?

mikehardy:
We are on v8 now? Why would anything happen on v4? 🤔

jnnkB:
Yeah you right. Just looked at https://github.com/react-native-device-info/react-native-device-info/issues/60 which is still tagged with the v4 version. Should have looked at the project a bit longer.

PR #602, opened 2019-03-25
fix: Make GCM inclusion optional

Jyrno42:
## Description

Fixed issue #381

[GCM](https://developers.google.com/cloud-messaging/) has been deprecated by google. This PR gives control over to the consumer whether they want to include GCM in their app or not. This is achieved by using Java reflections APIs to dynamically instantiate GCM InstanceID object only when it is available.

fixes issue #381

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [X] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [X] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Release Notes
 
+### 2.0.0
+
+ * Remove the `googlePlayServicesVersion` Gradle build config and allow users to include their GCM only if they need it
 
 ### 1.1.0
  * Add `supportedABIs()` (https://github.com/react-native-community/react-native-device-info/pull/598)
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ yarn add react-native-device-info
 ```
 
 > ⚠️ If you are on React Native > 0.47, you must use version 0.11.0 of this library or higher
-> ⚠️ If you want to use `DeviceInfo.getInstanceID` on Android make sure to [include Google play services base in your app](#optional-getinstanceid).
+
+> ⚠️ If you want to use `DeviceInfo.getInstanceID` on Android make sure to [include Google play services base in your app](#getinstanceid).
 
 ## Linking
 
@@ -107,7 +108,7 @@ dependencies {
 }
 ```
 
-* <a name="optional-getinstanceid"></a> **_optional_** in `android/app/build.gradle` if you want to use `DeviceInfo.getInstanceID`:
+* **_optional_** in `android/app/build.gradle` if you want to use `DeviceInfo.getInstanceID`:
 
 ```diff
 dependencies {
@@ -556,7 +557,16 @@ const instanceId = DeviceInfo.getInstanceID();
 
 **Notes**
 
-> ⚠️ To use this on Android you must have [Google play services base installed in your app](#optional-getinstanceid)
+To get correct value on android you must add the following in your `android/app/build.gradle`:
+
+```diff
+dependencies {
+    ...
+    implementation "com.facebook.react:react-native:+"  // From node_modules
++   implementation "com.google.android.gms:play-services-gcm:+"
+}
+```
+
 > See https://developers.google.com/instance-id/
 
 ---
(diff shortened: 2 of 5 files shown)


Jyrno42:
Technically this will be a breaking change so I used version 2.0.0 in the changelog - let's discuss though.

mikehardy:
I am generally a fan keeping with deprecation fixes and dependencies, so I like this, but with the breaking change of course we do need to think about it. I'll review this in a bit and read up on things, but in general with Firebase I think GCM is dead but FCM is alive and we can grab the instance ID that way, so rather than just deprecating we could also potentially bring in the new dependency in the same optional way and still populate the instance id?

mikehardy:
https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId#getInstanceId()

mikehardy:
In general the code changes seem sane, and it is nice to remove a deprecated dependency from the module build code, even if users still need the dependency in their code. The docs for that seem good. I would still like to see the firebase part added rather than just taking out the GCM part - looks like you just do a static getInstance() method call then getInstanceId() and there's nothing more to it. The react-native-firebase project (from invertase) has a great firebase starter app that makes it pretty trivial to try the thing out (I just used it, and while I had an issue with APIs lower than 18, and it wasn't a full react-native 0.58 forward port it should be fine to check this functionality) - https://github.com/invertase/react-native-firebase-starter

Jyrno42:
Seems reasonable. I do wonder whether the firebase id usage should be the default behavior or an explicit opt-in? Since there might be users who might want to keep getting GCM ids while they have firebase installed as a dependency for some other reason.

mikehardy:
that's a good point. Maybe the names shouldn't be conflated, and as long as it is a breaking change it should be getGCMInstanceId() and getFCMInstanceId()? I think all of this works better as an opt-in, I have one app that is all Firebase, one that is 100% open source (and can't bundle play services) etc, so there are lots of cases where the flexibility of opt-in seems useful

mikehardy:
This part is out of my league (I'm a beginner at iOS, my background is Android native) but here's the implementation pointer for iOS (if for no other reason that I'll know where to find it later) https://developers.google.com/instance-id/guides/ios-implementation

mikehardy:
hey @Jyrno42 before you get any farther looking into firebase, I want to take a step back. I'm not sure this really belongs in the code at all. The PR for it originally was lightly defended (#47) and never got an iOS implementation I don't think (#60).

#358 contains a proposal accepted in principal to nuke the google-proprietary GCM/FCM instanceID and go with a regular UUID, but the actual implementation is still native.

I see no reason not to do what is proposed here: https://github.com/react-native-community/react-native-device-info/issues/60#issue-163782044 and just be done with as a breaking change.

That is no criticism of your work here, it's the right direction I just think if we're going to break we should go all the way and have a getUniqueId() and clearUniqueId() (that's important for user control...) cleanly in javascript and shed all the dependencies.

What do you think?

mikehardy:
@machour I meant to reference you in the above, you were active on #358 and I think now that we have semver if there is ever a time to clean the unique id thing up, this is it - and before I (accidentally) dig in deeper by exchanging the play services GCM dependency for a firebase dependency

Jyrno42:
Ok. Ping me if I should continue with this or close it.

mikehardy:
I have not heard dissent, so I'm going to close this. If people want the firebase device id (which is what this feature exposes really, assuming everyone updates from GCM to FCM) then they should integrate firebase. 

And alternatively for a proprietary unique id that we provide, we have tracking issues.

I believe (correct me if wrong!) that this will satisfy your actual goal at least, which is that react-native-device-info will be google / proprietary dependency free with 2.0, and the unique id crowd should be satisfied as well

PR #605, opened 2019-03-28
fix: get correct ios build version

### LINKED ISSUE #209: getBuildNumber() returns 1 when my build # in xcode is 4
screenshot 2017-08-29 20 26 08

mikehardy:
This would maybe fix #209 but it needs testing, and a changelog entry if so


mikehardy:
unnecessary, can't reproduce

PR #406, opened 2018-05-18
Support 'dom' Platform.OS for react-native-dom

peacechen:
## Description

This PR adds support for `react-native-dom`, a more comprehensive library to run RN apps in the browser.  The change is a simple one-liner addition to check whether `Platform.OS === 'dom'`.

@machour 
It looks like there's a lot of cruft in this PR, but it's because my original PR was never merged.  I've sync'd with master and merged in order to submit another PR.

## Compatibility

Web / DOM

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
rest N/A

PR #606, opened 2019-03-28
chore: Refresh example to react-native@0.59.2 + react-native-device-info@master

mikehardy:
The example isn't shipped with the module so I don't think it needs a changelog entry, it's basically a gigantic programmatic piece of documentation

I made a couple tiny changes to the refresh script, then the rest is just the result of running it

I'll let this sit for a couple days for consideration and in the absence of disagreement, I'll merge it in

### DIFF
--- a/example/.flowconfig
+++ b/example/.flowconfig
@@ -24,7 +24,6 @@
 [libs]
 node_modules/react-native/Libraries/react-native/react-native-interface.js
 node_modules/react-native/flow/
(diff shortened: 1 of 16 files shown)


mikehardy:
One of the things fixed here - setting the example to master - almost tripped me up while testing #406 - the upgrade to RN0.59.x is a side benefit but having the example go to master is very useful

PR #484, opened 2018-08-28
Fixed correct root project props

Thunderbird7:
## Description
to use `rootProject.hasProps` instead `root.hasProps` to get variables from root project

Fixed issue #<issue-number>

Added `yourNewMethodName()` that allows ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
* [ ] I added the documentation in `README.md`.
* [ ] I mentionned this change in `CHANGELOG.md`.

PR #607, opened 2019-03-28
chore: convert gradle config to modern style

mikehardy:
this is largely based off personal experience plus a look at react-native-share
Also Obsoletes #484

This is what modern gradle configs look like in modules, and I don't believe this has any compatibility impact. Builds cleanly and runs the example for me


mikehardy:
@machour with the example pointed to master now (I just rebased this so that change is present on this branch), this is trivial to check out and test

mikehardy:
Sorry - that's not quite true, you'll want to change the package reference in the example to 'github:mikehardy/react-native-device-info#gradle-safeExtGet' then npm install and react-native run-android to test it

machour:
I won't be able to test this soon and I'd hate to slow you down.
I would just advise doing a first release with what have already been merged, then a new release with this change, so that if people start to complain about build problems, we'll know exactly where to look.
Wdyt ?

mikehardy:
I love that idea. I think it's all ready on master for a minor bump

PR #610, opened 2019-03-28
feat: Add preferred Languages function

fjtrujy:
## Description

This PR exposes the functions to return the preferred languages set up by the user in the Android and iOS devices.
As a user, you can customize what are your preferred languages. Right now, `react-native-device-info` just return the first value from this list, but there are scenarios where we need to have the whole preferred languages in order to see if they fit with the supported language that the `react-native-app` support.

Added `getPreferredLocales()` that returns the list of preferred languages set up in the Android or iOS device.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [✅] I have tested this on a device/simulator for each compatible OS
* [✅] I added the documentation in `README.md`
* [ ✅] I mentioned this change in `CHANGELOG.md`
* [✅] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [✅ ] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Release Notes
 
 ### 1.3.0
- * feat: Add support for preferred languages function (https://github.com/react-native-community/react-native-device-info/pull/610/files)
+ * feat: Add support for preferred languages function (https://github.com/react-native-community/react-native-device-info/pull/610)
 
 ### 1.2.0
  * feat: Support 'dom' Platform.OS for react-native-dom (https://github.com/react-native-community/react-native-device-info/pull/406)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Release Notes
 
+### 1.3.0
+ * feat: Add support for preferred languages function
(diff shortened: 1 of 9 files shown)


mikehardy:
This looks fantastic and is something I was thinking about. My next step is usually to test it and add it to the example (I haven't added that as a step in the PR template yet) to make sure it works but I have read the code and it is minimal + clean. We should be able to merge this as soon as I've run a local test to prove it

Thank you!

mikehardy:
I cleaned up the conflict caused by my own work with the 1.2.0 release

mikehardy:
Looked fantastic in testing, I just added the example bit then cleaned up the changelog. Thank you! I see you have a couple branches, if you have anything else, send it in and I'll see what I can do

mikehardy:
react-native-device-info 1.3.0 published 🎉

mikehardy:
https://github.com/react-native-community/react-native-device-info/issues/627

PR #611, opened 2019-03-28
docs: add maintainers information describing current processes

mikehardy:
@machour here is another quick docs one for you - I tried to write down everything I remembered about the exact processes (enough to stumble through them) as well as some friendly advice for future people? Did I make any mistakes? forget anything vital?


machour:
Great!

PR #618, opened 2019-04-02
Add the 3th generation iPad pro to the device name list

nbennink:
## Description
I've added the 3rd generation Ipad pro to the list with available devices. This fixes a problem where it just returns "iPad" for these devices.

I've specifically added 3th generation to the name of the device. This way you can actually tell the difference between the multiple versions of the iPad Pro. This is a must have because it's the only way to tell the difference between the iPad pro with physical home button (generation 1 and 2) and the iPad pro without physical home button (generation 3). I kept the name of generation 2 the same for now but I can rename it for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Release Notes
 
+### Next:
+ * Added the 3rd generation iPad pro to the deviceNamesByCode list (https://github.com/react-native-community/react-native-device-info/pull/618)
+	
+**iOS warning:**  The list with device names (returned by `getModel()`) is maintained by the community and could lag new devices. It is recommended to use `getDeviceId()	` since it's more reliable and always up-to-date with new iOS devices. We do accept pull requests that add new iOS devices to the list with device names.
+
 ### 1.4.1
  * fix: repair flow types from #436 - 'Object' vs 'object'
 
--- a/README.md
+++ b/README.md
@@ -636,6 +636,8 @@ const maxMemory = DeviceInfo.getMaxMemory(); // 402653183
 
 Gets the device model.
 
+**iOS warning:**  The list with device names is maintained by the community and could lag new devices. It is recommended to use `getDeviceId()	` since it's more reliable and always up-to-date with new iOS devices. We do accept pull requests that add new iOS devices to the list with device names.
+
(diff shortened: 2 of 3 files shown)


mikehardy:
Thank you for the contribution!

This looks like the right thing to do given the current style, but I have to admit, it seems silly that we are maintaining this database - it doesn't seem like the best use of our time trying to keep up with apple product launches.

How would you feel, as someone clearly consuming this information, about converting to a not-written-yet parallel API that just gave you the unfiltered device information, as described here: https://github.com/react-native-community/react-native-device-info/issues/585#issuecomment-478345235

It seems like the logic you'd have to do on the friendly names we are returning to handle edge cases would not be that much different if you were just doing the lookup in your own table, then we could stop trying to keep up with Apple.

If you think that could work I would take a PR for that and merge/release it immediately. Otherwise I guess it's more table updates here, forever ;-)

nbennink:
Yes I agree, I'm actually fine with a function that just returns the device id. I can maintain the list with device name in my own project if I need it (although I highly doubt it, since you can check on the device ID as well).  The only downside is that I think that some users still want to get the device name from this library.

mikehardy:
It appears that device id is already available as getDeviceId() so you might be set then. I understand users might want to get the model mapping here but it seems like a race we can't win. I need to think about it a bit more to see who is actually consuming the info or if there is a better way. There may be no better way and this is how we need to do it

mikehardy:
I slept on it and here's what I'm thinking:
1. the docs for getModel should contain a warning for iOS that it is community-maintained, it might lag releases, and getDevice is the most reliable, but we accept PRs
2. I will freely merge getModel PRs but won't update it myself, I think this is a community benefit and doesn't take a lot of effort as long as the community understands what they are getting with the API

I think that's a good balance

So for this one I need a changelog entry (just make it "next" for the version) and if you could add the warning on the getModel API like above, that would help. Then I'll merge this in and do a release

nbennink:
I've added a note to the changelogs and readme. The same problem is with `getNotch()`, which is also maintained by the community.

I can also add a console.warn when using `getModel()`, but I'm not sure if we want that now. Maybe in the future when you want to deprecate the function.

PR #639, opened 2019-04-12
Xiaomi Brand name Capital corrected

BeauCarrel:
## Description

Hello, I noticed that "hasNotch()" does not work on my Xiaomi Mi 8, I had a look why and noticed that the brand name return Xiaomi but the brand name of all Xiaomis in "devicesWithNotch" is xiaomi, so I corrected this and have tested on my Xiaomi Mi 8, hasNotch now works.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [ ] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -184,39 +184,39 @@ const devicesWithNotch = [
     model: 'X21 UD',
   },
   {
-    brand: 'xiaomi',
+    brand: 'Xiaomi',
     model: 'MI 8',
   },
   {
-    brand: 'xiaomi',
+    brand: 'Xiaomi',
     model: 'MI 8 Explorer Edition',
   },
   {
-    brand: 'xiaomi',
+    brand: 'Xiaomi',
     model: 'MI 8 SE',
   },
   {
-    brand: 'xiaomi',
+    brand: 'Xiaomi',
     model: 'MI 8 UD',
   },
   {
-    brand: 'xiaomi',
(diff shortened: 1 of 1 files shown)


mikehardy:
Hello! Thanks for the contribution - I'm a bit worried about the quantity of devices changed vs the number of reports we have of this as a problem. Are you sure it affects the other devices as well? Did this change with an update recently or something? I'm just worried we'll fix your device for your version but break all the others.

Could you alter this PR to not mess with the names but instead change the comparison function to convert to lower-case and compare that way? I really want to merge this but I want to make sure we don't break other people accidentally...

machour:
Agreed @mikehardy, comparison should be done in lowercase for brand & models.

BeauCarrel:
Hi guys, I went to implement a lowercase comparison and realised that it was already implemented, I then found out that the Xiaomi telephones were added in the 1.4.3 update (my project being on a slightly earlier version) and this provided the false positive making me think my fork actually did something useful.
My first PR, beginner's mistake ;) I'll go ahead and close this PR

mikehardy:
haha - I'm laughing because I've done that before (fixed an already fixed issue). No harm of course and a beginners mistakes mean you've begun, which is where all sorts of wise sayings start ;-). Cheers

PR #642, opened 2019-04-17
Add permissions for Android

jpr-longgame:
## Description

Fixed an issue where gradle build would barf on missing permissions for WiFi and phone numbers.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

* [n/a] I added the documentation in `README.md`
* [n/a] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [n/a] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Next
 
+ * fix: Update Android permissions for contacts, WiFi status
+
 ### 1.5.0
  * feat: Add `getBuildId` method to gets build number of the operating system. (https://github.com/react-native-community/react-native-device-info/pull/640)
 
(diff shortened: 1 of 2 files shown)


mikehardy:
Hi there!

I can see what you're after I think, but this will not be integrated as it will be default merge into projects and result in them potentially building and submitting apps with sensitive permissions to the Play store.

It is the responsibility of developers to add the required permissions to their own manifest.

I am open to suggestions on developer tool messaging (gradle warning suppression, user documentation etc) but I would like to see an issue opened with a clear description of exactly what the issue is prior to implementing something

PR #646, opened 2019-04-25
feat: implement hasSystemFeature() method for Android devices

markhomoki:
## Description
Added `hasSystemFeature(feature: string)` that allows to check whether a specific system feature is available for Android devices.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [ ] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Next
 
+### 1.6.0
+ * feat: implement hasSystemFeature() method for Android devices (https://github.com/react-native-community/react-native-device-info/pull/646)
+
 ### 1.5.1
  * chore: Update deviceinfo.js entry for xioami mi 8 lite (https://github.com/react-native-community/react-native-device-info/pull/644)
 
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ import DeviceInfo from 'react-native-device-info';
 | [isAutoDateAndTime()](#isAutoDateAndTime)         | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
 | [isAutoTimeZone()](#isAutoTimeZone)               | `boolean`           |  ❌  |   ✅    |   ❌    | 0.29.0 |
 | [supportedABIs()](#supportedABIs)                 | `string[]`          |  ✅  |   ✅    |   ❌    | 1.1.0  |
+| [hasSystemFeature()](#hassystemfeaturefeature)    | `Promise<boolean>`  |  ❌  |   ✅    |   ❌    | ?      |
 
 ---
 
@@ -967,6 +968,8 @@ Tells if the device is currently in landscape mode.
 const isLandscape = DeviceInfo.isLandscape(); // true
 ```
 
+---
+
 ### hasNotch()
 
 Tells if the device has a notch.
@@ -976,6 +979,8 @@ Tells if the device has a notch.
 ```js
 const hasNotch = DeviceInfo.hasNotch(); // true
 ```
+
+---
 
 ### getDeviceType()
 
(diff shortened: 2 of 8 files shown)


mikehardy:
Interesting...I can understand the need definitely.
I feel like we will have a large and growing set of APIs similar to this though, if we put this in.

How about we go larger, and instead expose the basic API? https://developer.android.com/reference/android/content/pm/PackageManager.html#getSystemAvailableFeatures()

You could return the whole array of available features - which could be very generally useful to the community - and easily pick out if the system had the feature you needed (fire tv)

What do you think? I'd love to merge a change that exposed all available features...

markhomoki:
Good idea, initially that what I was looking for. I'm happy to do another PR tomorrow instead of this.

markhomoki:
Updated and squashed into 1 commit. Please review and let me know what you think.

mikehardy:
This certainly looks more generally applicable - your use case is nailed. My use case is actually to get the list of features though, rather than querying one by one, I'd really like to see the whole list emitted and have the consuming library check if the key they want is there.

I can see your API as being nice sugar on top of that but for the truly general case, could you add a parallel API that just emits the whole set of strings using https://developer.android.com/reference/android/content/pm/PackageManager.html#getSystemAvailableFeatures() ?

I'll trade you that change request (since I can tell your use case is settled already) for me doing the web polyfill + changelog + example app update :smile: maybe we have a deal?

markhomoki:
That's fine too. Are you happy then to keep 2 methods within the repo? One to look for a specific feature (as it is my current PR), and the other one to return an array of features? I'll start working on today and update this thread once I finished.

markhomoki:
@mikehardy Please see the updated code, getSystemAvailableFeatures() method has been added as well. Let me know if you're happy with the implementation.

mikehardy:
Yeah I'm fine with 2 APIs in general - thank you! The code looks good on re-read just now, I'll pull it, and run it today and hopefully merge it + release it. 💪

mikehardy:
Strange - when I run this on my emulator I see a 'null' at the end of the feature list:

```

         'android.hardware.telephony',
         null ],
```

but when I log out the native code it has an entry, just one that looks a touch different:

```

2019-04-29 13:23:11.136 7343-7532/com.example I/System.out: Feature info is: FeatureInfo{4978608 android.hardware.telephony v=0 fl=0x0}
2019-04-29 13:23:11.136 7343-7532/com.example I/System.out: Feature info is: FeatureInfo{18ab1a1 glEsVers=2.0 fl=0x0}
2019-04-29 13:23:11.137 7343-7531/com.example I/ReactNativeJS: loaded info
```

So I'm looking at that really quickly, but - it runs! it looks good in my testing except that, and I'm looking at it now...

markhomoki:
Ah okay. Maybe the f.name needs to be wrapped in an if statement. I havent seen null in my case but it can vary based on the device.

mikehardy:
Oh seriously !?! Look at the FeatureInfo.toString implementation, they have a specific case for "glEsVers", and it was the one that spits out null:

```java
    @Override
    public String toString() {
        if (name != null) {
            return "FeatureInfo{"
                    + Integer.toHexString(System.identityHashCode(this))
                    + " " + name + " v=" + version + " fl=0x" + Integer.toHexString(flags) + "}";
        } else {
            return "FeatureInfo{"
                    + Integer.toHexString(System.identityHashCode(this))
                    + " glEsVers=" + getGlEsVersion()
                    + " fl=0x" + Integer.toHexString(flags) + "}";
        }
    }
```

So, yeah, I'm inclined to just do a null check on name but not try to replicate their case-specific logic, as checking OpenGL information is a separate, subtle topic that would need to be it's own API https://developer.android.com/guide/topics/graphics/opengl#version-check

Just doing a null check on f.name with a continue-if-null workaround would be perfect

mikehardy:
react-native-device-info 1.6.0 published 🎉

@markhomoki  flawless victory :-), thank you

markhomoki:
That's great. Thanks for the update. I found some inconsistency in the ordering in the README.md. I might create another PR and fix it.

Issue #214, opened 2017-09-06
getBuildNumber() returns string in iOS and number in Android


grifotv:
It would be nice to keep it consistent. Should the method `getBuildNumber()` return a string or a number?

adyanced:
In my case the returned buildNumber is alway added by 0x100000. So versionCode 6 will be displayed as  1048582. I read the code and find it's not dute to react-native-device-info, maybe a bug of android sdk.
My project's info:
    compileSdkVersion 25
    buildToolsVersion "25.0.3"

leethree:
Agree it should be consistent. Returning a number on Android doesn't match both the documentation and the flow typings.

ericlewis:
Ideally, this should always return a number I think. In case developers would like to operate on it's value.

mikehardy:
Just merged the change that makes them both string, will go in v3

PR #648, opened 2019-04-26
Fix android build number type inconsistency

ericlewis:
## Description

Fixed issue #214

Parse the return integer from PackageManager in to a string, also returns "0" by default instead of 0. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [x] I updated the web polyfill (`web/index.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Notes
 
 ### Next
+ * fix: Android `getBuildNumber` now returns a string (https://github.com/react-native-community/react-native-device-info/pull/648)
 
 ### 1.5.1
  * chore: Update deviceinfo.js entry for xioami mi 8 lite (https://github.com/react-native-community/react-native-device-info/pull/644)
--- a/README.md
+++ b/README.md
@@ -345,13 +345,9 @@ Gets the application build number.
 const buildNumber = DeviceInfo.getBuildNumber();
 
 // iOS: "89"
-// Android: 4
+// Android: "4"
 // Windows: ?
 ```
-
-**Notes**
-
-> There is a type inconsistency: Android return an integer instead of the documented string.
 
 ---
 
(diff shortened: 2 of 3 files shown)


ericlewis:
Tested on Pixel 2 API 28 emulator. Should be pretty simple since it is just an integer conversion.

mikehardy:
Very nice - the callstack react-native bug-squish-a-thon bears fruit in my repo :-). As it happens I was recently all over cross platform versioning. I am confident this change is correct. It is also of course trivial.

However, I will be the first to admit my javascript is weak as I'm an old Java dinosaur just getting up to speed now that Javascript has real language capabilities ;-). So let me ask a question - it is my understanding that with type coercion built in to javascript, people that have been using this as an integer will see continue having this coerced to a string and thus it is actually somewhat backwards-compatible. Is that understanding correct?

Additionally I believe the types are already documented as string, so people using typescript et al will be unaffected (if anything, happier)

I just want to double-check as this would be the difference between flagging it as breaking change or just a patch.

Thank you!

ericlewis:
If people have been using this as an integer, say: buildNumber == 5 or buildNumber > 5. It will continue working just fine, since the JavaScript with coerce the string to numbers in those cases. Those who have been using it like this: buildNumber === 5 will have issues, since it matches both type & value. 

Sent via Superhuman iOS ( https://sprh.mn/?vip=ericlewis777@gmail.com )

mikehardy:
Ok - in that case this is minor, but is a breaking change so it goes in the 2.0 bucket and has to sit unfortunately. We normally release changes super quickly, so I'm sorry to say that, but we have a breaking change queue we have done most of the thinking work on already and I want to batch them up. So consider this "accepted" just pending unrelated work. Thank you very much for the contribution and the '===' clarification

mikehardy:
At long last! Rolling through the breaking-change stack now. Thanks again for this

PR #649, opened 2019-04-28
chore: fix pointer types and macro usage based on XCode warnings

mikehardy:
## Description

Fix iOS build issues - there were two incompatible pointer types, and we were not following convention in a macro usage

This makes it so when you include this module it builds cleanly

This should be backwards compatible and just a patch version

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
* [ ] I added the documentation in `README.md`
* [ ] I mentioned this change in `CHANGELOG.md`
* [ ] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [ ] I updated the web polyfill (`web/index.js`)

PR #652, opened 2019-04-30
fix: building issue on tvOS / header on iOS

markhomoki:
## Description
Fixed a build issue on tvOS in xcode 10.2. tvOS Header Search Paths were missing and couldn't find the `<React/RCTUtils.h>`. Also, `batteryState` is not available for tvOS and xcode failed to build, so I updated the if statement.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Next
 
+### 1.6.1
+ * fix: fix building issue on tvOS / headers on iOS (https://github.com/react-native-community/react-native-device-info/pull/652)
+
 ### 1.6.0
  * feat: implement hasSystemFeature() method for Android devices (https://github.com/react-native-community/react-native-device-info/pull/646)
 
--- a/ios/RNDeviceInfo.xcodeproj/project.pbxproj
+++ b/ios/RNDeviceInfo.xcodeproj/project.pbxproj
@@ -157,6 +157,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
 				en,
 			);
 			mainGroup = DA5891CF1BA9A9FC002B4DB2;
@@ -324,6 +325,14 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_STYLE = Automatic;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
+				HEADER_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(BUILT_PRODUCTS_DIR)/usr/local/include",
+					"$(SRCROOT)/../../React/**",
+					"$(SRCROOT)/node_modules/react-native/React/**",
+					"$(SRCROOT)/../react-native/React/**",
+					"$(SRCROOT)/../../../node_modules/react-native/React/**",
+				);
 				OTHER_LDFLAGS = "-ObjC";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = appletvos;
@@ -351,6 +360,14 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_STYLE = Automatic;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
+				HEADER_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(BUILT_PRODUCTS_DIR)/usr/local/include",
+					"$(SRCROOT)/../../React/**",
+					"$(SRCROOT)/node_modules/react-native/React/**",
+					"$(SRCROOT)/../react-native/React/**",
+					"$(SRCROOT)/../../../node_modules/react-native/React/**",
+				);
 				OTHER_LDFLAGS = "-ObjC";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = appletvos;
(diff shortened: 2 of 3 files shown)


mikehardy:
I am very excited about the header paths thing, I was in the area but hadn't fixed it yet. Does this fix it only for the tvOS target or for all targets, because it was a problem for the iOS build as well. And incidentally you can snag yourself some stackoverflow reputation easily for this one https://stackoverflow.com/questions/55790714/lexical-or-preprocessor-issue-in-react-native-while-building-ios-app-react-rctu

secondly, bummer about the tvOS build fail - that is unexpected though, we had been thinking tvOS would return unknown, not fail the build. This mean that tvOS won't even show the warning though - do you develop for tvOS and have access to it? We'd like the warning to show up in dev mode for tvOS as well, but we obviously don't want the build to break

markhomoki:
Well, I never had any issue with Header Search Paths on iOS with react-native-device-info, so this fix is only for tvOS. 

Regarding the build fail, we develop for tvOS and there is no batteryState for TV devices. I'm not sure if it's necessary to show a warning for TV as it quite obvious that there are no battery in TVs, but if you would like to I can spend some time to add the overlay without breaking the build.

mikehardy:
If you think it's obvious to tvOS users then I am okay with not having the warning. I still like it for iOS at least. I definitely have problems with the same header search path issue in iOS - it builds but there is the error in XCode if you inspect, and I'm not alone as indicated in the stackoverflow, so if you could extend it to the iOS targets it should have everyone happy - I'd pull local, test and publish today

markhomoki:
Thanks, I'll look into and see if I can fix it. Earliest can be tomorrow.

mikehardy:
Since I'd need to test it locally anyway I went ahead and tried applying the same header block to the iOS targets and now for the example project I have a clean build (if I also apply #649 pointer type fixes). So I'd say this is basically good to go, I'll push my extra fix for iOS targets to your master then merge it in and publish. Thanks!

mikehardy:
Okay - I have published this as v1.6.1 and I bundled #649 in as well. You are likely the most interested party and are using XCode, so when you get a chance if you could double-check it to make sure #649 didn't blow anything up, and that the work here was merged well, I would really appreciate it. Cheers!

markhomoki:
Amazing. Thanks. I'll run some tests tomorrow and will let you how it goes.

markhomoki:
Upgraded and all worked fine. Builds without a problem on tvOS and Android TV. 

However, when I specifically open `Libraries/RNDeviceInfo.xcodeproj/RNDeviceInfo/RNDeviceInfo.m` in xcode it shows a warning next to `#import <React/RCTUtils.h>` saying `Lexical or Preprocessor Issue Group` - `React/RCTUtils.h' file not found`. 

It still builds fine and does NOT show the error during or after the build process, only shows when I open that file in xcode.

mikehardy:
I see that as well on the direct code open in the project directly, and likewise it builds fine (and without warnings other than upgrading to suggested settings now). I'm at a loss as to what's causing it but I'm happy there appears to be progress. 🤷‍♂ If after rumination you can think of what to do to make it better, I am obviously happy to merge things. Thanks!

markhomoki:
I think there are still some dependency or header path issues. It builds fine, it's not a blocker. Anyway, when I got some time I'll look into and try to fix.

AlexSugak:
Hi! I am experiencing the same error (`'React/RCTUtils.h' file not found`) trying to build the app with react-native-device-info 1.6.1.

Here is my podfile:
```
target 'MyApp' do
    use_frameworks!
    pod 'Fabric'
    pod 'Crashlytics'

    pod 'React', :path => '../node_modules/react-native', :subspecs => [
    'Core',
    'CxxBridge', # Include this for RN >= 0.47
    'DevSupport', # Include this to enable In-App Devmenu if RN >= 0.43
    'RCTText',
    'RCTNetwork',
    'RCTWebSocket', # Needed for debugging
    'RCTAnimation', # Needed for FlatList and animations running on native UI thread
    'RCTImage', # Needed for react-native-svg
    # Add any other subspecs you want to use in your project
    ]

    # Explicitly include Yoga if you are using RN >= 0.42.0
    pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'
    # Third party deps podspec link
    pod 'DoubleConversion', :podspec => '../node_modules/react-native/third-party-podspecs/DoubleConversion.podspec'
    pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.podspec'
    pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
    pod 'react-native-video', :path => '../node_modules/react-native-video/react-native-video.podspec'

    pod 'GoogleSignIn', '~> 4.4.0'
    pod 'RNSVG', :path => '../node_modules/react-native-svg'
    pod 'react-native-orientation', :path => '../node_modules/react-native-orientation'

    pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'
end
```

Tried unlinking/linking, re-installing pods etc.
Also tried adding React header search paths to the ios project itself, but with no luck.

Any other suggesstions?

AlexSugak:
OK, if I disable `use_frameworks!` flag, it works fine.
I wonder why ¯\_(ツ)_/¯

mikehardy:
Hmm - that might be a clue to where the config is messed up - useful information, thank you

cinder92:
i don't have `use frameworks!` and still appearing same error, can't build my app

PR #654, opened 2019-05-04
Constant that is not constant

MatheusAlvesA:
## Description

I found a problem. The function getFreeDiskStorage() never updates the returned value when necessary. If you are performing downloads with you App this info must update or can cause bugs.

**Note:** My code editor(Atom) removed some white spaces automagically.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❓     |
| Android |    ✅     |
| Windows |    ❓     |

## Checklist


mikehardy:
Hi there! Thank you for this - we are just about to restructure most of the API to do as this does (stop using constants and go to Promises, because this is one API of many that should be returning dynamic values. I'm going to tag this for #307 as that's when we will convert them all. I apologies for not just reviewing and merging this immediately but we want to batch all the breaking changes since so we don't have library consumers doing code modifications release after release

MatheusAlvesA:
No problem.
I've never made a pull request before, so just the fact that my code is useful already makes me happy.

**^^**

mikehardy:
I apologize for not merging this as part of the move to a Promise-based API but I'm doing the work right now and it's harder to fix the conflicts than it is to just migrate this as part of migrating the rest of them. Thanks for your patience either way - this feature will be in v3

Issue #659, opened 2019-05-13
ios: podspec fails to find latest version when installing via podfile


chanpod:
## Summary

Just like the title says. Linking wasn't working so I tried doing it manually in the podfile but I was getting the error

> fatal: Remote branch 1.6.1 not found in upstream origin

I updated the podspec file in the node_modules to (I just added a v before the tag)

`require 'json'

package = JSON.parse(File.read(File.join(__dir__, 'package.json')))

Pod::Spec.new do |s|
  s.name         = "RNDeviceInfo"
  s.version      = package['version']
  s.summary      = package['description']
  s.license      = package['license']

  s.authors      = package['author']
  s.homepage     = package['repository']['url']
  s.platform     = :ios, "9.0"
  s.ios.deployment_target = '9.0'
  s.tvos.deployment_target = '10.0'

  s.source       = { :git => "https://github.com/react-native-community/react-native-device-info.git", :tag => "**v**#{s.version}" }
  s.source_files  = "ios/**/*.{h,m}"

  s.dependency 'React'
end`

and pod install worked and RNDevice info was successfully added to the build. (And I stopped getting the RNDeviceInfo is null error)

mikehardy:
Interesting. You don't provide the actual commands you ran, the issue template if followed would result in us knowing your full set of commands and the versions you used, so we could reproduce. I can't reproduce now, but we have been using 'v' tags since repository inception and this is the first report. On the other hand, I changed that line in the podspec just a month ago, so I suppose I did it incorrectly. If I can't reproduce via actual steps I don't know for sure though.

Could you include your steps to reproduce?

Also, I will assume the reproduction is easy (this seems like an obvious bug...) so I will say you can easily (since it's just the podspec) use the github web interface to edit that file and add your proposed change, then propose a PR, that way you'll get credit for the fix - well-earned, and should help everyone in the future as RN0.60 will move to Pods by default

chanpod:
Let me get this successfully archived and pushed to the store for testing and I'll take a step back, clean all my pod files and node_modules and do it fresh to make sure it wasn't a fluke. I updated from RN 0.56 to 0.59 and I've been playing the "get it to build and archive" dance for a while. So it might have just been me. :)

chanpod:
ah, I see what happened. I don't know about this stuff to know if this is normal or what I reported really is a bug, bug I was using

[...]

PR #660, opened 2019-05-13
fix: added v to tag name in podspec

chanpod:
When missing the v for the tag name, directly referencing the podspec file can result in a "tag not found" error when doing a pod install.

## Description

Fixed issue #659 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |   ❌     |

## Checklist

* [x] Ran pod install and was able to build and archive the project


mikehardy:
published as 1.6.2 🚀

PR #661, opened 2019-05-14
Updated Readme.md - include pod install note post-link

### LINKED ISSUE #353: iOS automatic linking is not working.

ishan123456789:
Updated automatic linking which requires extra step in case of IOS, helps fix #353

## Description

Fixed issue #353
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅ |
## Checklist

* [ ✅ ] I added the documentation in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ react-native link react-native-device-info
 ```shell
 rnpm link react-native-device-info
 ```
-*For IOS*
+*For iOS users using Pods*
 You still need to run `pod install` after running the above link command inside your `IOS` folder.
 
 ### Manual
--- a/README.md
+++ b/README.md
@@ -37,12 +37,13 @@ yarn add react-native-device-info
 ```shell
 react-native link react-native-device-info
 ```
-
 (or using [`rnpm`](https://github.com/rnpm/rnpm) for versions of React Native < 0.27)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,12 +37,13 @@ yarn add react-native-device-info
...
 react-native link react-native-device-info
 ```
-
 (or using [`rnpm`](https://github.com/rnpm/rnpm) for versions of React Native < 0.27)
 
 ```shell
 rnpm link react-native-device-info
 ```
+*For IOS*
+You still need to run `pod install` after running the above link command inside your `IOS` folder.
```
mikehardy: I like the location - this is the exact spot I think - and it's almost perfect for me, but since this section is not Pod-specific and we still provide direction for pod or non-pod, could you say *for iOS users using Pods* ? Then it's merge time :-) - thank you!

ishan123456789: K will do that

mikehardy:
Excellent - I'm sure that will help people in the future

PR #628, opened 2019-04-05
chore: Update and cleanup project

friederbluemle:
## Description

A bunch of project cleanups and updates. The changes are in individual commits for easier review. Please let me know if there are any questions with individual patches. See explanations below for more information:

### Clean up lock files

A project should have **either** a `yarn.lock` file **or** a `package-lock.json` file. The existence of `yarn.lock` in the root folder suggests usage of Yarn which results in the following warning:

    package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

The React Native sample project in the `example` folder _should_ use Yarn (`yarn.lock` *already* added in 85f17d01f7824ab29ff4e698a5c2101e50e9d134) - I deleted the redundant `package-lock.json` file here.

### Remove extra Gradle wrapper

There was a misplaced Gradle wrapper sitting in the `android/` folder. The executable bit of `gradlew` was not even set. Removed the unnecessary scripts together with the wrapper jar file.

### Update .gitignore

Consolidated the .gitignore files and removed some duplications

### Update example project

Minor UI updates and improvements, as well as committed the automated updates to the Xcode project file by CocoaPods (`pod install`) to avoid a dirty working directory after running this necessary set-up step.

### Simplify Podfile and add **Podfile.lock**

There were some redundant/extraneous declarations/statements in `Podfile`. Also adds the missing `Podfile.lock`.

### Update refresh-example script

Added the `pod install` step in the `ios` folder of the example project. Running `./refresh-example.sh` should now result in **zero changes** (if the dependencies are up-to-date). Also reordered a couple of steps to remove the temporary directory earlier (pod install requires `Podfile` that was saved in the temp folder).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅ *   |

\* Due to lack of devices/OS I was not able to test the Windows part.

Tested on Android and iOS devices and emulator/simulator. Opened and successfully ran the project both in Android Studio 3.3.2 and Xcode 10.2

PR #662, opened 2019-05-14
Remove yarn.lock file

dratwas:
## Description
I removed `yarn.lock` file which is not needed since this package uses `package-lock.json`


mikehardy:
Hi there! I like this, but I'm in the process of getting a larger cleanup into shape with #628 with @friederbluemle and it includes this change, so I'm going to resolve this in favor of the ongoing work there. I think that one's nearly ready to go so this should be in master shortly

PR #663, opened 2019-05-15
chore: integrate react-native-community/eslint standards

mikehardy:
## Description

react-native-community has some nascent standards, and [the use of lint is one of them](https://github.com/react-native-community/.github/issues/9).

I've patterned this off the [work in react-native-maps](https://github.com/react-native-community/react-native-maps/pull/2862), netinfo etc - [guidance here](https://github.com/facebook/react-native/tree/master/packages/eslint-config-react-native-community)

Basically, this installs the requisite lint package and since we had no lint standard before, I just extended the community standard in in package.json

Our prettierrc was already similar, so there were very only 4 errors to fix

To ensure we don't skew over time, I installed lint-staged and husky, and configured the module to run eslint in fix mode as a commit hook

PR #664, opened 2019-05-15
feat: add isLocationEnabled, getAvailableLocationProviders

jasongaare:
## Description

- Adds `isLocationEnabled()` that allows the determination of whether or not the user has location services enabled or disabled on their device. Returns a simple `boolean` value. Useful for determining if the app has access to location services. Sometimes, even though location permissions have been granted, the user has still turned off GPS on their device. 

NOTE: this is NOT checking the permission to access location, rather the availability of location services. 

- Adds `getAvailableLocationProviders()` that returns an object of platform-specific location providers/services and boolean values as to whether they are available or not.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -358,6 +360,28 @@ public void getSystemAvailableFeatures(Promise p) {
     p.resolve(promiseArray);
   }
 
+  @ReactMethod
+  public void hasLocationServicesEnabled(Promise p) {
+      boolean locationEnabled = false;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT){
+          try {
+              int locationMode = Settings.Secure.getInt(this.reactContext.getContentResolver(), Settings.Secure.LOCATION_MODE);
```
mikehardy: Can you use the mode with a default value since we default to Settings.Secure.LOCATION_MODE_OFF anyway? That way we can avoid the possibility of exception, get rid of the try/catch (and what made me think about it) not have to decide whether to printStackTrace or p.reject() etc

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -358,6 +360,28 @@ public void getSystemAvailableFeatures(Promise p) {
     p.resolve(promiseArray);
   }
 
+  @ReactMethod
+  public void hasLocationServicesEnabled(Promise p) {
+      boolean locationEnabled = false;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT){
```
mikehardy: Can you add the case for Build.VERSION_CODES.P and https://developer.android.com/reference/android/location/LocationManager.html#isLocationEnabled() also?

These APIs move so fast (to my way of thinking), but P does already have about 33% of the market, worth moving into their APIs as long as we check versions

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mikehardy: also - it's peppered in the code a bit, but in javascript 'this.' references are redundant

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mikehardy: Hmm - now that I think about it, I wonder of having a return shape that wasn't a simple boolean might be interesting? For instance, a lot of apps are interested in what level of GPS is allowed - GPS-only, A-GPS, wifi only etc. What if there was a return shape that was an map, with name and enabled pairs? Then a wrapper that give you a simple boolean if you wanted, by scanning the main return map for any entry that had a value of true? Then the API on Android would be https://developer.android.com/reference/android/location/LocationManager.html#getProviders(boolean) and it is since API1

default/index.js
```diff
@@ -46,5 +46,6 @@ module.exports = {
   getPowerState: () => Promise.resolve({}),
   hasSystemFeature: () => Promise.resolve(false),
   getSystemAvailableFeatures: () => Promise.resolve([]),
-  getCameraPresence: () => Promise.resolve(false)
+  getCameraPresence: () => Promise.resolve(false),
+  hasLocationServicesEnabled: () => Promise.resolve(false)
```
mikehardy: this is very minor - but I'm working on react-native-community lint standards with #663 (not merged yet) and it should have a trailing comma

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -511,6 +512,12 @@ - (NSDictionary *)powerState
     resolve(@(isCharging));
 }
 
+RCT_EXPORT_METHOD(hasLocationServicesEnabled:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
+{
+    BOOL locationServicesEnabled = [CLLocationManager locationServicesEnabled];
```
mikehardy: Thinking along the lines of return object shape, if it wasn't a simple boolean we could easily expose the rest of the availability items here as well: https://developer.apple.com/documentation/corelocation/determining_the_availability_of_location_services - a lot of those are more related to abilities as opposed to whether something is on or not, so the wrapper that allowed for API consumers to also just have a simple yes/no would have to be a little more picky for iOS, but in general just returning a map of those should be easy? (I'm more used to android where it's just a WritableMap with some pushStrings and  you're set

mikehardy:
Hey - this is great! I could use this in an app for work immediately, so I'm motivated to get this in. What do you think about the general idea of expanding the return object shape to contain all the location-related info we can get out of the devices and then just wrapping for a yes/no for those that want it? I don't intend to complicate too much, I just think that it would pretty easy to add all the method calls as long the area was open for work anyway

jasongaare:
For you - I can look into adding those shapes. It could follow a similar pattern to the `getProviders` API you linked, where you can add the `enabledOnly` boolean to return the yes/no, otherwise a shape of available providers. 

The shapes would certainly need to deviate per-platform then, right? Would you prefer it still be a singular function, with different shapes returned based on platform?

mikehardy:
Yeah, the hardware stuff always deviates a bit. I like the idea of a rollup yes/no API especially for  that reason - we will have familiarity and can scan the varies-by-platform return for the items that let us answer API consumers.

But for the next-level of API consumers that need to behave in very specific ways I do really like the idea of binding all the underlying APIs and exposing all the hardware status where possible. In that vein I'd actually not even send in the boolean true, I'd send in false on android for getProviders and just send them all back, let API consumers sort it out in their apps

jasongaare:
Okay I took a stab at moving in the direction I think you are suggesting. Hopefully I'm not too far off base 😬

Anyway, I added a `getAvailableLocationServices` function and returned objects with keys and then boolean values for each key, for the respective platforms. 

Let me know what you think, and I can make your other changes as well and we can get this merged once it feels right.

mikehardy:
Yeah, that's pretty much it - that looks nice. Between the boolean and the everything options, if people don't have their GPS cases handled now, we can't help them :-) - the new API needs the rest of the administrivia to match (readme, empty polyfill, example API add, types) but it seems fine as is to me. There's a possibility to slim the native portion by re-implementing the boolean "is it enabled" version as a filter on the getAvailableLocationServices return, but I'm not sure if that's an improvement or not so I'll say "implementor's choice" and you can leave it like this or alter if you want. Thanks!

jasongaare:
I have made changes and this is ready for review again. I decided to stick with two separate methods: `isLocationEnabled()` and `getAvailableLocationProviders()` due to my experience debugging where the Android getProviders API did not return any providers until _after_ I had allowed location permission on my device, which I do suppose makes sense, but I didn't see it documented anywhere. In turn, that made me concerned that the new API level 28 function `isLocationEnabled` would not work either until location permissions were granted, but that didn't seem to be the case in my testing.

For my use case, it is helpful for me to know before prompting for location permissions whether location services are turned on for the device (hence the original PR content). Thus, different methods.

I did teeter on making `getAvailableLocationProviders` into two platform-specific functions, since they do return fundamentally different information... but ultimately decided having two similarly named function per platform was more confusing that having two different response objects per platform.

Let me know if I missed anything updating the docs, types, etc. Thanks!

mikehardy:
That's very interesting re: permissions and return values. Sounds like the right decision to me definitely, and nothing beats code that's actually tested.

I just re-read it and it looks completely clean. I'm going to pull it locally to test but I can't see a thing to change by looking. Excellent work, thank you!

mikehardy:
Tested out clean but I also saw what you saw w.r.t. availability without permissions, and I added permissions to the manifest in the example so users (and I) could do a better test. All checked out and I was able to exercise it completely that way on android. Checking iOS then I'll ship it

mikehardy:
v1.8.0 is out! :sparkles: :rocket:

PR #669, opened 2019-05-21
types: make the typescript location / power returns specific

mikehardy:
These are just type fixes from a couple recent PRs

I'm still getting better at Typescript so I didn't realize these would be problematic as submitted until I attempted to use them in a project and there were complaints about the return values.

Since they have known types from the native code, I made them specific.

I don't know enough about flow to do those but would love if someone did, or maybe we can leave them generic.

Also, I am not sure about the breaking nature of this or not - but I think if we are going from less-specific to more-specific it shouldn't break anyone?

### DIFF
--- a/deviceinfo.d.ts
+++ b/deviceinfo.d.ts
@@ -3,6 +3,19 @@
 
 export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Unknown';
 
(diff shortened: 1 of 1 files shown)


mikehardy:
https://github.com/react-native-community/react-native-device-info/releases/tag/v2.0.2

PR #670, opened 2019-05-21
chore: move ios device model dictionary from native to javascript

marekchen:
## Description

avoid update native code when add new ios devices

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist


mikehardy:
Interesting! I like this idea, thank you contributing this. Let me think about it a little bit and test it, but for a library that starts in javascript and just binds in to native, it does seem to make sense to have the frequently-changed parts more accessible to users, yes.

mikehardy:
This seems to work well for me, thanks for the contribution!

mikehardy:
https://github.com/react-native-community/react-native-device-info/releases/tag/v2.0.2

PR #678, opened 2019-05-28
fix: adding conditions for tvOS location providers

vdimitrovv:
Adding conditions to only check if location services is enabled, because tvOS doesn't support `significantLocationChangeMonitoringAvailable`, `headingAvailable` and `isRangingAvailable`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| tvOS     |    ✅     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -400,12 +400,19 @@ RCT_EXPORT_METHOD(isLocationEnabled:(RCTPromiseResolveBlock)resolve rejecter:(RC
 
 RCT_EXPORT_METHOD(getAvailableLocationProviders:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
+#if !TARGET_OS_TV
     resolve(@{
               @"locationServicesEnabled": [NSNumber numberWithBool: [CLLocationManager locationServicesEnabled]],
               @"significantLocationChangeMonitoringAvailable": [NSNumber numberWithBool: [CLLocationManager significantLocationChangeMonitoringAvailable]],
               @"headingAvailable": [NSNumber numberWithBool: [CLLocationManager headingAvailable]],
               @"isRangingAvailable": [NSNumber numberWithBool: [CLLocationManager isRangingAvailable]]
               });
+#endif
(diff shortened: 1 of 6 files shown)


mikehardy:
Supporting location and power information in combination with tvOS is difficult! This is the 2nd or 3rd snag we have had there. @markhomoki - you are also an active member here using tvOS, does this look good to you? It looks good to me but I know you have a vested interest...

@vdimitrovv your second fix here (at least, that's just looking at recent changes), thank you 🏆  😄

markhomoki:
This looks good to me apart from the double if statements. Could be done with if/else.

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -400,12 +400,19 @@ - (NSDictionary *)powerState
 
 RCT_EXPORT_METHOD(getAvailableLocationProviders:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
+#if !TARGET_OS_TV
```
markhomoki: Can you change to if else, please?

mikehardy: Seems reasonable, thanks for the review @markhomoki !

mikehardy:
Haha and you already did it ;-)

mikehardy:
I'll pull these locally and build them just to make sure but I don't see any reason why I won't merge and release shortly. Thanks again to both of you

vdimitrovv:
Change done 👍

For further context, both @markhomoki and I are on the same team :-)

mikehardy:
> Change done 👍
> 
> For further context, both @markhomoki and I are on the same team :-)

Oh well that's just sneaky, you're ganging up on me ;-). I'm really happy we have tvOS users active though - it's a big help as it's not a target I build normally

mikehardy:
Sorry for the delay on this - just did testing and noticed my example app suffered from the same "what's this tvOS thing?" problem, where it just expected battery status etc to be there.

I fixed that and re-pushed to your branch so the PR was good (though please double-check) and I'll publish 2.0.3 in a minute

mikehardy:
...and it's out - seems fine to me in iOS and tvOS but please holler if I messed anything up :-), thanks again

PR #680, opened 2019-06-03
perf(Android): getConstants() in thread to parallelize load, speed up for most

mcuelenaere:
## Description

RNDeviceModule.getConstants() can take a lot of time (~300ms on a Galaxy S7), so try to run it as soon as possible in the background so that by the time it gets called, the constants have already been calculated and we can just return the map.

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -15,6 +15,7 @@ import android.content.res.Configuration;
 import android.location.LocationManager;
 import android.net.wifi.WifiManager;
 import android.net.wifi.WifiInfo;
+import android.os.AsyncTask;
 import android.os.Build;
 import android.os.Environment;
 import android.os.StatFs;
@@ -50,6 +51,7 @@ import java.util.TimeZone;
 import java.lang.Runtime;
 import java.net.NetworkInterface;
 import java.math.BigInteger;
+import java.util.concurrent.ExecutionException;
 
 import javax.annotation.Nullable;
 
@@ -61,11 +63,24 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
   DeviceType deviceType;
 
-  public RNDeviceModule(ReactApplicationContext reactContext) {
+  Map<String, Object> constants;
+  AsyncTask<Void, Void, Map<String, Object>> futureConstants;
+
+  public RNDeviceModule(ReactApplicationContext reactContext, boolean loadConstantsAsynchronously) {
     super(reactContext);
 
     this.reactContext = reactContext;
     this.deviceType = getDeviceType(reactContext);
+    if (loadConstantsAsynchronously) {
+      this.futureConstants = new AsyncTask<Void, Void, Map<String, Object>>() {
+        @Override
+        protected Map<String, Object> doInBackground(Void... args) {
+          return generateConstants();
+        }
+      }.execute();
+    } else {
+      this.constants = generateConstants();
+    }
   }
 
   @Override
@@ -408,9 +423,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     return sharedPref.getString("installReferrer", null);
   }
 
-  @Override
-  public @Nullable
-  Map<String, Object> getConstants() {
+  private Map<String, Object> generateConstants() {
     HashMap<String, Object> constants = new HashMap<String, Object>();
 
     PackageManager packageManager = this.reactContext.getPackageManager();
@@ -514,5 +527,21 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
       constants.put("supportedABIs", new String[]{ Build.CPU_ABI });
     }
     return constants;
+  }
+
+  @Override
+  public @Nullable
+  Map<String, Object> getConstants() {
+    if (this.constants == null && this.futureConstants != null) {
+      try {
+        this.constants = this.futureConstants.get();
+      } catch (InterruptedException e) {
+        return null;
+      } catch (ExecutionException e) {
+        throw new RuntimeException(e.getCause());
+      }
+    }
+
+    return this.constants;
   }
 }
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java
@@ -11,12 +11,21 @@ import java.util.Collections;
 import java.util.List;
 
 public class RNDeviceInfo implements ReactPackage {
+  private boolean mLoadConstantsAsynchronously;
+
+  public RNDeviceInfo() {
+    this(false);
+  }
+
+  public RNDeviceInfo(boolean loadConstantsAsynchronously) {
+    mLoadConstantsAsynchronously = loadConstantsAsynchronously;
+  }
 
   @Override
   public List<NativeModule> createNativeModules(ReactApplicationContext reactContext) {
     List<NativeModule> modules = new ArrayList<>();
 
-    modules.add(new RNDeviceModule(reactContext));
+    modules.add(new RNDeviceModule(reactContext, mLoadConstantsAsynchronously));
 
     return modules;
   }


noomorph:
What are your measurements and how did you measure them?

mcuelenaere:
This was measured by adding a ReactMarker listener and printing out durations of each step, where we focus here only on the GET_CONSTANTS part of RNDeviceInfo.

Below are 3 measurements from before and after the patch, each starting from a cold environment (app was killed). As you can see, the GET_CONSTANTS duration gets reduced to almost nothing.

BEFORE:
```
   +325.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
   +229.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +6.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo

   +288.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
   +265.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +3.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo

   +286.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
   +262.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +4.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo
```

AFTER:
```
   +276.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
     +0.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +1.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo

   +382.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
     +0.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +1.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo

   +280.000000 ms --                    GET_CONSTANTS [START]: RNDeviceInfo
     +0.000000 ms --                CONVERT_CONSTANTS [START]: RNDeviceInfo
     +1.000000 ms --                  CONVERT_CONSTANTS [END]: RNDeviceInfo
     +0.000000 ms --                      GET_CONSTANTS [END]: RNDeviceInfo
```

Your mileage may vary, since this depends on how much time is in between the `RNDeviceModule` constructor being called and the `getConstants()` method being invoked (+ scheduling and thread synchronization overhead).

tl;dr: if you have a decent amount of RN modules being loaded in your app (and your device has multiple cores), this should have a positive effect.

mikehardy:
just marking it as "needs-changes" because that's a tag I review periodically even when PRs go to sleep

mcuelenaere:
@mikehardy the behaviour is optional now and disabled by default

mikehardy:
How did I miss that? Let me re-read, because that sounds perfect...

mikehardy:
Oh you just did it haha, thought I was losing my mind. This does sound perfect now, I will re-read and probably merge

Thanks for the contribution!

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -61,11 +63,28 @@
...
   DeviceType deviceType;
 
+  Map<String, Object> constants;
+  AsyncTask<Void, Void, Map<String, Object>> futureConstants;
+
   public RNDeviceModule(ReactApplicationContext reactContext) {
+    this(reactContext, false);
+  }
+
+  public RNDeviceModule(ReactApplicationContext reactContext, boolean loadConstantsAsynchronously) {
```
mikehardy: This makes it optional, but now for this method to be invoke with async load, does it need some corresponding hook in RNDeviceInfo.java? Like RNDeviceInfo needs two constructors with a similar flag, it stores the state, then when [RNDeviceInfo.createNativeModules is called](https://github.com/mcuelenaere/react-native-device-info/blob/55dc03dc465786f745ee260532891121da18547f/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceInfo.java#L16) it sends the flag through? 

Otherwise I'm not sure how the new async style is invoked since the default (and only currently available) style from linking is to just new RNDeviceInfo and there are no other control flow entrypoints to toggle the behavior https://github.com/mcuelenaere/react-native-device-info/blob/55dc03dc46/example/android/app/src/main/java/com/example/MainApplication.java#L27

mikehardy:
I pulled this locally and have it ready for testing - I think it needs one more tweak after I requested it be optional then it's probably good to go

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mcuelenaere: Oh, you're right. It's because locally I'm using Turbo Modules and thus creating the RNDeviceModule myself. I'll adapt `RNDeviceInfo` to also take a constructor argument.

mcuelenaere:
@mikehardy `loadConstantsAsynchronously` is now propagated from `RNDeviceInfo` to `RNDeviceModule`

mikehardy:
I'll repull and retest - super-speedy thanks!

mikehardy:
https://github.com/react-native-community/react-native-device-info/releases/tag/v2.1.0 🚀

ItsNoHax:
@mikehardy @mcuelenaere First of all thanks a lot to both of you for the fix. I would just like to point out that the change log and the README should reflect these changes. I'm still unclear on whether I should use true or false in the constructor.

mikehardy:
@ItsNoHax the release notes do mention it? https://github.com/react-native-community/react-native-device-info/blob/master/CHANGELOG.md#210 (they say use `true`)

I suppose the README for install could be updated, it might help to explain that the constructor even has an optional boolean argument, and what it means. Happy to take a PR for that - you can edit it directly using the GitHub UI and propose some words that you think would help people?

ItsNoHax:
Done, let me know what you think: https://github.com/react-native-community/react-native-device-info/pull/697

I left it under the main installation header as I was afraid people might miss it when linking automatically.

zhao812:
expo how install react-native-device-info

elyalvarado:
Hi guys, how will this work with RN>=0.60 auto loader?

mikehardy:
@jslok to be honest I am not sure - but I am curious myself as I am just migrating my work project to RN60 now.

My current understanding is that to initialize a module with a non-default constructor (specifically: one that takes an argument) in the auto-linking world, you have to use a react-native-config.js to *disable* autolinking, then manually link it on the platform you disabled.

@thymikee is that correct? auto-linking only works with default constructors?

I initially merged this patch as a new option and did not set the deferred method to default but assuming my understanding of auto-linking is correct and it will only work with a default constructor, I'd accept a patch that reversed which style was default now that it has been released a while and shown to be working.

thymikee:
@mikehardy disabling linking in `react-native.config.js` affects both autolinking and `link`. 

Passing a custom constructor is possible through [`android.packageInstance`](https://github.com/react-native-community/cli/blob/master/docs/dependencies.md#platformsandroidpackageinstance):

```js
module.exports = {
  dependencies: {
    'react-native-device-info': {
      platforms: {
        android: {
          packageInstance: 'new RNDeviceInfo(true)'
        }
      }
    }
  }
};
```

mikehardy:
@thymikee I really appreciate both your responsiveness, and your spot-on response, thank you! I promise I'll be good at auto-linking soon :-) 

Okay then, @jslok I would happily take either a PR that made the parallel version default + a quick doc snippet on how to disable it (with the snippet above I think but reversed to false) or a doc enhancement that had this snippet showing how to enable it.

I think it's ready for default though, I believe more will benefit from it's being enabled than will be harmed so I would prefer just enabling by default

Dror-Bar:
I don't have a file named react-native.config.js. This is unclear from the read me if I should create it and/or where should I create it.

mikehardy:
@Dror-Bar if you wanted to propose a quick readme PR that made it clear to you that would help. Specifically addressed at the confusion: the file does not exist in the repository, but if you do *not* want the parallel initialization, you will need to create the file, and it should contain the stanza prescribed in the react-native-config.js in the readme

If you are okay with the new default / parallel initialization is okay for your project, you do not need to do anything

My decision to merge this and alter the default was based on the assumption it was reasonably tested (so would not break anyone) and that in general it was a performance increase for everyone or at least did not harm things. If this is not true for your project I would love to know why in order to see if my assumption was correct - and I'd apologize for regressing your project also, that definitely wasn't intended

PR #685, opened 2019-06-06
fix(Android): Avoid NullpointerException on install referer listener

nhnam:
action is Nullable, so we call equals from action is not safe

```
java.lang.RuntimeException · Unable to start receiver com.learnium.RNDeviceInfo.RNDeviceReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
```

## Description

Fixed issue #<issue-number>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceReceiver.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceReceiver.java
@@ -10,7 +10,7 @@ public class RNDeviceReceiver extends BroadcastReceiver {
   @Override
   public void onReceive(Context context, Intent intent) {
     String action = intent.getAction();
-    if (action.equals("com.android.vending.INSTALL_REFERRER")) {
+    if ("com.android.vending.INSTALL_REFERRER".equals(action)) {
       SharedPreferences sharedPref = context.getSharedPreferences("react-native-device-info", Context.MODE_PRIVATE);
       SharedPreferences.Editor editor = sharedPref.edit();
       editor.putString("installReferrer", intent.getStringExtra("referrer"));


mikehardy:
I get your point about nullable action, and the proposed change is in general just good defensive coding practice, so I'm probably going to just merge this but I must ask :-), this actually happened to you? Because it should never happen, we filter for just that action so it should only ever be that action? https://github.com/react-native-community/react-native-device-info/blob/master/android/src/main/AndroidManifest.xml#L10

In some research it appears that someone is directly calling your action, which explains why it might start with the action being null: https://stackoverflow.com/questions/15048883/intent-getaction-is-returning-null

So I can believe this happens, and I think the patch is correct, I'm just surprised it's needed but hasn't come up before - it looks like it's more than a year old https://github.com/react-native-community/react-native-device-info/pull/344

nhnam:
Hi, sorry for my bad fix. As you knew, the `action` is marked `@Nullable` so basically it could be null. We could either check for nullability first of use my way to just avoid the exception. 
Here goes I explain how could it happen. 
We're using `react-native-device-info` in our project and, another one which also has a receiver which is `react-native-appsflyer`
They notice here [https://support.appsflyer.com/hc/en-us/articles/207032126-AppsFlyer-SDK-Integration-Android](https://support.appsflyer.com/hc/en-us/articles/207032126-AppsFlyer-SDK-Integration-Android)
Please refer to #MultipleInstallBroadcastReceiver and #SingleInstallBroadcastReceiver. which I think it related to my issue, the NullPointerException.
So there are 2 options: 
`If you do not have a receiver listening on the INSTALL_REFERRER, use SingleInstallBroadcastReceiver` ==> we use this since we didn't know that `react-native-device-info` also has another Receiver. If we know that there are more BroadcastReceiver in the app, we would use the MultipleInstallBroadcastReceiver.

So, basically the appsflyer will catch INSTALL_REFERRER first then broadcast back to others Receiver (react-native-device-info), and in this case, the intent.getAction() is null.

mikehardy:
makes sense! I'm happy to merge it, just wanted to understand more. I really appreciate the details on the explanation

mikehardy:
You can point to master in your package.json until this is released - hopefully I can put this out shortly though

nhnam:
Sure, thanks @mikehardy!

mikehardy:
v2.1.1 just went out, thanks again for the contribution :smile:  :rocket:

PR #686, opened 2019-06-07
Readme: Clarify location of RNDeviceInfo.xcodeproj

joeyyang:
Under the iOS "Manual Linking" section, there is no file with the extension `.xcodeproj` in the location specified in the README. I have updated this section to specify the correct location of this file.


mikehardy:
:-) obvious oversight there and I can see how it would be confusing - thanks for fixing this!

Issue #691, opened 2019-06-17
Docs: Incomplete instructions for iOS Manual Linking using CocoaPods


ashwinpilgaonkar:
## Summary

|             |     |
| ----------- | --- |
| Affected OS | iOS   |

## Current behavior

The instructions in the README under Linking -> Manual -> iOS (with CocoaPods)
specify that `pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'` should be added to the Podfile, after which `pod install` should be run.

However, running `react-native run-ios` after the above mentioned steps, results in the build failing and the below given error appearing on the console.

```
<rn-project-name>/node_modules/react-native-device-info/ios/RNDeviceInfo/RNDeviceInfo.m:13:9: 'React/RCTUtils.h' file not found

#import <React/RCTUtils.h>
```

## Expected behavior

Running `react-native run-ios` after following the instructions in the README under Linking -> Manual -> iOS (with CocoaPods) should result in a successful build.

PR #692, opened 2019-06-17
docs: Update Podfile contents under Linking -> iOS (via CocoaPods)

ashwinpilgaonkar:
## Description

Fixed issue #691 

Added a few more contents to the Podfile to result in a successful build.

Adding `pod 'React', :path => '../node_modules/react-native'`
prevents CocoaPods from installing a "React" Pod in the `ios/Pods` directory.

Not adding this line results in multiple versions of "React" being installed in the same project (one in `/node_modules` and the other in `ios/Pods`) which probably results in a conflict.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,31 @@ You still need to run `pod install` after running the above link command inside 
 <details>
     <summary>iOS (via CocoaPods)</summary>
 
-Add the following line to your build targets in your `Podfile`
+Add the following lines to your build targets in your `Podfile`
 
-`pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'`
+```
+pod 'React', :path => '../node_modules/react-native'
+
+# Explicitly include Yoga if you are using RN >= 0.42.0
+pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'
+
+pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'
+  
+# React-Native is not great about React double-including from the Podfile
+post_install do |installer|
+  installer.pods_project.targets.each do |target|
+    if target.name == "React"
+      target.remove_from_project
+    end
+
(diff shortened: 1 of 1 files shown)


mikehardy:
I requested a change but thank you for noticing this and proposing a change to help others - react-native + iOS + pods is very confusing to folks that aren't very experienced with it and this should help

Janain:
之前SDK 是26，不升级的情况下，androidx运行报清单文件   android:appComponentFactory="androidx.core.app.CoreComponentFactory"
        android:debuggable="true"


编译异常

mikehardy:
@Janain this repository is developed in English - sorry, I have no idea what you are trying to say and github doesn't have any sort of translation facility built in

ashwinpilgaonkar:
> A more durable documentation change would be to point at the Podfile in the example and say "when done, your Podfile should look roughly like [the standard Podfile from the react-native template](https://github.com/react-native-community/react-native-device-info/blob/master/example/ios/Podfile), but with this line added..."

I agree with you, but the problem is that the iOS example Podfile contains many additional pods (i.e DoubleConversion, glog, Folly and React subspecs) which may not be necessary for everyone and may result in some users being confused about what to add and what not to.

The lines I added to the README are the bare essential pods required to be added, in order to get this library working with iOS + CocoaPods.

ashwinpilgaonkar:
> I requested a change but thank you for noticing this and proposing a change to help others - react-native + iOS + pods is very confusing to folks that aren't very experienced with it and this should help

It is indeed!
Actually, the reason I proposed a change was because I was confused myself (initially) and it took me quite a while to figure out how it all works. :smile:

mikehardy:
That's a fair defense - thanks again for taking the time to submit this

ashwinpilgaonkar:
> That's a fair defense - thanks again for taking the time to submit this

No problem!
Thanks for merging.

PR #702, opened 2019-06-21
fix: add missing devices to devices with notch list

Mhany95:
* Added some devices that have notches to the list of devices with notches.


mikehardy:
[v2.1.3](https://github.com/react-native-community/react-native-device-info/releases/tag/v2.1.3) :rocket:

Issue #703, opened 2019-06-21
_this.getModel is not a function


fernandofranca:
# Bug

Using the method `hasNotch` causes the `_this.getModel is not a function` error.

## Environment info

React native info output:

```bash
  React Native Environment Info:
    System:
      OS: macOS 10.14.1
      CPU: (4) x64 Intel(R) Core(TM) i3-8100B CPU @ 3.60GHz
      Memory: 31.78 MB / 8.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.15.3 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - ~/.npm-packages/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 26, 28
        Build Tools: 28.0.3
    IDEs:
      Android Studio: 3.3 AI-182.5107.16.33.5314842
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.6.3 => 16.6.3 
      react-native: 0.58.5 => 0.58.5 
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-rename: 2.4.1
      solidarity-react-native: 2.1.2
```

Library version: 2.1.3

## Steps To Reproduce

Just run `DeviceInfo.hasNotch()`

Describe what you expected to happen:

1. Return a boolean value

mikehardy:
tentatively closing this - I can't reproduce - but if I'm wrong of course will re-open for investigation

fernandofranca:
I've got it. 
I was using the `hasNotch` function by reference as DeviceInfo looked like a static class:

```
async function getDeviceInfo() {
  const deviceInfoMap = {
    ...
    applicationName: DeviceInfo.getApplicationName,
    hasNotch: DeviceInfo.hasNotch,
    ...
  }

  let deviceInfo = {}
  Object.keys(deviceInfoMap).forEach((key) => {
    let getterFn
    try {
      getterFn = deviceInfoMap[key]
      deviceInfo[key] = getterFn()
    } catch (err) {
      console.group('getDeviceInfo.js')
      console.log(`Failed to execute getter function for "${key}" key. Error: ${err}`)
[...]

PR #704, opened 2019-06-21
fix: update hasNotch method reference to getModel

fernandofranca:
Fix "_this.getModel is not a function"

## Description

Fixed issue #703

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -344,6 +344,31 @@ const deviceNamesByCode = {
   'AppleTV6,2': 'Apple TV 4K', // Apple TV 4K
 };
 
+function getModel() {
+  if (Platform.OS === 'ios') {
+    var deviceName;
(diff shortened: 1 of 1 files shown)


mikehardy:
Hmm - I don't understand.

Please explain why it works in the example, as is: https://github.com/react-native-community/react-native-device-info/blob/master/example/App.js#L61

I just tested it, fresh checkout of current master (which is equivalent to version 2.1.3 as of now), `cd example && \rm -fr node_modules && npm i && react-native run-ios` and separate Terminal  `react-native run-android`.

Can't reproduce? It works fine for me.

Closing provisionally, but I'm interested

PR #714, opened 2019-07-04
Fix: crash on android api level less than 23

huynhquangtoan:
## Description
 Fix: crash on android api level less than 23

- android.os.Build.VERSION. BASE_OS
- android.os.Build.VERSION. PREVIEW_SDK_INT
- android.os.Build.VERSION. SECURITY_PATCH

**Added in API level 23**


mikehardy:
Oh of course :man_facepalming: - my app even goes down to API16 I just didn't have a chance to test it yet and have not implemented CI here that tests the whole API skew yet.

Thank you very much for noticing this and sending in a patch

mikehardy:
Ok v2.2.1 is live with this patch, thanks again and sorry this affected you

Issue #723, opened 2019-07-16
Export ES6 compatible module


NikhilVerma:
# Feature Request

Export react-native-device-info to be compatible with destructured ES6 imports

## Why it is needed

It will make it less verbose to use React Native device info

## Possible implementation

Just need to add additional exports to the deviceinfo.js

### Code sample

```js
import { getDeviceLocale } from 'react-native-device-info';
```

Taym95:
@mikehardy what do you think? I can work on this!

mikehardy:
I like the idea - how would it affect backwards compatibility? Or more directly: would this be a breaking change? Or would all existing codebases be fine no matter how they were importing?

I am not against breaking changes at all, I just want to know what a release with ES6 compatibility would mean to installed users and I am still coming up to speed on javascript (with it's seemingly infinite variations...) as a language

Taym95:
Yes, will be backwards compatible only old version can not use destructured ES6 imports.

mikehardy:
OK, thanks for explaining it and verifying, I'd be willing to take a pr like this, please be sure to update pr template as this would be another thing to do in PRs right? If people add something

PR #727, opened 2019-07-22
Export react-native-device-info to be compatible with destructured ES6 imports

Taym95:
## Description

Fixed issue #723

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,10 @@ allprojects {
 
 ```js
 import DeviceInfo from 'react-native-device-info';
+
+// or
+
+import { getUniqueID, getManufacturer } from 'react-native-device-info';
 ```
 
 ## API
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -648,239 +648,73 @@ export function getAvailableLocationProviders() {
 }
 
 export default {
-  getUniqueID: function() {
-    return RNDeviceInfo.uniqueId;
-  },
-  getInstanceID: function() {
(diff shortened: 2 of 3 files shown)


mikehardy:
I like the idea but I wish this didn't add essentially a duplicate entry for each function. Is there any way (breaking or non-breaking) to do this without duplicating each function entry? Just trying to keep the barrier as low as possible for people that propose PRs - we already have like 5 things we ask people to do :-)

Taym95:
Yeah, I got your point, but unfortunately, this is the only way to implement destructured ES6 imports!

mikehardy:
ok - I will likely merge but I want to think on it just a bit more by looking at other modules to see how they did it (usually I examine react-native-community/netinfo) - maybe there's nothing to be done but maybe there's an organizational trick I/we could use

Taym95:
Yes absolutely, That is exactly what I did, I checked how `react-native-community/netinfo` implemented it.

mikehardy:
Man, I guess it's just verbose then. Some things javascript is great at. Some things, not so much...module systems... :man_shrugging:

Taym95:
Yes exactly, I removed duplicate function implementation, so now contributor can add function to be exported :
```
export function getExample() {
  return RNDeviceInfo.example;
}
```
then add it to default export :
```
export default {
	//other functions
	...,
	getExample
}

mikehardy:
Thanks for your patience! Big refactor and release coming up as a v3, this will be in it

Issue #709, opened 2019-07-01
Please add missing devices to devices with notch list : Huawei P30 (PRO)


Ludus-Team:
# Feature Request

Please add missing devices to devices with notch list : Huawei P30 (PRO)

Thank you :)

mikehardy:
Please propose a PR to do so that you verify works on the device, and I’ll merge and release

PR #734, opened 2019-07-24
Add Huawei P30 to devices with notch list

Taym95:
## Description

Fixed issue #709

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ❌      |
| Android |     ✅      |
| Windows |     ❌      |

## Checklist

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -89,6 +89,14 @@ const devicesWithNotch = [
   },
   {
     brand: 'Huawei',
+    model: 'P30 Lite',
+  },
+  {
+    brand: 'Huawei',
+    model: 'P30 Pro',
+  },
+  {
+    brand: 'Huawei',
     model: 'Nova 3',
   },
   {


deviceinfo.js
```diff
@@ -87,6 +87,14 @@ const devicesWithNotch = [
...
     model: 'Mate 20 Pro',
   },
+  {
+    brand: 'Huawei',
+    model: 'P30 Lite',
+  },
+  {
+    brand: 'Huawei',
+    model: 'P30 Pro',
+  },
```
machour: Could you please move these additional lines to the bottom (keeping the whole array alphabetically sorted)

Taym95: Hi,  I added it after `Huawei`, it's already sorted alphabetically, right?

Taym95: I copied the array to javascript console and tried `.sort()` function and the array still the same.

mikehardy: I do like that it is alphabetized but I'd be hard-pressed to try harder than actually testing sort() output :-)

PR #745, opened 2019-08-02
Add `@ReactModule` annotation and `NAME` constant

janicduplessis:
## Description

Add `@ReactModule` annotation and `NAME` constant. This will eventually be used by TurboModules annotation processor.

## Compatibility

N/A

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -41,6 +41,7 @@ import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.module.annotations.ReactModule;
 
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


mikehardy:
Seems fine - and thank you! - a couple questions:

- when does TurboModules come in and what's the effect? A pointer to a good doc you'd recommend is sufficient, I can read up :-)
- when did this API come into existence? Are we cutting off compatibility to some older react-native versions by adding this? If so which ones?

janicduplessis:
There is some info about it here https://github.com/react-native-community/discussions-and-proposals/issues/40. Basically this enables using TurboReactPackage more easily, it allows loading native modules lazily at the moment they are used in JS.

The annotation class was added in September 2016 so it should be fine.

mikehardy:
Awesome thanks for both bits of info - I'm traveling so please be patient but I'll double check and likely merge and release, next opportunity. I appreciate the help!

mikehardy:
Thanks for your patience! Shaping up a v3 release with most of the backlog now

mikehardy:
@janicduplessis 
The only thing I don't like about this, is that now Android Studio can't find the import or the ReactModule symbol. Is there something we need to also add in build.gradle in order for Android Studio state to be sane again?

janicduplessis:
Hmm that’s strange, it should work fine. Maybe its a cache issue or something? I’ll try to run the sample from the repo to see if I get the issue too.

mikehardy:
Oh it runs fine - as in you can use it in a project that depends on react-native, but in this repo and react-native-localize at least, if you open <library>/android in Android Studio by itself to work on the library, it can't find the symbol. That's puzzling to me

mikehardy:
@janicduplessis I hope you haven't spent much if any time looking yet - it helps to actually have a devDependency on react-native and include it in the build.gradle if you, you know, want Android Studio to have access to symbols you want to use :man_facepalming:  :sweat_smile: 

All good now.

janicduplessis:
Haha I forgot about this. Awesome!!

PR #746, opened 2019-08-03
Remove obsolete gradle dependencies

se-bastiaan:
## Description

This pull request proposes two changes to the dependencies defined in the native Android code of react-native-device-info.

1. Remove the direct dependency on the Android support library or AndroidX
The native Android code does not utilise any of the functionalities provided by these libraries directly. SupportLib/AndroidX is a dependency of all play-services libraries but you do not need to specify them directly. They were added in #590 to fix some kind of linting problem which I'm unable to reproduce. So that makes the specification here obsolete.
2. Replace the gcm dependency by a direct dependency on `play-services-iid`
The instance id interface that `react-native-device-info` uses for the unique device id on Android is actually part of `com.google.android.gms:play-services-iid` which is a dependency of 
`com.google.android.gms:play-services-gcm` that is currently used. Using the right dependency here will decrease the size of the imported code for everyone using `react-native-device-info`. 
I do not believe this to be a breaking change since the version of both `gcm` and `iid` is the same. If users of the package use any `com.google.android.gms:play-services-*` libraries gradle's depedency resolution should make sure the change is dealt with correctly.

To test this PR I pointed the example app to this branch on my fork but obviously I have not committed these changes. Maybe it would be good to put the package itself in a different subdirectory so that the example app can directly import the local sources instead of using the repository as dependency.

## Compatibility
Not applicable. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Notes
 
 ### Next
+* chore: Remove obsolete gradle dependencies (https://github.com/react-native-community/react-native-device-info/pull/746)
 
 ### 2.3.2
 * fix: load module async by default with option to load sync (https://github.com/react-native-community/react-native-device-info/pull/741)
--- a/README.md
+++ b/README.md
@@ -47,11 +47,6 @@ AndroidX is supported in a non-breaking / backwards-compatible way by using over
   ext {
     // dependency versions
     googlePlayServicesVersion = "17.0.0" // default: "16.1.0" - pre-AndroidX, override for AndroidX
-    compileSdkVersion = "28" // default: 28 (28 is required for AndroidX)
-    targetSdkVersion = "28" // default: 28 (28 is required for AndroidX)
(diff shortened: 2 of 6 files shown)


mikehardy:
@se-bastiaan thank you for investigating and testing this! I'm preparing a breaking change release now to resolve #756 and I've taken this work and implemented it as the first commit there, so this PR won't go through but the commit is 100% based on the work here, I appreciate it

grit96:
This *is* a breaking change because `play-services-gcm` and `play-services-iid` versions differ.

See extract below from https://dl.google.com/dl/android/maven2/index.html:

```
play-services-gcm
16.0.0
16.1.0
17.0.0

play-services-iid
16.0.0
16.0.1
17.0.0
```

`16.1.0` in `play-services-gcm`, but `16.0.1` in `play-services-iid`.

So by setting `googlePlayServicesVersion` to `16.1.0` for pre-AndroidX (as per the README), it actually breaks the Android build with `Could not find com.google.android.gms:play-services-iid:16.1.0.`

PR #750, opened 2019-08-13
removing RNDeviceReceiver by default and updated Readme.md for the same.

Pradeet:
**getInstallReferrer** is not a very used function, hence it doesn't make sense to Add RNDeviceReceiver by default.

In this PR i have removed the receiver entry from AndroidManifest.xml and updated **README.md** regarding the same.

## Description

Fixed issue:
Removing unnecessary broadcast receiver.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [ x] I added the documentation in `README.md`
* [ x] I mentioned this change in `CHANGELOG.md`
* [ x] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [ x] I updated the dummy web/test polyfill (`default/index.js`)
* [ x] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Release Notes
 
 ### Next
+* breaking(Android): removing RNDeviceReceiver by default. (https://github.com/react-native-community/react-native-device-info/pull/750)
 
 ### 2.3.2
 * fix: load module async by default with option to load sync (https://github.com/react-native-community/react-native-device-info/pull/741)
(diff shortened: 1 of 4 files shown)


mikehardy:
Hmm - unnecessary to you is used-in-production by others, so at minimum this is a breaking change. But - what is the harm in having it? As far as I know it has no performance impact on app size, runtime, or permissions surface area. It should not really affect app startup time. So, from the perspective of a request where a breaking change is proposed, I don't see a benefit. I could be wrong?

Pradeet:
Our app goes through security analysis and we were instructed to remove all unwanted broadcast receivers.
Adding unwanted Broadcast receivers can be a security breach for the app.

mikehardy:
Use 'tools:node="remove"' to remove it on merge

There is always a tension between security and features, and your need for security doesn't mean the whole library should lose a feature I don't think.

mikehardy:
https://developer.android.com/studio/build/manifest-merge

PR #751, opened 2019-08-13
Mention that `getUniqueID()` can be different in app extensions

Skn0tt:
## Description

I stumbled over this when implementing a share extension, which has a different bundle identifier than my main app.
Since I used `getUniqueID()` to construct a User-Agent that needed to be stable for Authentication purposes, this broke auth for me.


mikehardy:
Fascinating! This should help save some future developers some unnecessary new gray hairs, thanks!

PR #760, opened 2019-08-31
V3 - Promise API, lots of bugfixes, iOS deprecation fix, API moves to react-native-localize

mikehardy:
Related issue tracking things: #307 

Commits have the description as things happen


mikehardy:
I won't merge this yet, but I will publish a react-native-device-info@next build from it

I expect there are bugs, as I have it running on android and iOS now but I haven't checked the output of certain APIs yet (like memory size etc) to make certain everything is the same

This let's people review though

All review / thoughts are welcome

mikehardy:
This is released.

If anyone wants to try v3.0.0-beta.1 of react-native-device-info  is up tagged on npm as @latest - it's a big ol' breaking change because instead of making your app startup really slow (like .6 seconds slower than it would otherwise) by loading everything in constants in init, it's all Promises now. It also evicts the deprecated WebView stuff that causes App Store warnings. I would especially love a Windows tester, I don't even have a test environment there so I just hacked it and prayed https://github.com/react-native-community/react-native-device-info/releases/tag/v3.0.0-beta.1

chrisheninger:
Pulled `3.0.0-rc.1` into my app– after adjusting the few places I'm using it to expect promises everything is working as intended. 🙂👍

mikehardy:
Excellent! Thanks for reporting in @chrisheninger I really appreciate it. I'm having much the same experience - seems to be working fine but I have a bit of app-bootstrapping re-organization to do in order to handle the Promise's async-ness vs my need for concrete answers. That's to be expected though

Unless I hear from anyone else I will likely publish rc.1 either as-is or just with documentation changes within a day or two, just want to finish fully integrating in my work project and triple-verifying everything first

Cheers

chrisheninger:
Not sure where it belongs (maybe just in this PR is sufficient)– but it might be a good idea to leave some documentation/examples on how to use the library now that it's promise-based.

I initially ran my project after updating (without changing any of my existing code) and some of the render errors that threw as a result were a bit confusing/cryptic. 😅 

Below are a couple examples I threw together– I didn't run them but I think they're correct.

Example using hooks:
```
import React, { useState } from "react";
import DeviceInfo from "react-native-device-info";

const App = () => {
  const [versionNumber, setVersionNumber] = useState("");

  useEffect(() => {
    DeviceInfo.getVersion()
      .then((version) => {
        setVersionNumnber(version);
      })
      .catch((err) => console.error(err));
  }, []);

  return (
    <View>
      <Text>Your app is running version {versionNumber}</Text>
    </View>
  );
};

export default App;
```

Example using class component:
```
import React, { useState } from "react";
import DeviceInfo from "react-native-device-info";

class App extends Component {
  constructor(props) {
    super(props);

    this.state = {
      versionNumber: "",
    };
  }

  componentDidMount() {
    DeviceInfo.getVersion()
      .then((version) => {
        this.setState({ versionNumber: version });
      })
      .catch((err) => console.error(err));
  }

  render() {
    const { versionNumber } = this.state;
    return (
      <View>
        <Text>Your app is running version {versionNumber}</Text>
      </View>
    );
  }
}

export default App;
```

It's important to note that with both of these examples– since we're now promise-based– will first render with empty string and then with the version number once the promise has been resolved. To prevent this from happening, the developer would need to have some sort of loading indicator rendering in place of the text if the versionNumber has not been set in state yet.

Hope that helps someone out. 🙂 

Thanks for all the work you've done on this lib as well as jetifier @mikehardy– that upgrade from RN 0.59->0.60 was a bit rough and your explanations in various issues made everything much smoother. 👍

mikehardy:
Good points all...most of my change is in my bootstrapper, sounds similar.

I already have an example App.js that shows general usage, if you wanted to do a quick PR that showed some alternate styles like you pasted here (clearly flagged as such like "hey, you might use it this way..." to get some of the info in App.js that would be great. Also, I just started a wiki page to document the transition: https://github.com/react-native-community/react-native-device-info/wiki/V2-to-V3-Migration-Guide

Finally, for bootstrapper example I think the right thing is to alter the index.js in the example here https://github.com/react-native-community/react-native-device-info/blob/v3/example/index.js#L9 so that prior to doing the AppRegistry.registerComponent it did the required DeviceInfo stuff and only did registerComponent in a .then() hanging off it - as I think that is the heart of bootstrapping in react-native apps? (that being: don't registerComponent until you have all your must-haves set up?)

mikehardy:
and you're welcome :-), nice to hear the effort helped others. rn60 was a very tough upgrade...but pods and auto-linking - so much better!

chrisheninger:
I didn't see you had an example App.js– what you have in `async componentDidMount` is much more in-depth and does the job perfectly. 👍

mikehardy:
3.0.0-rc.2 is out, with a huge addition - I broke down and added a xxxSync() version of each API.

While ingesting the v3 changes into my work project (which is partially synchronous, and depends on device info during bootstrap) it was very difficult to handle the fully asynchronous API. I was near to having it all working but it was a real mess and I thought I probably wouldn't be the only one having this difficulty.

And [react-native actually lets you be evil and block the JS thread if you really want](https://medium.com/@some_day_man/synchronous-returns-in-react-native-native-modules-453af33d5999). Use responsibly.

I expect this will be the final release candidate unless someone discovers a bug.

I am also still uncertain about Windows support. I have done my best there but I have no react-native-windows toolchains set up, so I was coding blind and haven't even attempted to compile it. Any help appreciated.

mikehardy:
Just waiting on https://github.com/react-native-community/react-native-localize/pull/65 to merge now - once that's merged I'll release

mikehardy:
The react-native-localize PR is merged, just waiting for release now

PR #769, opened 2019-09-09
Fix bug where getModel return undefined because deviceId is undefined

rnnyrk:
## Description

Can't find a related issue, but when testing the RC I found that `deviceId` in `getModel` would be undefined because of a inconsistency of the deviceId const in `getModel`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

All others seem unrelated ;)

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -541,9 +541,9 @@ export async function getModel() {
   if (!model) {
     if (OS === 'ios') {
       let deviceName;
-      let device = await RNDeviceInfo.getDeviceId();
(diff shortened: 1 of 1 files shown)


mikehardy:
Obsoleted by https://github.com/mikehardy/react-native-device-info/commit/a7129c15dcdd79480eec23011db31154cd1ca7df (I wanted to keep the different variable name, but obviously I needed to *use* it) - will be released as rc.5 shortly - thanks for reporting this @rnnyrk !

PR #781, opened 2019-09-13
Added tvOS Compiler Flags

hufkens:
## Description

The latest version doesn't compile for the tvOS target because if the WebKit imports and usage. This PR adds some additional compiler flags so that it works again for tvOS.
 
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| tvOS/iOS |    ✅   |

## Checklist

* [ x ] I have tested this on a device/simulator for each compatible OS

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -15,7 +15,7 @@
 #import "RNDeviceInfo.h"
 #import "DeviceUID.h"
 
-#if !(TARGET_OS_TV)\
+#if !(TARGET_OS_TV)
 #import <WebKit/WebKit.h>
 #import <LocalAuthentication/LocalAuthentication.h>
 #endif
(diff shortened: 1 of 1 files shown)


hufkens:
No worries. Glad to contribute.

mikehardy:
@hufkens I am unfortunately unable to test this, and I really tried! I know you've been active in the tvOS community, so I will just go on faith and merge it

I mentioned the difficulty here https://github.com/react-native-community/react-native-tvos/issues/5#issuecomment-531254964 

The general state of tvOS right now makes it hard to support these things but I'm more than happy to give it a best effort of course, and will merge anything you see that's needed. Thanks!

hufkens:
@mikehardy It's indeed too bad that it will be removed from the core. And it's not really easy to test indeed. Just make sure nothing is broken in iOS. It should be, because I only added flags for tvOS.

mikehardy:
yeah - it seemed obviously correct just looking at it, and I did confirm it was fine on iOS. I actually sort of enjoy seeing things (like, actually with my own eyes seeing things) work on the TV simulator. I've had it running before. So if that tvos repo issue receives some traction I'll probably include an 'exampletvos' sample app as well and make it so testing here is easy.

mikehardy:
@hufkens  released as 3.0.1!

benoitdion:
+1. I'm still pretty unclear on what will have to be included in the new tvos repo. With the way things are working it feels like tvOS checks like the ones added here or even in core (https://github.com/facebook/react-native/pull/26332) will continue to exist if we want to avoid forking hell.

PR #784, opened 2019-09-13
Expose getDeviceType as a static fn

benoitdion:
## Description

This is useful for applications that need to query the device type on the native side before react native has started (e.g.: Application#onCreate)

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -55,6 +55,7 @@ import static android.provider.Settings.Secure.getString;
 @ReactModule(name = RNDeviceModule.NAME)
 public class RNDeviceModule extends ReactContextBaseJavaModule {
   public static final String NAME = "RNDeviceInfo";
+  private static DeviceType deviceType;
 
   public RNDeviceModule(ReactApplicationContext reactContext) {
     super(reactContext);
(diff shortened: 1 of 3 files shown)


mikehardy:
Interesting - direction-wise I was thinking of re-adding the constants construction on init and exposing those to the JS layer, then connected to this you could expose a constants-fetching function that was static. All of the items you contemplate getting I could see as part of the constants fetched during object construction

Does that make sense?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -109,76 +116,10 @@ public boolean isEmulatorSync() {
   }
 
   @ReactMethod(isBlockingSynchronousMethod = true)
-  public boolean isTabletSync() { return getDeviceType() == DeviceType.TABLET; }
+  public boolean isTabletSync() { return constants.isTablet(); }
```
benoitdion: this would likely go away in favor of the constant

benoitdion:
Makes sense @mikehard, updated the PR with an alternate approach. It's still a draft so mainly looking for feedback on the general approach before I clean things up

mikehardy:
Sorry I haven't reviewed this yet - had a crash fix PR come in and my work project is currently in flames in iOS release mode, so I haven't been able to give it a time slice. I'm interested in this though, and I'm sure we can make something work. I appreciate your patience

mikehardy:
Thank you everyone for the effort - I really appreciate it!

Test it please https://github.com/react-native-community/react-native-device-info/releases/tag/v4.0.1-beta.5

benoitdion:
Thanks for the update @mikehardy. The issue with the implementation in v4.0.1-beta5 is that the constants aren't available until after the module has been initialize by the RN system.

Issue #780, opened 2019-09-12
Sync API: undefined is not an object (evaluating 'N.startsWith')


cristianoccazinsp:
# Bug

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 1.59 GB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    npm: 6.9.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 28
      Build Tools: 28.0.3
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.60.5 => 0.60.5 
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    react-native-cli: 2.0.1
```

Library version: 3.0.0

## Steps To Reproduce
Getting the above error on some devices while on production and using the sync API.

My code:
```_deviceString = `${DeviceInfo.getBrandSync()};${DeviceInfo.getModelSync()};${DeviceInfo.getUniqueIdSync()}`;```

The error:
```
Error · Failed to get device string: undefined is not an object (evaluating 'N.startsWith')
```

Device:
```
osName iOS
osVersion 12.4.1
iPad Mini (5 Gen)
```

Any ideas?

rbscott:
It looks like this isn't actually a race condition. The issue is that the variable `device` isn't initialized until you call `getDevice` or `getDeviceSync`. This means if you call `getModel` before calling one of those two methods, the whole thing crashes. This only happens on devices that are not inside of `deviceNamesByCode`. Currently for instance, `iPad11,3` isn't in that list, so if you launch on the _iPad Air 3rd gen_ in the simulator an app will instantly crash.

mikehardy:
Very interesting - thanks for the extra information. I'm staring at the getModelSync() method right now and I can't see how it could crash but I'm must be missing something obvious, I suppose I'll see with your PR :-)

[...]

PR #791, opened 2019-09-17
Fix crash in getModel on iOS. Fixes: #780

rbscott:
## Description
Fix crash in getModel on iOS. Fixes: #780

Add a few extra devices from:
https://www.theiphonewiki.com/wiki/Models

Fixed issue #780

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |
| Windows |    N/A     |

## Checklist

* [n/a] I updated the dummy web/test polyfill (`default/index.js`)
* [n/a] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -262,6 +262,7 @@ const deviceNamesByCode = {
   'iPod4,1': 'iPod Touch', // (Fourth Generation)
   'iPod5,1': 'iPod Touch', // (Fifth Generation)
   'iPod7,1': 'iPod Touch', // (Sixth Generation)
+  'iPod9,1': 'iPod Touch', // (Seventh Generation)
   'iPhone1,1': 'iPhone', // (Original)
   'iPhone1,2': 'iPhone 3G', // (3G)
   'iPhone2,1': 'iPhone 3GS', // (3GS)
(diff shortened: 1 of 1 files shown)


deviceinfo.js
```diff
@@ -544,18 +551,19 @@ export async function getModel() {
         deviceName = deviceNamesByCode[_deviceId];
         if (!deviceName) {
           // Not found on database. At least guess main device type from string contents:
-          if (device.startsWith('iPod')) {
+          if (_deviceId.startsWith('iPod')) {
```
mikehardy: oh of course! wrong variable reference. That was the key part I was missing.

deviceinfo.js
```diff
@@ -544,18 +551,19 @@ export async function getModel() {
...
             deviceName = 'iPhone';
-          } else if (device.startsWith('AppleTV')) {
+          } else if (_deviceId.startsWith('AppleTV')) {
             deviceName = 'Apple TV';
           }
         }
       }
-      model = deviceName;
+      // Use unknown, otherwise getModel will never be cached.
+      model = deviceName || 'unknown';
```
mikehardy: Nice catch

mikehardy:
Great - first I reproduced on master locally on a Simulator by corrupting the lookup table entry for my Simulator (so that it would not be found), then I applied your PR and my previous lookup-table-corruption-reproduction, and everything worked as expected.

Thanks again

rbscott:
@mikehardy Nice! Glad I could help. Glad this was merged in so quickly.

Issue #732, opened 2019-07-23
Android: Support for simulated cutout option in developer tools


petekp:
# Feature Request

Android dev tools provide a setting that allows you to simulate a device cutout / notch on any device. Wondering if it might be possible for `DeviceInfo.hasNotch()` to recognize whether this setting is enabled and to return `true` if it is enabled.

<img src="https://user-images.githubusercontent.com/950779/61735610-2c54b200-ad39-11e9-9944-406b1ddda771.png" width="300" />

## Why it is needed

It would allow a developer to use a single simulator for testing UI changes both with and without a notch. Otherwise, two simulators are necessary.

mikehardy:
Oh that's cool - I didn't even know that existed. I'd love for that to be possible and would merge a PR for that immediately but I'm traveling right now so don't have time to look at it myself

mikehardy:
I think you might like https://github.com/th3rdwave/react-native-safe-area-context - there's a handy tutorial linked at the bottom of the README that shows it working on an emulator even

PR #794, opened 2019-09-17
Added 2019 iPhone model names for 'hasNotch' functionality

anthonyemery:
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    NA     |
| Windows |    NA     |

## Checklist

* [na?] I added the documentation in `README.md`
* [x ] I mentioned this change in `CHANGELOG.md`
* [na?] I updated the typings files (`deviceinfo.d.ts`, `deviceinfo.js.flow`)
* [ na?] I updated the dummy web/test polyfill (`default/index.js`)
* [ na?] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+## 3.1.3
+
+- chore: add support for 2019 iPhone models in 'hasNotch' mapping
+
 ## 3.1.2
 
 - fix: fix iOS crash in getModel for unknown modules (https://github.com/react-native-community/react-native-device-info/pull/791) thanks @rbscott!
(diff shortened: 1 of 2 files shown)


mikehardy:
Excellent, thanks! This looks fine and certainly follows current convention

Please note that as a future direction (read as: not this 3.x series) I want the ability to answer this question ("hasNotch?") to move to react-native-safe-area-context, which seems like a safer / more thorough way to do it (and also already supports the notch in emulators / simulators a la #732 here)

https://github.com/react-native-community/react-native-device-info/issues/594#issuecomment-527744091

mikehardy:
v3.1.3 should be out on npmjs.com now :rocket:

Issue #796, opened 2019-09-19
getReadableVersion causes runtime exception for react-native-web projects


tvanlaerhoven:
# Bug

When calling getReadableVersion() or getReadableVersionSync() from a react-native-web platform, an exception is thrown because undefined functions RNDeviceInfo.getAppVersion() and RNDeviceInfo.getBuildVersion() are called. 
Other DeviceInfo methods are correcly guarded by a platform check.

## Environment info

```bash
System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 34.84 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.14.0 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 22, 23, 24, 25, 26, 27, 28, 29
      Build Tools: 23.0.1, 23.0.2, 23.0.3, 24.0.2, 25.0.0, 25.0.1, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.1, 27.0.2, 27.0.3, 28.0.2, 28.0.3, 29.0.0
      System Images: android-19 | Google APIs Intel x86 Atom, android-22 | Android TV Intel x86 Atom, android-22 | Intel x86 Atom_64, android-22 | Google APIs Intel x86 Atom, android-24 | Android TV Intel x86 Atom, android-24 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom, android-25 | Google APIs Intel x86 Atom_64, android-26 | Google APIs Intel x86 Atom, android-27 | Android TV Intel x86 Atom, android-28 | Android TV Intel x86 Atom
      Android NDK: 15.2.4203891
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5692245
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native-tvos:  0.60.4-6 
  npmGlobalPackages:
    create-react-native-app: 2.0.2
    create-react-native-tizen-app: 0.0.1
[...]

PR #797, opened 2019-09-19
Check platform before querying readableVersion

tvanlaerhoven:
For web platforms, querying the readable version resulted in a crash. The extra platform check makes sure no undefined methods are called.

## Description

Fixed issue #796

Added platform check in getReadableVersion and getReadableVersionSync methods.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ✅     |
| Web | ✅     |

## Checklist

### DIFF
--- a/deviceinfo.js
+++ b/deviceinfo.js
@@ -839,11 +839,19 @@ export function getVersionSync() {
 }
 
 export async function getReadableVersion() {
-  return (await RNDeviceInfo.getAppVersion()) + '.' + (await RNDeviceInfo.getBuildNumber());
+  if (OS === 'android' || OS === 'ios' || OS === 'windows') {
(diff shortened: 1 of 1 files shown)


mikehardy:
It is slightly better and I *really* appreciate the willingness to PR but as I mentioned on #796 because react-native-web is more of a polyfill than anything (which in my mind means the methods themselves shouldn't really be called on the platform, but including the module shouldn't crash) I'm not that interested in return correctness and we already platform-check the underlying methods, so I just chained through to them

mikehardy:
If you are willing to actually flesh out web implementations though, that's a much different story and I'd happily merge real web implementations of things

PR #799, opened 2019-09-19
feat(🔥): Rewrite in TypeScript

Johan-dutoit:
This is more of a draft and open for discussion before we merge it in.


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 # Release Notes
 
+## 3.1.3
```
mikehardy: This looks to be just slightly outdated, could you rebase from master?

Clearly we need to get this reviewed and merged quickly before anything else even small comes in :-)

Johan-dutoit: Right yes, I wasn't expecting a 3.1.4 😉

package.json
```diff
@@ -2,25 +2,35 @@
...
+    "analyze": "npx tsc --noEmit",
+    "dev-sync": "cp -r deviceinfo* *podspec windows web android ios src example/node_modules/react-native-device-info/",
     "lint": "npx eslint ./ --ignore-pattern example --ignore-pattern node_modules --fix --quiet",
-    "pretty": "prettier --write deviceinfo.js",
     "shipit": "np",
-    "ts-check": "npx tsc deviceinfo.d.ts --noEmit",
     "test": "echo no tests defined",
-    "precommit": "lint-staged && yarn analyze"
+    "precommit": "lint-staged && yarn analyze",
+    "prepare": "bob build"
```
mikehardy: is "prepare" a community standard for this? I'm always looking for a "build" target personally, I'd think twice if I was coming into this as a fresh potential-contributor just trying a patch really quickly before working more deeply

Johan-dutoit: `prepare` gets triggered automatically when you create an npm package to be uploaded.  This should mean that `shipit` will also use it when it runs.

src/internal/nativeInterface.ts
```diff
@@ -0,0 +1,25 @@
+import { Platform, NativeModules, PlatformOSType } from 'react-native';
+import { DeviceInfoNativeModule } from './privateTypes';
+
+let RNDeviceInfo: DeviceInfoNativeModule | undefined = NativeModules.RNDeviceInfo;
+
+const OS: PlatformOSType = Platform.OS;
+
+// @ts-ignore
+if (OS === 'web' || OS === 'dom') {
```
mikehardy: not saying this is wrong, but after a year I just finally had to introduce my first ts-ignore in my work project. Is there nothing worth it (from a cost/benefit perspective) that can be done to type these for real?

package.json
mikehardy: Ah - so it is a standard (standard hook at least). That makes sense. This might be pedantic then, but could you add a build target and have prepare call it? I think it will just avoid doubt and if it chains to one underlying command should not increase maintainence

src/internal/types.ts
```diff
@@ -0,0 +1,14 @@
+export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'unknown';
+
+export type BatteryState = 'unknown' | 'unplugged' | 'charging' | 'full';
+
+export interface PowerState {
+  batteryLevel: number;
```
mikehardy: I could be wrong (I have not double-checked) but the entries in PowerState might be optional? Like they may be missing and thus need a '?'

src/internal/nativeInterface.ts
Johan-dutoit: I felt dirty adding this one.  `web` is typed part of `PlatformOSType`, which sadly doesn't include `dom`.  I also wasn't sure if `dom` was actually a valid value?  Perhaps something that was added as part of the cover everything `web` related.  

We could extend/union the current type and add `dom`, however I first wanted to understand where `dom` actually came from.

mikehardy: I believe I added dom so that it would work with jest as well, maybe there's a typing in there

src/internal/types.ts
Johan-dutoit: This was copied from master, would need to check the native implementation on iOS to confirm this.  I know Android always returns these values, having recently implemented it.

src/internal/nativeInterface.ts
Johan-dutoit: Oh great, I'll dig into that.

mikehardy: don't bother investigating I just checked and couldn't see it https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/jest I might be missing something transitive but for now I'd merge and we can fix later

src/internal/types.ts
mikehardy: Looks like lowPowerMode is not always populated, but I suppose we could alter that to return false (our default boolean). Implementors choice though I think I have a slight preference for returning a default value then making library consumers have to special case that value https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/RNDeviceInfo.m#L455

Johan-dutoit:
I've added the changes of 3.1.4 to the PR.

mikehardy:
interesting - github still thinks this is in conflict. not sure how to resolve that other than a force-push (very distasteful though I am pragmatic) or an actual rebase (then patching up locally). thoughts?

mikehardy:
I think just fix the conflict now and it's ready. I'd think major version just because typings - flow - moved but with a note that nothing should have changed for people, it's an abundance of caution more than anything else. But I could be swayed for it just to be a minor version, I just think people don't take type changes seriously enough personally

Johan-dutoit:
Assuming this is because `deviceinfo.js` changed in master and it was deleted in my branch.

Could merge master into my branch, resolve the conflicts (i.e. delete files) and then merge back into master.

As github suggests,

```
git checkout master
git merge --no-ff feat/typescript
git push origin master
```

mikehardy:
whatever works :-), it's not easy doing a change-all-files type change, so if that works great, we can merge then be back to normal. That's also why I'm "hovering" / "helicoptering" on this and being ridiculously interactive by the way, want to make sure it lands without any more fuss

mikehardy:
a draft for discussion originally perhaps, but it looked like a fair straight port, and I like it. I don't see anything to change now, and I'd merge it. Want to think on it for a little bit, or just merge - I'll leave that up to you. Also, you could squash or make a merge commit, also up to you - I'll let you push the button :-). Thanks!

src/internal/types.ts
mikehardy: maybe do one or the other as a separate PR so this can land now

Johan-dutoit:
Hah, I was planning on squashing it.

I'm still thinking about versioning.  I'm leaning towards a minor update.  As the importing of types will differ.  Perhaps that's a breaking change as you suggested.

mikehardy:
I see types as breaking changes unless they go less restrictive somehow. To me if I was using typescript, and I get a new version, and tsc doesn't compile anymore - breaking change. But I don't fear large major version numbers (and thus major version bumps). Who cares if it's react-native-device-info v32 ? Even though it sounds sort of silly I think it's a human perception issue when technically things should be marked as breaking more often

mikehardy:
Sweet! Needs an entry in the CHANGELOG though - you can just commit that directly to master and push it

mikehardy:
```
mike@isabela:~/work/react-random/react-native-device-info (master) % yarn dev-sync
yarn run v1.17.3
$ cp -r deviceinfo* *podspec windows web android ios src example/node_modules/react-native-device-info/
cp: cannot stat 'deviceinfo*': No such file or directory
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Johan-dutoit:
Hmm I removed deviceinfo from dev-sync!  Wonder if the merge of master brought it in again.

mikehardy:
this is where the easy build comes in - because the example depends on the git ref for the module in package.json and the build is easy, I can just cd in there and say 'yarn' and it all works :-), so I can still test

PR #800, opened 2019-09-20
Minor updates to the docs for getUniqueId()

RohanTalip:
Added punctuation, as the text wraps when rendered as is, and not having punctuation is confusing.  Added quoting.


mikehardy:
Excellent - I just added those during the v2->v3 changes so they were pretty raw, I appreciate the fixes

PR #801, opened 2019-09-20
refactor: change `Unknown` to `unknown` for consistency

Johan-dutoit:
For consistency, change the casing of `Unknown` to `unknown` like the rest of the library.

If we're happy with this, then we can add it to 4.0.x as a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # Release Notes
 
+## 4.0.1
+
+- BREAKING CHANGE: DeviceType: Renamed `Unknown` to `unknown` for consistency across the project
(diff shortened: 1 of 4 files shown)


mikehardy:
Yeah - I like this and I see you already thought of my only hesitation, it's a breaking change but if it goes with v4.0.0 and is noted carefully, works for me

Johan-dutoit:
Added a changelog.  Went with 4.0.1, as I fear 4.0.0 has been burnt.

mikehardy:
that's okay - it's as of this morning so I think it can still hop on the 4.x breaking change train. merge at will

Issue #802, opened 2019-09-20
Could not find com.google.android.gms:play-services-iid:16.1.0.


grit96:
In commit ea570d2bbe26b631681bfcc0572d71cc14a7c186, the gradle dependency `play-services-gcm` was swapped to `play-services-iid` but still uses `googlePlayServicesVersion` to determine the package version.

This causes a problem `play-services-gcm` and `play-services-iid` versions differ.

See extract below from https://dl.google.com/dl/android/maven2/index.html:

```
play-services-gcm
16.0.0
16.1.0
17.0.0

play-services-iid
16.0.0
16.0.1
17.0.0
```

`16.1.0` in `play-services-gcm`, but `16.0.1` in `play-services-iid`.

So by setting `googlePlayServicesVersion` to `16.1.0` for pre-AndroidX (as per the README), it actually breaks the Android build with `Could not find com.google.android.gms:play-services-iid:16.1.0.`

grit96:
As a workaround I am now setting `googlePlayServicesVersion` to `16.0.0`

mikehardy:
Very interesting - a more precise patch (which I'll commit later) will be to first check for googlePlayServicesIidVersion first (from the ext{} block in your android/build.gradle) and then fall back to googlePlayServicesVersion.

That will allow more precise control but not break installed base

Sorry it caused you to step back in versions

mikehardy:
@Johan-dutoit I'm going to do a 4.0.2 that has a fix for this once you've settled #801

mikehardy:
Okay the PR is ready - should merge soon. The chunk you want (if you want to use patch-package until it's merged+released is https://github.com/react-native-community/react-native-device-info/pull/804/files#diff-7ae5a9093507568eabbf35c3b0665732R46 )

beisert1:
We upgraded to react native 0.60.5 and updated this package from 0.21.5 to 3.1.4 and we are getting this "Could not find com.google.android.gms:play-services-iid:16.1.0" error.

Any update on when this fix will be officially released?

mikehardy:
Sorry - we're blocked on release for a moment while we ingest a conversion to typescript #144 and work through #776 - this one is a really trivial patch and I would use patch-package without delay so you are unblocked

mikehardy:
released as `@next` on NPM or 4.0.1-beta.2

Maushundb:
I'm getting this also on 5.3.0, is this expected?

[...]

PR #804, opened 2019-09-20
fix: android use googlePlayServicesIidVersion if available

mikehardy:
#802 showed that the initial attempt at a backwards-compatible dependency on the specific iid library prompted by #746 vs the full play services library had an error in the pre-AndroidX versioning (and will fail in the future for AndroidX)

This adds a new gradle variable googlePlayServicesIidVersion which targets the specific dependency version, and we will only fall back to the existing googlePlayServicesVersion if the IidVersion is not available. So this should fix the issue and still be compatible with the installed base.

While I was in the area I altered our build.gradle style to conform with current react-native-community best practices from https://github.com/react-native-community/discussions-and-proposals/issues/151#issuecomment-532787908 and I committed a new version of example/yarn.lock for current development purposes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,20 @@
 
 # Release Notes
 
-## 4.0.1
+## 4.0.1-beta.2
 
-- BREAKING CHANGE: DeviceType: Renamed `Unknown` to `unknown` for consistency across the project
-- JavaScript replaced with TypeScript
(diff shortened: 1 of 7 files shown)


mikehardy:
@Johan-dutoit I tested this in all the variations I could think of, I think it's good to go but a second pair of eyes is best :eyes:

Johan-dutoit:
LGTM 🔥

mikehardy:
released as `@next` on NPM or 4.0.1-beta.2

Issue #824, opened 2019-10-03
Support wifi IP address on Apple TV


gcesarmza:
# Feature Request

Currently this module returns "0.0.0.0" when the Apple TV is connected to a wifi network.

## Why it is needed

The reason the current implementation is not working is that the network interface names are different for iPhone (`en0`) and Apple TV (`en1`)

mikehardy:
This is published as 4.0.1-beta.3 or @next on npmjs.com - hope to have a stable v4.x release shortly

PR #825, opened 2019-10-03
Support wifi IP address on Apple TV

gcesarmza:
## Description

Fixed issue #824

## Checklist

* [ ] I added the documentation in `README.md` (not needed)

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -382,13 +382,21 @@ RCT_EXPORT_METHOD(getSupportedAbis:(RCTPromiseResolveBlock)resolve rejecter:(RCT
         temp_addr = interfaces;
         while(temp_addr != NULL) {
             if(temp_addr->ifa_addr->sa_family == AF_INET) {
-                // Check if interface is en0 which is the wifi connection on the iPhone
-                if([[NSString stringWithUTF8String:temp_addr->ifa_name] isEqualToString:@"en0"]) {
-                    // Get NSString from C String
-                    address = [NSString stringWithUTF8String:inet_ntoa(((struct sockaddr_in *)temp_addr->ifa_addr)->sin_addr)];
-
-                }
-
+                NSString* ifname = [NSString stringWithUTF8String:temp_addr->ifa_name];
+                #if TARGET_OS_TV
+                    if(
+                        // Check if interface is en0 which is the ethernet connection on the Apple TV
+                        [ifname isEqualToString:@"en0"] ||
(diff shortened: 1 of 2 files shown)


mikehardy:
I really appreciate the work - but I wonder if you have tried this? https://github.com/react-native-community/react-native-netinfo#type-is-wifi - we are really trying to move the network information stuff to the netinfo package completely, as it is purpose built to handle these things

gcesarmza:
Unfortunately that module doesn't even compile when trying to build for Apple TV with React Native 0.60

mikehardy:
Oh no! @matt-oakes are you aware tvOS isn't compiling netinfo right now? I know tvOS support in general in react-native-land is in flux and has to be supported on the side, I'm not sure if it's your intention to work with tvOS or not

You might be interested in this: https://github.com/react-native-community/react-native-device-info/pull/781 and this (for "how to test tvOS") https://github.com/react-native-community/react-native-tvos/issues/5#issuecomment-531254964

@gcesarmza I'll see review this more closely then and see what we can do

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -382,13 +382,22 @@ - (NSString *) getIpAddress {
...
-
-                }
-
+                NSString* ifname = [NSString stringWithUTF8String:temp_addr->ifa_name];
+                #if TARGET_OS_TV
+                    if(
+                        // Check if interface is en0 which is the ethernet connection on the Apple TV
+                        [ifname isEqualToString:@"en0"] ||
+                        // Check if interface is en1 which is the wifi connection on the Apple TV
+                        [ifname isEqualToString:@"en1"]
```
mikehardy: out of curiosity, why even make this '#if TARGET_OS_TV', is there any harm en just adding the en1 check to the main case and making it the only case?

gcesarmza: I don't know if it might affect in any way the logic for iPhone. That's why I put it in a separate block. I can update the PR though, let me know!

mikehardy: I have an iPhone and can test it, it seems like it would Just Work, and the simpler the better

mikehardy: by which I mean, please update the PR, I'll pull it and check :-)

gcesarmza: Thanks, the PR is updated now

mikehardy:
This is published as 4.0.1-beta.3 or `@next` on npmjs.com - hope to have a stable v4.x release shortly

Issue #843, opened 2019-10-14
hasNotch - wrong docs?


cristianoccazinsp:
# Bug

Sorry I'm not following the template, this is just a tiny issue with the docs. 

hasNotch does not appear in the v3 to v4 migration guide. However, shows as `boolean` return type on the docs, but uses a promise based example on the examples below. Which one is it? Promise or sync?

mikehardy:
Ah - hasNotch is actually just a pure-javascript lookup table, so it should be simply boolean - https://github.com/react-native-community/react-native-device-info/blob/master/src/index.ts#L759 - PRs always appreciated for docs, thanks for noticing this :pray:

mikehardy:
fixed :-)

mohak1990:
Still facing the same issue. Returning promise on DeviceInfo.hasNotch() check and not a Boolean value.

mikehardy:
I missed the type in the flow type definitions but @mohak1990 just fixed it in #881 :clap:

PR #845, opened 2019-10-15
Fix wrong example code of hasNotch() in docs

cereme:
## Description

Fixed issue #843 

Though return type of hasNotch in API table was right, example code was still Promise-based code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1260,9 +1260,8 @@ Tells if the device has a notch.
 #### Examples
 
 ```js
-DeviceInfo.hasNotch().then(hasNotch => {
+let hasNotch = DeviceInfo.hasNotch();
   // true
-});
 ```
 
 ---


mohak1990:
Still facing the same issue. Returning promise on DeviceInfo.hasNotch() check and not a Boolean value.

mikehardy:
@mohak1990 you are spamming me with notifications, why? Once is good.

https://github.com/react-native-community/react-native-device-info/blob/master/src/index.ts#L761

Looks good to me? 

https://github.com/react-native-community/react-native-device-info/blob/master/example/App.js#L88

works?

Suspect every part of your local project, caching, versions etc. `npx react-native-clean-project` if you have to

mohak1990:
Screenshot 2019-10-30 at 12 36 37 PM

mikehardy:
@mohak1990 Pictures aren't really helpful as proof of anything - prefer references to source like I did above, as that allows a collaborator to expand their reference context as much as needed to solve the problem.

I think I found it though - may I infer you are using flow? https://github.com/react-native-community/react-native-device-info/blob/master/src/index.js.flow#L84

Looks like that type was missed. Hit me with a PR and I'll merge it right in with credit for persisting to find this thing!

PR #850, opened 2019-10-17
Hooks

magrinj:
## Description

I've add 3 hooks, I made a first PR to know if everything seems good to you, because I plan to add hook for every function

Added: 
- [x] useBatteryLevel() hook
- [x] useBatteryLevelIsLow() hook
- [x] usePowerState() hook

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+## 4.0.3-rc.1
+
+- feat: Add `useBatteryLevel()`, `useBatteryLevelIsLow()`, `usePowerState` methods
+
 ## 4.0.2
 
 - getDeviceType converted to a constant
--- a/README.md
+++ b/README.md
@@ -1395,15 +1395,23 @@ DeviceInfo.getAvailableLocationProviders().then(providers => {
 });
 ```
 
-## Events
+## Hooks & Events
 
 Currently iOS & Android only.
 
-### RNDeviceInfo_batteryLevelDidChange
+### useBatteryLevel or RNDeviceInfo_batteryLevelDidChange
 
 Fired when the battery level changes; sent no more frequently than once per minute.
 
 #### Examples
+
+```js
+import { useBatteryLevel } from 'react-native-device-info'
+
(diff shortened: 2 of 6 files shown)


mikehardy:
Hmm - looks neat but I wonder about the possible code expansion - will this effectively double the size of the library? We're already starved for maintainers and that's a pile of new code.

Or is this just for events?

magrinj:
@mikehardy In the current case of this PR, I've implemented it just for the events because it's the more common use case for hooks (due to the possible update of the data, and the need of the component to re-render).

It can also be implemented on the other async functions, but you're right that's going to add some code for each function. So, it's lees needed on these functions, because the value is updated just one time. Hook can be created by library users if you want less to maintain.

As an example, each async hook for the other function should look like this:
```js
export function useSerialNumber(): string | null {
  const [serialNumber, setSerialNumber] = useState< string | null>(null);

  useEffect(() => {
    const setValue = async () => {
      const value: string = await getSerialNumber();
      setSerialNumber(value);
    };

    setValue();
  }, []);

  return serialNumber;
}
```

mikehardy:
Ok - I can imagine a good use case here for the events, and I'd be willing to merge that, but holy boilerplate batman, that's a lot of code for a marginal use case on the others, I'd like to avoid that for now. Maybe if we had a code generator or something ;-)

If I were to merge something that was for event only, is this PR already complete, or is there something more you'd add to it before review?

magrinj:
@mikehardy Ok seems good to me, I totally understand ! :) Yes a code generator should be enough for the other ones.

In this PR, I've just put the hook for the events, everything should be good if you want to review it and merge it !

mikehardy:
I'll do a little more local testing, but it looks good to me - thanks for this! I have read all about hooks but not used them yet, appears to be a much easier way to listen for events when consuming the library's API

mikehardy:
released!

magrinj:
Perfect, thanks a lot! :)

Issue #846, opened 2019-10-15
Crash: HostObject::get(propName:RNDeviceInfo)


renatoniro:
# Bug

When upgrading this project from v3.1.4 -> current 4.0.1, we have noticed an issue with the app crashing almost immediately on launch, citing a `HostObject::get(propName:RNDeviceInfo)` fatal error (that doesn't seem to give much more info than what's shown in this screenshot). 

<img width="1127" alt="Screen Shot 2019-10-11 at 4 14 32 PM" src="https://user-images.githubusercontent.com/10636415/66870541-e9346480-ef5e-11e9-990c-6fa00312b900.png">

Numerous other references to this same obscure error seem to mention the possibility of nil return values in the NSDictionary that is returned over the bridge via the `constantsToExport ` function. 

I have narrowed this particular instance of the error down to the call of `[self getAppName]` in  `RNDeviceInfo.m` [here](https://github.com/react-native-community/react-native-device-info/blob/643577cfadd92b9577b68ff2906af931ae81689b/ios/RNDeviceInfo/RNDeviceInfo.m#L67). Commenting out that line seems to resolve the crash. 

Digging deeper, it looks like the implementation of [that function](https://github.com/react-native-community/react-native-device-info/blob/643577cfadd92b9577b68ff2906af931ae81689b/ios/RNDeviceInfo/RNDeviceInfo.m#L142) retrieves `CFBundleDisplayName` which can be nil and get us into the state described above (indeed, this was exactly what was going on in our project).

According to [Apple Docs](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/TP40009249-109585), it looks like `CFBundleDisplayName` is now marked as 'required' and the older `CFBundleName` is 'recommended' despite there not being any enforcement of that rule (to date?) and despite conflicting info on [this page](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html) (conversely saying it is 'recommended' for ios).

For good measure I will implement `CFBundleDisplayName` in our app to get around this, but thought I would suggest these fixes to resolve the issue for anyone who may be in a similar scenario (upgrading an older project to newer RN + newer library versions): 
- Update the getAppName function to grab the value of `CFBundleName` as a fallback to `CFBundleDisplayName`, since that can be nil (as other repos do)
[...]

PR #855, opened 2019-10-21
fix(ios): app display name being null

rizwankce:
## Description

Fixed issue #846 

---
`getAppName` will return `CFBundleName` for those cases where `CFBundleDisplayName` is null


mikehardy:
This looks good - sorry I didn't get time to fix it myself but I've been busy with my work project. I will try to get this merged and released today, but either way you can use patch-package confidently to start using this change now - as the general approach is certainly approved (fall back to CFBundleName) and this looks like as clean an implementation as possible

rizwankce:
@mikehardy No Pressure! Thanks for your time. My project doesn't need this change. However couldn't resist raising PR when I saw the issue with detailed information and possible fix (which only takes 2-5mins)

mikehardy:
Got a chance to test this, works fine as expected, packing up changes for a v5 and this will be out in a moment

mikehardy:
released!

PR #858, opened 2019-10-22
Using latest iid library from firebase

andresesfm:
## Description
iid library: com.google.android.gms:play-services-iid has been deprecated as per: https://developers.google.com/android/reference/com/google/android/gms/iid/package-summary
It has been replaced with firebase
Fixed issue #<issue-number>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1261,7 +1261,7 @@ Tells if the device has a notch.
 
 ```js
 let hasNotch = DeviceInfo.hasNotch();
-// true
+  // true
 ```
 
 ---
@@ -1406,11 +1406,11 @@ Fired when the battery level changes; sent no more frequently than once per minu
 #### Examples
 
 ```js
-import { useBatteryLevel } from 'react-native-device-info';
+import { useBatteryLevel } from 'react-native-device-info'
 
 const batteryLevel = useBatteryLevel(); // 0.759999
 
-<Text>{batteryLevel}</Text>;
+<Text>{ batteryLevel }</Text>
 ```
 
 ```js
@@ -1436,11 +1436,11 @@ Fired when the battery drops is considered low
 #### Examples
 
 ```js
-import { useBatteryLevelIsLow } from 'react-native-device-info';
+import { useBatteryLevelIsLow } from 'react-native-device-info'
 
 const batteryLevelIsLow = useBatteryLevelIsLow(); // 0.19
 
-<Text>{batteryLevelIsLow}</Text>;
+<Text>{ batteryLevelIsLow }</Text>
 ```
 
 ```js
@@ -1461,11 +1461,11 @@ Fired when the battery state changes, for example when the device enters chargin
 #### Examples
 
 ```js
-import { usePowerState } from 'react-native-device-info';
+import { usePowerState } from 'react-native-device-info'
 
 const powerState = usePowerState(); // 'charging'
 
-<Text>{powerState}</Text>;
+<Text>{ powerState }</Text>
 ```
 
 ```js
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1406,11 +1406,11 @@ Fired when the battery level changes; sent no more frequently than once per minu
 #### Examples
 
 ```js
-import { useBatteryLevel } from 'react-native-device-info'
+import { useBatteryLevel } from 'react-native-device-info';
 
 const batteryLevel = useBatteryLevel(); // 0.759999
 
-<Text>{ batteryLevel }</Text>
+<Text>{batteryLevel}</Text>;
```
mikehardy: I like the semi-colon additions but you've altered whitespace at the same time, always a no-no for drive-by contributions ;-)

android/build.gradle
```diff
@@ -45,6 +45,6 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
-    implementation "com.google.android.gms:play-services-iid:$iidVersion"
+    def iidVersion = safeExtGet('iidVersion', safeExtGet('iidVersion', '19.0.1'))
```
mikehardy: That's a breaking change, not worth it in my opinion

android/build.gradle
```diff
@@ -45,6 +45,6 @@ repositories {
...
+    implementation "com.google.firebase:firebase-iid:$iidVersion"
```
mikehardy: did IID move completely to firebase? I'm using firebase heavily but hadn't noticed! Would explain the change above, but to make it non-breaking you have to chain the safeExtGet's so that it falls back to the existing one and doesn't break the installed userbase

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -724,11 +724,11 @@ public double getTotalMemorySync() {
   @SuppressWarnings({"ConstantConditions", "deprecation"})
   public String getInstanceIdSync() {
     try {
-      if (Class.forName("com.google.android.gms.iid.InstanceID") != null) {
-        return com.google.android.gms.iid.InstanceID.getInstance(getReactApplicationContext()).getId();
+      if (Class.forName("com.google.firebase.iid.FirebaseInstanceId") != null) {
+        return com.google.firebase.iid.FirebaseInstanceId.getInstance().getId();
```
mikehardy: I'll have to do some research quickly to make sure these are exactly equivalent - but in your research is that your conclusion? People rely on these ids being exact + precise for device id

mikehardy:
This looks really clean now and I verified that yes indeed - we're all moving IIDs to firebase now, thanks for this!

miblanchard:
@mikehardy Any idea when a new version with these changes will make it to npm?

mikehardy:
I just did a huge batch today and I'm out of time, it might be a day or three sorry - there is always patch-package in the meantime - or I suppose since it's merged and committed you can depend on github using the commit hash then as a postinstall step run the build inside the package in node_modules to go from typescript -> javascript and have it be ready for use

miblanchard:
That timing works for us. Thank for all your hard work in the community!

Issue #643, opened 2019-04-18
Headphones status


chenjianfang:
How to judge whether the headphones are plugged in, and the headphones are playing sound？？？

mikehardy:
I don't know, I haven't even looked for those APIs on the platforms and the library doesn't support them. Have you searched on stackoverflow, or in the Android or iOS developer documentation to see if it is even possible?

machour:
Seems to be doable for Android, but a bit trickier for iOS. 

https://forums.xamarin.com/discussion/105066/detecting-if-headphones-are-plugged-in-on-android-and-ios (these are xamarin docs but they do mention os based solutions)

mikehardy:
okay - looks possible but it would need someone to implement, marking it as such

RWOverdijk:
https://stackoverflow.com/questions/21292586/are-headphones-plugged-in-ios7

Not sure if this perhaps changed over time, but it looks usable. I'll be implementing this for my project (so not here) but I thought I'd share the link anyway.

Taym95:
I will take this issue over, PR soon!

mikehardy:
Closed in #860

PR #860, opened 2019-10-23
Added headphones support

chakrihacker:
## Description

Fixed issue #643 

Added `isHeadphonesConnected()` that allows checking wired headset or Bluetooth headphones connected status

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1141,6 +1141,14 @@ export function isLocationEnabledSync() {
   return false;
 }
 
+export function isHeadphonesConnected() {
+  return RNDeviceInfo.isHeadphonesConnected();
+}
+
+export function isHeadphonesConnectedSync() {
+  return RNDeviceInfo.isHeadphonesConnectedSync();
+}
+
 export async function getAvailableLocationProviders() {
   if (OS === 'android' || OS === 'ios') {
     return RNDeviceInfo.getAvailableLocationProviders();
@@ -1343,6 +1351,8 @@ export default {
   getSystemAvailableFeaturesSync,
   isLocationEnabled,
   isLocationEnabledSync,
+  isHeadphonesConnected,
+  isHeadphonesConnectedSync,
   getAvailableLocationProviders,
   getAvailableLocationProvidersSync,
   useBatteryLevel,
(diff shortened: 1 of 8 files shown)


mikehardy:
Oh this is very cool! Thank you!
Let me get the current release out (working on that this afternoon) then I'll review this and test it and hopefully get it out as a minor version after if it looks good

src/index.ts
```diff
@@ -1141,6 +1141,14 @@ export function isLocationEnabledSync() {
   return false;
 }
 
+export function isHeadphonesConnected() {
+  return RNDeviceInfo.isHeadphonesConnected();
```
mikehardy: These need to be restricted to android and ios
```suggestion
  if (OS === 'android' || OS === 'ios') {
    return RNDeviceInfo.isHeadphonesConnected();
  }
  return false;
```

src/index.ts
```diff
@@ -1141,6 +1141,14 @@ export function isLocationEnabledSync() {
...
+}
+
+export function isHeadphonesConnectedSync() {
+  return RNDeviceInfo.isHeadphonesConnectedSync();
```
mikehardy:
```suggestion
  if (OS === 'android' || OS === 'ios') {
    return RNDeviceInfo.isHeadphonesConnectedSync();
  }
  return false;
```

mikehardy:
oh one other thing - if you could add it to example/App.js and README (the method table, and the example section) - that helps show usage

chakrihacker:
i will finish it today

chakrihacker:
Where should I add polyfill for web??

mikehardy:
Oh good question - I restructured the code so the web is no longer a polyfill, it stays up to date automatically. I need to update the PR template to remove that bit - ignore web polyfill :-)

chakrihacker:
I have tested Bluetooth and normal headphones for android, and for ios Bluetooth is tested. Need someone to test wired headset of type-c

mikehardy:
Cool - do you think this is ready to go from your perspective then? I won't have all the headphones types to check either but for a new feature if there is a problem someone else can try other types and fix it - it's impossible to break current users because there aren't any and you can only test what you have access to...

chakrihacker:
It's ready to go then

mikehardy:
Sorry missed this in my last sweep but it does look good to go, thank! Will release a new minor version in a little bit with this included

PR #863, opened 2019-10-25
Android build fails due to extra parenthesis

andresesfm:
Removing extra parenthesis that makes Android Studio fail to load project

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -45,7 +45,7 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    def firebaseIidVersion = safeExtGet('firebaseIidVersion', null))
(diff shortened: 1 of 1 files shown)


mikehardy:
@acoates-ms are you aware of something changing in github actions windows build-land? Your github action was magic at the start but now we appear to be having a problem with the syntax of commands that worked, with an error message that only leads to powershell stackoverflows, and we're not specifying any shell?

https://github.com/react-native-community/react-native-device-info/pull/863/checks?check_run_id=274240992

So how could it work before but then switch to powershell now etc? Is there a way to force a path or be explicit about which shell is in use?

acoates-ms:
@mikehardy  - Yeah GitHub actions are still in Beta, which is code for still making lots of breaking changes.  See #864 for a fix.

mikehardy:
This only had failing CI for windows for unrelated reasons, those are fixed now and I'm merging this in - thank you!

PR #866, opened 2019-10-25
Refactor device type resolver

andresesfm:
## Description

Refactors the methods used to determine de device type on Android. This allows Android native interoperability (described in the README.md

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -29,5 +29,4 @@ Added `yourNewMethodName()` that allows ...
 * [ ] I added the documentation in `README.md`
 * [ ] I mentioned this change in `CHANGELOG.md`
 * [ ] I updated the typings files (`privateTypes.ts`, `types.ts`)
-* [ ] I updated the dummy web/test polyfill (`default/index.js`)
 * [ ] I added a sample use of the API (`example/App.js`)
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -26,19 +26,24 @@ jobs:
         run: npm install -g react-native-cli
 
       - name: Init new project
-        run: cd ${{ runner.temp }} && react-native init testcli --version 0.59.10
+        run: react-native init testcli --version 0.59.10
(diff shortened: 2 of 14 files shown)


mikehardy:
Interesting - I'd never thought of that as a use case but I haven't had to handle react-native brown-field apps as of yet, I assume that's the use for you?

Looks sane enough and even without the use case it separates this out as a nice coherent chunk of logic.

How have you tested it?

andresesfm:
Hello, thank you for looking at my PR @mikehardy. We have a "pure" react-native app. Here is the scenario we need this for: We want call `requestOrientation` as the first thing when we create the main activity. This is so the tablet version of the app starts in landscape. 

We tried three things: 1) Doing it in JS but since RN takes a second to launch, we would see portrait execution for a split second. So we went to 2) Doing the check on native, this works well but we logic implemented was different from the one in react-native-device-info.  Finally, 3) We went with your implementation for consistency and it works well but that meant I had to copy-paste the same chunk of code. 

So, this proposed solution would solve all our problems, it means that we can reuse the logic implemented in the library on both native and js without code duplication. As an added bonus, decouples that functionality that does not require `reactContext`, any Context object would do

mikehardy:
That's an interesting problem + solution, I know from work with the splash screen module that yes - there's sometimes quite a long time in native code before JS loads things. Yeah, I think it passes muster just as a nice encapsulation of the logic.

So you tested this on phone and tablet and it looks good to you? Seems like the answer is yes but I want to confirm

andresesfm:
Yes, we have tested on both phones and tablets of different sizes/densities.

mikehardy:
> Yes, we have tested on both phones and tablets of different sizes/densities.

Excellent! merged - I have a couple unreleased changes on master now and hope to get a release out shortly but obviously you're safe in using this now - either depending on the git commit hash (safe + reproducible) or master as you see fit.

Thanks!

PR #867, opened 2019-10-26
Check classes exist

andresesfm:
## Description

Fixed issues caused when not finding either firebase or GMS Id providers. Extracted to a single-purpose class and added Unit tests.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
@@ -1,6 +1,7 @@
 package com.learnium.RNDeviceInfo.resolver;
 
 import android.content.Context;
+
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public DeviceIdResolver(Context context){
+        this.context = context;
+    }
+
+    public String getInstanceIdSync() {
+        try {
+            return getFirebaseInstanceId();
+        } catch (ClassNotFoundException e) {
+            System.err.println("N/A: Add com.google.firebase:firebase-iid to your project.");
```
mikehardy: The only thing I don't like about this is that structurally library consumers will definitely get a System.err.println() with a dire message, but then just after getGmsInstanceId() might succeed meaning things are deprecated but not technically in error.

Is there a way to more gently handle the failure in getInstanceIdSync - ideally with the catches setting flag so you can then offer different messaging if both are missing vs one, but recognizing that is a larger restructure, even a slight textual change acknowledging GMS still exists but firebase is the future vs text that sounds like a hard-fail would work for me

mikehardy:
Nice - thanks again - just the note I left about softening the first soft-fail case but otherwise I like the refactor and compartmentalization

android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
andresesfm: Hello, I modified so it only writes a message if both classes are not found. Thank you

mikehardy:
Looks good thank you!

PR #868, opened 2019-10-27
Adding useFirstInstallTime()

schie:
## Description

Added `useFirstInstallTime()` that simply wraps async function, `getFirstInstallTime()`, into a hook.

_**Note**: The src can be used as a template for wrapping the rest of the async functions into hooks where applicable._

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1487,6 +1487,22 @@ deviceInfoEmitter.addListener('RNDeviceInfo_powerStateDidChange', { batteryState
 });
 ```
 
+---
+
+### useFirstInstallTime
+
+Gets the time at which the app was first installed, in milliseconds.
+
+#### Example
+
+```jsx
+import { usePowerState } from 'react-native-device-info';
+
+const { loading, firstInstallTime } = useFirstInstallTime(); // { loading: true, firstInstallTime: 1517681764528}
+
+<Text>{loading ? 'loading' : firstInstallTime}</Text>;
+```
+
 ## Troubleshooting
 
 When installing or using `react-native-device-info`, you may encounter the following problems:
--- a/README.md
+++ b/README.md
@@ -1498,9 +1498,9 @@ Gets the time at which the app was first installed, in milliseconds.
 ```jsx
 import { usePowerState } from 'react-native-device-info';
 
-const { loading, firstInstallTime } = useFirstInstallTime(); // { loading: true, firstInstallTime: 1517681764528}
+const { loading, result } = useFirstInstallTime(); // { loading: true, result: 1517681764528}
 
-<Text>{loading ? 'loading' : firstInstallTime}</Text>;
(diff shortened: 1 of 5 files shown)


mikehardy:
Hi there! Thanks for proposing this

On the face of it, it seems more complicated than I want, with a new type etc., but I can see how that's attempting to mask the async-ness of the underlying method

Is there anyway to do it such that it doesn't need a new type wrapper (with loading flag) and is just working with a number?

I'm not wise on how hooks work yet so I don't know if this is the normal pattern for handling async in hooks or not, but I'd be happy to read up on it if you can offer a link or two to docs or other examples

schie:
Hi @mikehardy,

Thanks for getting back so quickly.

I don't really have much documentation on why to use the flag itself, but there are a number of other libraries that I've used that use it. The first that comes to mind, is [Apollo's `useQuery` hook](https://www.apollographql.com/docs/react/api/react-hooks/#usequery).

I like the idea of a `loading` field because it lets me know why the value may be `null` and how to handle it (show a spinner, loading modal, etc...) 

With that being said, I can definitely change it to just a simple `null | number`. 

Also, the `loading` flag may not be necessary since it is very quick to fetch that information.

Thoughts?

mikehardy:
yes - there are a range of options, items like that could even be constants with no async to hide. This seems like the same sort of problem I was intending to solve the mirrored zzz() zzzSync() API mirroring, but expressed in hook semantics. Not necessarily bad, I'll follow the link you sent and study a bit

mikehardy:
Ok - I studied up, thanks for the link. I think the pattern is reasonable so I'm okay with adding it like this but I'd like make it generic so that every hook wrapping an async call doesn't need it's own type. Seems like we could have one AsyncHookResult type that has loading then a result that's one of the other types we can return? Does that make sense?

Seperately - I don't want to return nulls, I'd like to stick with the default return values for unknown items which would be -1 at the moment https://github.com/react-native-community/react-native-device-info/blob/master/src/index.ts#L780

schie:
Good idea. 

We could have an Generic type something  like the following:

```ts
interface AsyncHookResult<T>{
  loading: boolean;
  result: T|-1
}
```

So the hook would return `AsyncHookResult<number>`

Thoughts?

If all is kosher, I’ll update with the above.

mikehardy:
I think result is T|-1|false|'unknown' as those are our three default no-answer answers, but otherwise, yes!

schie:
Hi @mikehardy,

Happy Monday!

Should be good-to-go for you.

mikehardy:
That looks pretty sweet and should be a nice template for others to follow if they like. Thanks!

PR #869, opened 2019-10-29
Simplifying hookification and adding use device name

schie:
## Description

- Added `useDeviceName()` that wraps `getDeviceName()`
- Added internal utility function, `useOnMount(...)`,  to simplify the _'hookification'_ (patent pending) of some async functions

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1516,6 +1516,22 @@ const { loading, result } = useFirstInstallTime(); // { loading: true, result: 1
 <Text>{loading ? 'loading...' : result}</Text>;
 ```
 
+---
+
+### useDeviceName
+
+Gets the device name.
+
+#### Example
+
+```jsx
+import { useDeviceName } from 'react-native-device-info';
+
+const { loading, result } = useDeviceName(); // { loading: true, result: "Becca's iPhone 6"}
+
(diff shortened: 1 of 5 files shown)


mikehardy:
Nice - this softens the blow of the code multiplier that hooks otherwise is - so now it looks like (whitespace aside) effectively a one-liner to wrap any of the async calls in a hook? And this works for you on firstInstallTime and deviceName in testing?

schie:
Yep! It should be super simple to wrap most of the other async functions. 

There are a few that will require a bit more than the wrapper I made. I’ll work on those soon for another PR.

mikehardy:
Great - merged - it sounds like you have an internal plan for more development, I could release now or wait - I don't believe anyone is waiting for anything and there are no bugfixes queued so I'm inclined to wait unless you need a release for your own purposes with this work included (which would be fine too). 

Cheers

schie:
Waiting is fine with me! I'll probably have another PR out tonight

PR #872, opened 2019-10-30
Fix npmignore and files field

friederbluemle:
Right now the published tarball of the repo potentially contains a ton of files that should not be published.

This changes the `files` field in `package.json` to include specific directories, now combined with targeted `.npmignore` files for each platform.

On my development machine the numbers are down around 50% (mainly due to .idea and .gradle folders).

### Before

```
=== Tarball Details ===
name:          react-native-device-info
version:       5.2.1
filename:      react-native-device-info-5.2.1.tgz
package size:  188.5 kB
unpacked size: 828.7 kB
total files:   143
```

### After

```
=== Tarball Details ===
name:          react-native-device-info
version:       5.2.1
filename:      react-native-device-info-5.2.1.tgz
package size:  98.5 kB
unpacked size: 511.5 kB
total files:   65
```


mikehardy:
Thank you! :bowing_man: you should have seen the size of the releases on npm temporarily before I realized what had happened and published smaller point releases. 200MB one time! That motivated me to look for a solution but I didn't have a good one yet. This is great

My only question, why not e.g. '/lib/' vs 'lib/' - i.e., rooted instead of rootless? Does it matter? Seems like rooting it would be the most precise expression but I may be missing why it doesn't even matter

friederbluemle:
Yeah, very good observation. :) Just a couple of days ago I was wondering about the same exact thing (leading slash required or not). What I found was that with or without the leading slash, _only_ root directories (basically on the same level as the package.json) are considered.

friederbluemle:
Also, I have a quick question: It's correct to publish _both_ `lib/` _and_ `src/`, correct? (I saw `lib/` is generated, and I'm not familiar with `bob`)

mikehardy:
That leading slash thing is...perplexing. So it's like a regex, except it's not? Okay then ;-)

Technically you don't have to publish src but I have a hard rule for my repos which is to handle the "patch-package" use case well. That means specifically that anyone should be able to go right into the node_modules folder on my module, tweak something (in the actual implementation file) as a test for a fix, run a one-line build called "build", and have it work. Then  you run patch-package and you're good plus you're confident proposing a PR. I learned that from reactxp vs a few other projects. reactxp got it right (source included! one line build!) and gained me as a contributor as my drive-by dev experience is so easy. The other projects were less fun to patch up

friederbluemle:
Makes sense, thanks for explaining! 👍

mikehardy:
Looks fine on review of course, and I usually do RNDI releases over the weekend these days, so I'll exercise it completely shortly. Thanks again!

Issue #873, opened 2019-10-30
preferredContentSizeCategory is called from non-main thread


paddlefish:
# Bug
When using getFontScaleSync or getFontScale the iOS API application.preferredContentSizeCategory is called from non UI thread.  This prints an annoying log to the console and causes launch to pause for a few seconds.

## Environment info

iOS

Library version: HEAD

## Steps To Reproduce

call getFontScale on a hardware device (not simulator)
observe console
```
2019-10-30 16:59:40.013167-0500 MYAPPSNAME[502:81320] [reports] Main Thread Checker: UI API called on a background thread: -[UIApplication preferredContentSizeCategory]
PID: 502, TID: 81320, Thread name: (none), Queue name: com.facebook.react.RNDeviceInfoQueue, QoS: 0
Backtrace:
4   MYAPPSNAME                          0x000000010245d58c -[RNDeviceInfo getFontScale] + 84
5   MYAPPSNAME                          0x000000010245d998 -[RNDeviceInfo getFontScale:rejecter:] + 132
6   CoreFoundation                      0x000000019bda6d90 DA838E75-6B30-360E-9661-C4800A7E1BF6 + 1252752
7   CoreFoundation                      0x000000019bc76bd0 DA838E75-6B30-360E-9661-C4800A7E1BF6 + 7120
8   CoreFoundation                      0x000000019bc777a8 DA838E75-6B30-360E-9661-C4800A7E1BF6 + 10152
9   MYAPPSNAME                          0x000000010257eef0 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 1896
10  MYAPPSNAME                          0x00000001025828ec _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicE + 660
11  MYAPPSNAME                          0x0000000102582460 _ZZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEiENK3$_0clEv + 144
12  MYAPPSNAME                          0x00000001025823c4 ___ZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEi_block_invoke + 28
13  libdispatch.dylib                   0x0000000103fbd828 _dispatch_call_block_and_release + 24
14  libdispatch.dylib                   0x0000000103fbec04 _dispatch_client_callout + 16
15  libdispatch.dylib                   0x0000000103fc5b74 _dispatch_lane_serial_drain + 744
[...]

PR #874, opened 2019-10-30
Ensure preferredContentSizeCategory is only called from the main thread

paddlefish:
## Description

Fixed issue #873

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    n/a     |
| Windows |    n/a     |

## Checklist

* [n/a] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [n/a] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -382,7 +382,10 @@ RCT_EXPORT_METHOD(isEmulator:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromis
     
     // Shared application is unavailable in an app extension.
     if (application) {
-        NSString *contentSize = application.preferredContentSizeCategory;
+        __block NSString *contentSize = nil;
+        RCTUnsafeExecuteOnMainQueueSync(^{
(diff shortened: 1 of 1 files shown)


mikehardy:
Wow, RCTUnsafeExecuteOnMainQueueSync is a really ugly sounding method name so I did some research, and that is some subtle deadlock avoidance right there - seems like the perfect fix with the context that you must check if you are on the main queue first and this method wraps that up. Thank you very much!

mikehardy:
I don't have fancy auto-semantic-release-bot stuff working on this repo so it's a manual release - I'll try to get it out in the next few days but obviously it's merged so you can rely on the git hash or patch-package in a pinch in the meantime, thanks again

paddlefish:
I didn’t name that function... but after writing almost the same code I noted that other places used some macros to do some thread shuffling and found that Gem. Yes the name is terrifying but it seems to do exactly what’s needed. There is a “once” version that does the thread hop only once and caches the result.... but the font scale can change during the lifetime of an app so I chose to err on the side of caution and not use that.

PR #875, opened 2019-10-31
Adding useLandscape()

schie:
### Added `useLandscape()`

This hook listens to `Dimensions`' change events, then updates state on change.
This allows for effects to be made when the device enters or leaves landscape mode. (show drawer menu button in header when not in landscape, show / open drawer menu when in landscape, etc...)

I decided to not have the return type as `AsyncHookResult<boolean>` since there won't be a point where the the result isn't known, so a `loading` prop would be unnecessary.
 
With that being said, if we want to have `AsyncHookResult<boolean>` as the return type, I can change it to be so.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1502,6 +1502,24 @@ deviceInfoEmitter.addListener('RNDeviceInfo_powerStateDidChange', { batteryState
 
 ---
 
+### useLandscape
+
+Tells if the device is currently in landscape mode.
+
+_This hook listens to dimension change events, determines if the window is in landscape, then updates the result accordingly._
+
+#### Example
+
+```jsx
+import { useLandscape } from 'react-native-device-info';
+
+const isLandscape = useLandscape(); // true or false
+
+<Text>{isLandscape}</Text>;
+```
+
+---
+
 ### useFirstInstallTime
 
 Gets the time at which the app was first installed, in milliseconds.
--- a/example/App.js
+++ b/example/App.js
@@ -16,6 +16,7 @@ import {
   useBatteryLevel,
   useBatteryLevelIsLow,
   usePowerState,
+  useLandscape,
   useFirstInstallTime,
   useDeviceName,
 } from 'react-native-device-info';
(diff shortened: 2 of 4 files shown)


mikehardy:
I'm not against this in general - I'm leaning towards merge, but my long-term plan is to kill all the display-oriented stuff because to handle it well needs a more integrated solution, and it exists in react-native-safe-area-context / react-native-safe-area-view. I would really like to steer future work there so those libraries did what people needed with regards to "hasNotch? what insets? landscape?" etc.

Without being obstructionist, do you think you could open the conversation with react-native-safe-area-context to see how they'd feel about ingesting this functionality there?

schie:
Hi @mikehardy,

I’ll definitely investigate and ask the folks at `react-native-safe-area-context`, then get back to you.

schie:
hi @mikehardy,

it looks like [**react-native-community/react-native-hooks** already has a very similar hook](https://github.com/react-native-community/react-native-hooks#usedimensions), were they are simply just spitting out the updated dimensions on change.

While not exactly the same, the same goal can be accomplished with that. 

So, if the plan is to gut all of the display/dimension, we _could_ just close this.

Thoughts?

mikehardy:
I'd go with that one (react-native-hooks) vs digging deeper on display-oriented things.

I just discovered https://github.com/marudy/react-native-responsive-screen and https://github.com/wonday/react-native-orientation-locker today as well - between safe-area-context and safe-area-view and those it's just an area I want to play with lest it fractures even more :-), it needs unifying more than anything

But that's only big chunk of functionality I'm hoping to evict in the future, please don't let me deter you from any other things you have in mind, if any

schie:
@mikehard, 

No worries.

Are there any other things that you're looking to purge in the future? If not, I'll continue with the hookification.

Also, Do you think it makes sense to do a little refactoring? move hooks into it's own file and such? It might be a little easier to look at while developing and add less changes to the non-hook file. 🤷‍♂

mikehardy:
Since they are mostly wrapping existing API, having them separate might be nice? Can't hurt to try and will probably look better. "implementor's choice"

Other stuff - no plans at the moment

PR #880, opened 2019-11-05
Add more web functionality

### LINKED ISSUE #878: Cannot resolve "../../web"
# Bug

When using `react-native-web` I get this error:
```
./node_modules/react-native-device-info/lib/module/internal/nativeInterface.js
Module not found: Can't resolve '../../web' in '/<redacted>/app/node_modules/react-native-device-info/lib/module/internal'
```

## Environment info (web)
Tried with library versions: 5.0.0, 5.1.0, 5.2.0

## Steps To Reproduce
1. `react-scripts start`
2. Compiler error

Describe what you expected to happen:
1. Compilation works fine

## Fix
Change import to `../../../web`. Not sure of the effect of this as haven't explored the codebase.

robertherber:
## Description

Adds web support for `isCameraPresent`, `isLocationEnabled`, `isAirplaneMode` as well as all battery/charging-related APIs.

And it fixes #878, so that web builds.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web |    ✅     |

## Checklist


mikehardy:
@robertherber This is fantastic! I had no idea there was that much API available on the web.

 I've read it over once and I can't see anything wrong with the code. I especially appreciate the docs update. This is headed for a merge and soon

The only thing that worries me is that web was effectively completely broken, and silently. Is there any small webpack (or similar) config you've got laying around which could work like the current checks to make sure that web can make a bundle? Then I could hook that to a github action or at least pre-publish to make sure I never silently break web again. And we could have a little example webapp to show the stuff working

robertherber:
I just added a couple of more APIs as well as moved the web native implementation into the src folder (before the path had to be different for src and lib, I think this is a better solution).

I think the best would be to add a few simple tests. However - since I'm not very familiar with TypeScript and this library I think it's probably better if one of you core maintainers set up the basic test structure. I'd be happy to help writing a few tests.

mikehardy:
Hoping for a release of this shortly - I was really blocked on my work project until just today so hadn't had time. But my work project will also use the web support, so I'll be testing this directly. Thanks again!

PR #902, opened 2019-11-23
Improve types + fix errors + improve example

zoontek:
## Description

Hi 👋. First, sorry, this PR contains multiples fixes / features!

It starts by updating the `AsyncHookResult` type to help my friends who currently work on ReasonML binding: https://github.com/reason-react-native/device-info and allowing them to stay zero cost.

Then I split the methods directly exposed from native and the "JS" ones to type the exposed module (I sorted keys by alphabetical order for diffing, I was to afraid to miss a method in the changes). It revealed some errors:

- `getUserAgentSync` is missing on Android (I fixed that) and iOS (can't fix it because of the completion handler, so I added a comment)
- `getFontScale / getFontScaleSync` could return `number | 'unknown'` instead of `number | -1` according to the documentation
- `getPreviewSdkInt / getPreviewSdkIntSync` could type `number | 'unknown'` instead of `number | -1` according to the documentation
- `isHeadphonesConnected` return type was `boolean | Promise<boolean>` because the function was not async

To test that, I started the example app which used the package from the GitHub repository `master` branch, not the local one I just modified 😅, so I updated the example `package.json` to use that one.

I also updated the example app style with tabs to improve readability.

Screenshot 2019-11-23 at 14 43 28

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS **(except windows)**


src/index.ts
```diff
@@ -1281,127 +1294,130 @@ export function useIsEmulator(): AsyncHookResult<boolean> {
   return useOnMount(isEmulator, false);
 }
 
-export default {
-  getUniqueId,
-  getInstanceId,
-  getInstanceIdSync,
-  getSerialNumber,
-  getSerialNumberSync,
+const deviceInfoModule: DeviceInfoModule = {
```
mikehardy: DeviceInfoModule was pulled in from an internal type. Now it's being exported, is that exposing an internal-labeled item publicly?

example/package.json
```diff
@@ -6,12 +6,17 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "test": "jest"
```
mikehardy: None of this will persist. I have no intention of maintaining a static example and taking on the pointless work of keeping it up to date with react-native versions, it is clobbered every time by the refresh-example.sh script

example/App.js
```diff
@@ -8,7 +8,14 @@
  */
 
 import React, {Component} from 'react';
-import {ScrollView, StyleSheet, Text, SafeAreaView} from 'react-native';
```
mikehardy: This App.js file is the only one that is copied to safety during the example refresh run IIRC, so this will actually persist. But to the extent it deviates from the stock App.js from the template, deviations should be minimized - nothing with regard to style or whitespace or anything has value - only demonstrating functionality

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -705,18 +705,20 @@ public String getSecurityPatchSync() {
   @ReactMethod
   public void getSecurityPatch(Promise p) { p.resolve(getSecurityPatchSync()); }
 
-  @ReactMethod
-  public void getUserAgent(Promise p) {
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  public String getUserAgentSync() {
```
mikehardy: This will be a very slow synchronous method :-), but I see no harm in having it of course - users have their own needs

mikehardy:
I do like the tabs change for the example - that's a nice touch now that I see it. Will help me as I'm visually diffing during development

src/index.ts
```diff
@@ -1,9 +1,10 @@
-import { useState, useEffect, useCallback } from 'react';
-import { Platform, Dimensions, NativeEventEmitter, NativeModules } from 'react-native';
```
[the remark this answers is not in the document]
zoontek: For imports, I just ran VS Code `Organize imports` command

Screenshot 2019-11-23 at 15 14 55

MoOx:
For my concern, nothing is blocking me with current master in practise, it was only a theory issue.

I just had to understand the real types & my bindings are actually reflecting the actual code (as said in my issue, types where wrong somehow).

That said, the changes in this PR match what I did in the reasonml bindings, so it looks good to me :)

src/index.ts
zoontek: `DeviceInfoModule` is used in the `index.ts` file, but not exposed to the user (totally absent in the generated `index.d.ts`)

src/index.ts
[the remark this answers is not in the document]
mikehardy: How weird that it got one "right" (alphabetical) but messed up the other? fine by me other way. I'm certainly not going to go in there and changes imports manually ;-)

mikehardy:
So at this point please just re-run refresh-example script to stay within it's bounds (while preserving the App.js changes) and I think this is good to go?

zoontek:
@mikehardy Done! 🙂

mikehardy:
5.4.0 is out, thanks!
https://github.com/react-native-community/react-native-device-info/releases/tag/v5.4.0

PR #903, opened 2019-11-25
Added support to get Android’s current "Locating method"

oferRounds:
## Description

Added `getAndroidLocatingMethod()` that allows to get device current "Locating method" (HighAccuracy / BatterySaving / SensorsOnly / Off)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     | relevant only to Android
| Android |    ✅     |
| Windows |   ❌     | 

## Checklist

* [v] I have tested this on a device/simulator for each compatible OS
* [v ] I added the documentation in `README.md`
* [ ] I mentioned this change in `CHANGELOG.md` // was not sure if I the one who should do it
* [v] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [v] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -63,6 +63,24 @@ import static android.os.BatteryManager.BATTERY_STATUS_CHARGING;
 import static android.os.BatteryManager.BATTERY_STATUS_FULL;
 import static android.provider.Settings.Secure.getString;
 
+enum LocatingMethod
+{
+  Off("OFF"),
+  SensorsOnly("SENSORS_ONLY"),
+  BatterySaving("LOCATION_MODE_BATTERY_SAVING"),
+  HighAccuracy("HIGH_ACCURACY");
+
+  private String locatingMethod;
+
+  LocatingMethod(String locatingMethod) {
+    this.locatingMethod = locatingMethod;
+  }
+
+  public String toString() {
+    return this.locatingMethod;
+  }
+}
+
 @ReactModule(name = RNDeviceModule.NAME)
 public class RNDeviceModule extends ReactContextBaseJavaModule {
   public static final String NAME = "RNDeviceInfo";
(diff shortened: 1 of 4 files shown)


mikehardy:
Hi there! Thank you for this - I think getAvailableLocationProviders is very similar to this functionality - I just checked it on my phone and it seems like it conveys the same information at any rate. Can you tell me what this solves that is not possible with getAvailableLocationProviders?

oferRounds:
@mikehardy thanks! Actually you are right – it does seem equivalent, I some how missed it – sorry! 
I will go a head and close the PR

Thank you!

mikehardy:
No problem - I will never be bothered by someone actually hacking on the code and proposing something, hopefully if you see anything that is truly missing this doesn't put you off, I do happily accept PRs all the time :-). Cheers

oferRounds:
Thank you!
Actually just question I noticed now: what does the 'passive` boolean stand for on the Android’s result?

mikehardy:
comes from underlying platform: 
https://github.com/react-native-community/react-native-device-info/blob/master/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L482

https://developer.android.com/reference/android/location/LocationManager.html#PASSIVE_PROVIDER

> public static final String PASSIVE_PROVIDER
A special location provider for receiving locations without actually initiating a location fix.
>
> This provider can be used to passively receive location updates when other applications or services request them without actually requesting the locations yourself. This provider will return locations generated by other providers. You can query the Location#getProvider() method to determine the origin of the location update. Requires the permission Manifest.permission.ACCESS_FINE_LOCATION, although if the GPS is not enabled this provider might only return coarse fixes.
>
> Constant Value: "passive"

mikehardy:
(our philosophy during implementation of this, since things move semi-quickly on the underlying APIs both platforms, was just to get as much info as possible in native form, shape it similar on both platforms but in general just pass it through since requirements get so specific at project level while APIs underlying change - this way the conduit / this module doesn't have to change much to support everyone while the APIs move)

oferRounds:
@mikehardy - got you! thank you very very much for the detailed answer!

PR #907, opened 2019-12-09
feat(getDeviceToken) add getDeviceToken to get token using DeviceCheck API (ios only)

### LINKED ISSUE #658: Device Check API (iOS 11+)
## Description

Would love to see Device Check API implemented at some point:

https://developer.apple.com/documentation/devicecheck

## Use Case

It uniquely identifies a user, maintaining privacy. This would then be great for preventing a user from doing an action more than once within the app's context or for user administration etc

AntoineDoubovetzky:
## Description

Implements `getDeviceToken` using [DeviceCheck API](https://developer.apple.com/documentation/devicecheck)
(Closes #658 )

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |      ✅     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -14,6 +14,7 @@
 #import <React/RCTUtils.h>
 #import "RNDeviceInfo.h"
 #import "DeviceUID.h"
+#import <DeviceCheck/DeviceCheck.h>
 
 #if !(TARGET_OS_TV)
 #import <WebKit/WebKit.h>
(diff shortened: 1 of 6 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -375,6 +376,27 @@ - (BOOL) isTablet {
     }
 }
 
+RCT_EXPORT_METHOD(getDeviceToken:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
+    if (@available(iOS 11.0, *)) {
```
mikehardy: Nice - may I assume you've tested this on ios < 11? I have ios 9 available as a real device still and of course with simulators, just curious.

Also, I am not super strong on Xcode, is the '@available' syntax limited to a specific Xcode version? I have a small memory that this might be Xcode 11+ only?

To be clear, I am 100% okay with breaking changes, I just follow semver strictly when I do them so I want to be careful + clear :-)

mikehardy:
Wow, thanks @AntoineDoubovetzky - this looks great. I had one question that is semver related, but otherwise I think I just need to test it and merge it. Currently it's just me on this module and I'm pretty busy so please be patient for "real work" (read as: me actually pulling and testing) but I am very responsive for queries/questions. Cheers

ios/RNDeviceInfo/RNDeviceInfo.m
AntoineDoubovetzky: I've tested this on iOS 10.3 :)

Regarding the `@available` syntax, I didn't think about it but it was introduced in Xcode 9, so I guess it's ok:

image
(source [Xcode 9 Release notes](https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW878))

mikehardy: fantastic! thanks for double-checking. I don't see any semver concern then I'll get this packed up and released as soon as I get a chance - until then you can confidently use patch-package to integrate it with the knowledge the next upstream release will contain the identical change. Thanks again

mikehardy: or I guess now that it's committed you can depend on master and the commit hash if that's easier :-)

AntoineDoubovetzky:
Thanks @mikehardy!

mikehardy:
5.4.0 is out, thanks!
https://github.com/react-native-community/react-native-device-info/releases/tag/v5.4.0

Issue #909, opened 2019-12-11
Google deprecating install_referrer intent broadcast


msalpietro:
# Bug

Google is deprecating the install_referrer intent broadcast. Should update package to use new API if needed.

https://android-developers.googleblog.com/2019/11/still-using-installbroadcast-switch-to.html

## Environment info

N/A

Library version: 5.3.1

## Steps To Reproduce

N/A

## Reproducible sample code

N/A

raedhellal:
this is not a bug it is a feature @msalpietro.
Steps to migrate to the referrel library in:
https://developer.android.com/google/play/installreferrer/library.html

msalpietro:
I would agree. Thanks.

jacargentina:
Hi, i dont get it: i'm  not interested on anything about this install referrer thing; but i cant see clearly explained why should i adopt that library? I'm confused. Any advice?

kc-paymi:
> I would agree. Thanks.

Sorry but could you please let me know how do you plan to update the referrer library inside the device info library ? Won't that require us to make these changes to our app every time we set up a new project ? 
I thought you were right in pointing out that the API needs to be updated from the library maintainers in order for all apps to migrate to the new Referrer API.

msalpietro:
We're using the react-native-device-info package but not using the getInstallReferrer method so I haven't actually looked at how the changes would be implemented. I opened this purely because we received a notice from the play store that the install_referrer intent broadcast is being deprecated.

mikehardy:
It does look like there is something to know, but let me note that while I do a lot here, all of us are the maintainers. I have limited time and would love :pray:  a PR that implemented the required API movement here. Possibly as a new API name in this library (since it seems it would require a new dependency for those that want it - which is not everyone). They say "cost of adoption is low" so hopefully it's an easy PR for someone...

carloscuesta:
Should we close this issue since #947 was merged ? @mikehardy

mikehardy:
yeah - closing and about to release now

PR #913, opened 2019-12-13
Utilize install referrer API over BroadcastReceiver.

pickeloe5:
## Description

Google Play is deprecating the install_referrer intent,
they recommend switching to the new [Play Install Referrer library](https://developer.android.com/google/play/installreferrer/library.html)

I've not made any contributions to open source repos yet, and
I'm not super aware of the behavior of the install_referrer intent nor the new library, so
comments/notes/tips/hints are all gold 👍 

Fixed issue #909 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -48,7 +48,7 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    implementation "com.android.installreferrer:installreferrer:1.1"
+    implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1')}"
     def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
     if(firebaseIidVersion){
         implementation "com.google.firebase:firebase-iid:$firebaseIidVersion"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -48,6 +48,7 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
+    implementation "com.android.installreferrer:installreferrer:1.1"
     def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
     if(firebaseIidVersion){
         implementation "com.google.firebase:firebase-iid:$firebaseIidVersion"
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
@@ -1,0 +1,53 @@
+package com.learnium.RNDeviceInfo;
+
+import android.content.SharedPreferences;
+import android.content.Context;
+import android.os.RemoteException;
+
+import com.android.installreferrer.api.InstallReferrerClient;
+import com.android.installreferrer.api.InstallReferrerStateListener;
+import com.android.installreferrer.api.ReferrerDetails;
+
+import static com.android.installreferrer.api.InstallReferrerClient.InstallReferrerResponse;
+
+public class RNInstallReferrerClient implements InstallReferrerStateListener {
+
+  private SharedPreferences sharedPreferences;
+  private InstallReferrerClient referrerClient;
+
+  RNInstallReferrerClient(Context context) {
+    referrerClient = InstallReferrerClient.newBuilder(context).build();
+    sharedPreferences = context.getSharedPreferences("react-native-device-info", Context.MODE_PRIVATE);
+  }
+
+  private String getInstallReferrer() {
+    try {
+      return referrerClient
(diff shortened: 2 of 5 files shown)


android/build.gradle
```diff
@@ -48,6 +48,7 @@ repositories {
 
 dependencies {
     implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
+    implementation "com.android.installreferrer:installreferrer:1.1"
```
mikehardy: could alter this to be a variable just in case people need to override?

```suggestion
    implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1')}"
```

I think

mikehardy:
Well etiquette for PRs in this repo is "do your best", but most importantly "try something", so you're nailing it :-).

I had one suggestion above, and I have a general question which is: are there cases where this API is not available? As in, should we maintain both implementations for now? I want to make sure we don't break existing installs...

pickeloe5:
> I had one suggestion above, and I have a general question which is: are there cases where this API is not available? As in, should we maintain both implementations for now? I want to make sure we don't break existing installs...

That's probably a good concern to have, looking around the install referrer library documentation I'm not finding much about whether or not the new library is universally supported.
Keeping this logic in onReceive might prepare for either possibility.
image

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -0,0 +1,53 @@
...
+import android.content.Context;
+import android.os.RemoteException;
+
+import com.android.installreferrer.api.InstallReferrerClient;
+import com.android.installreferrer.api.InstallReferrerStateListener;
+import com.android.installreferrer.api.ReferrerDetails;
+
+import static com.android.installreferrer.api.InstallReferrerClient.InstallReferrerResponse;
+
+public class RNInstallReferrerClient implements InstallReferrerStateListener {
```
mcuelenaere: Shouldn't [startConnection](https://developer.android.com/reference/com/android/installreferrer/api/InstallReferrerClient.html#startConnection(com.android.installreferrer.api.InstallReferrerStateListener)) be called somewhere? What else is the point of implementing this interface?

pickeloe5: You're definitely right, I freehanded a fix for it rq because I'm a bit busy at work,
I'll have more time to give this later today

mikehardy: Thanks for the review @mcuelenaere !

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -0,0 +1,54 @@
...
+        .getInstallReferrer()
+        .getInstallReferrer();
+    } catch (RemoteException e) {
+      e.printStackTrace();
+      return null;
+    }
+  }
+
+  @Override
+  public void onInstallReferrerSetupFinished(int responseCode) {
```
skantus:
```suggestion
  Needed!
  
  referrerClient.startConnection(new InstallReferrerStateListener() {
    @Overrride
    public void onInstallReferrerSetupFinished(int responseCode) {
```

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -0,0 +1,54 @@
...
+        break;
+      case InstallReferrerResponse.FEATURE_NOT_SUPPORTED:
+        break;
+      case InstallReferrerResponse.SERVICE_UNAVAILABLE:
+        break;
+    }
+  }
+
+  @Override
+  public void onInstallReferrerServiceDisconnected() {
```
skantus: Should add useful comments?

```suggestion
  public void onInstallReferrerServiceDisconnected() {
     // Try to restart the connection on the next request to
     // Google Play by calling the startConnection() method.
```

mikehardy:
Sorry I've let this go a little stale - I've been busy with holidays and then work ramping up as well, just wanted to say I'm happy to merge something once everyone feels this is in shape. I thought it was pretty close last time we looked - just needs to address the comments above. Cheers!

skantus:
@mikehardy any update?

mikehardy:
None yet sorry - my January's hopping at work, causing my open source queue to back up a fair bit. It's definitely in the list and I'm sorry for the delay

mikehardy:
Actually, appears @pickeloe5 hasn't addressed the comments above, so this would be dormant anyway and eventually close if comments aren't addressed

mikehardy:
superceded by #947

PR #914, opened 2019-12-14
feat: getBottomNavigationBarHeight()

Rapsssito:
## Description
Added `getBottomNavigationBarHeight()` that allows ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -69,6 +69,7 @@ RCT_EXPORT_MODULE();
          @"brand": @"Apple",
          @"model": [self getModel],
          @"deviceType": [self getDeviceTypeName],
+         @"bottomNavigationBarHeight": [self getBottomNavigationBarHeight],
      };
 }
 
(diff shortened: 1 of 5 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -488,6 +489,15 @@ - (NSString *) getDeviceTypeName {
     return [DeviceTypeValues objectAtIndex: [self getDeviceType]];
 }
 
+- (int) getBottomNavigationBarHeight {
```
mikehardy: If it is constants I'd suppose you can do it in javascript 🤔

mikehardy:
Hey @Rapsssito 👋 - thanks for this!

I'm not against it in principle, I can see how it would be fundamental to layouts. However, what I wonder is if it might be better suited for https://github.com/th3rdwave/react-native-safe-area-context - I'm trying to steer all the layout-related things that are questions of where you can draw etc over there, as I feel like it is more focused and does a better job in general already.

What do you think?

ios/RNDeviceInfo/RNDeviceInfo.m
Rapsssito: You are right! Thanks for the feedback!

Rapsssito:
@mikehardy you are probably right. I spent a lot of time looking for the repository that could best adapt this method. I didn't know about https://github.com/th3rdwave/react-native-safe-area-context. I will move my PR there.

mikehardy:
@Rapsssito no worries - not many people know about that one yet as it is fairly new, but it appears to be high quality and I've chatted with the author before, this API *should* find a good home there based on the insets they make available I think

PR #932, opened 2020-01-15
[Fix #901] Web charging time typo

### LINKED ISSUE #901: web chargingtime/dischargingtime are deprecated and maybe incorrect
# Feature Request

chargingtime/dischargingtime should be removed

See deprecation notice here https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API

Also it should be `chargingTime` and `dischargingTime` so not sure this is even working...

## Why it is needed

typo issue + non standard web stuff (discouraged for many reasons if I am correct, to avoid abuse etc)

## Possible implementation

Just revert https://github.com/react-native-community/react-native-device-info/commit/692c50e1ad7d5f53c5aa12f866b4c667ed09fbcc

robertherber:
## Description

Fixed issue #901


Johan-dutoit:
LGTM, with the exception that `BatteryManager` is actually deprecated on web (see: https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager)

We should either considering removing this completely or add some warning to the docs (if it's not there already).

robertherber:
> LGTM, with the exception that `BatteryManager` is actually deprecated on web (see: https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager)
> 
> We should either considering removing this completely or add some warning to the docs (if it's not there already).

Good idea - I've added it to the docs. I added a "partial support" flag in the feature list as well. Maybe a definition of "partial support" is in place though - I'd suggest something along the following:

- Web: If the implementation is based on a W3C standard - ie. not deprecated or draft. There will always be the case of some APIs being partially supported by browsers.
- Other: available in last 2 versions of the OS maybe?

I do think it makes sense to keep it - since there is no replacing API. Having it working in some browsers is better than not at all - especially since "it works in some browsers" is such a common case for Web APIs.

mikehardy:
sorry this is sitting still guys - been busy with other things but it won't be forgotten!

PR #933, opened 2020-01-15
[Web] Types and buildtime variables

robertherber:
## Description

Converted web implementation to TypeScript. Also added getApplicationName, getBuildNumber, getBundleId, getLastUpdateTime, getReadableVersion, getVersion and isEmulator to web implementation. 

isEmulator is just checking if it's localhost for now - which I believe to be a fair equivalent of an emulator. Buildtime variables are CRA-compatible environment variables with same names (included two suggestions on ways set it up in the docs): 
- REACT_APP_VERSION => getVersion()
- REACT_APP_BUILD_NUMBER => getBuildNumber()
- REACT_APP_BUNDLE_ID => getBundleId()
- REACT_APP_LAST_UPDATE_TIME => getLastUpdateTime()

Some stuff collided when adding DOM libs to TypeScript, so I set `{ "skipLibCheck": true }` in tsconfig.json so TypeScript doesn't throw on errors that only apply to dependencies. From what I've seen this shouldn't affect the Type Inference in a bad way (it's even the default option when creating an Expo project).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web   |    ✅     |

## Checklist


Johan-dutoit:
This seems to be an uncompleted version of #934?

robertherber:
@Johan-dutoit Yes, I wanted to keep them separate, but the other one depends on this one. Could close this one and just go with the other one I guess :)

PR #934, opened 2020-01-15
Web example

robertherber:
## Description

I created an example project for the web. It's live at [https://robertherber.github.io/react-native-device-info/](https://robertherber.github.io/react-native-device-info/). It serves both the purpose of verifying that the web implementation is working well and also obviously a good example for starting using the library. I uses the default Expo TypeScript setup as a base.

I think somebody with proper repository rights should publish it to https://react-native-community.github.io/react-native-device-info/ or similar, and I'm thinking automate it with the publishing of new packages :) Just update the homepage in package.json and run `yarn deploy`.

Sorry for the heavy pull request - it builds on top of #933! The example project is the only difference.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Web |    ✅     |

## Checklist

* [x] I added a new Example project for web


mikehardy:
@robertherber excellent! Thanks for this batch of fixes - I'll be pretty busy today with meetings outside the office so it might take me a little bit to get to them but I like the idea of more web implementation in general so once I get a chance to look through I'm sure we can get it all in. Great stuff

src/index.ts
```diff
@@ -273,7 +273,12 @@ export function getApiLevelSync() {
 let bundleId: string;
 export function getBundleId() {
   if (!bundleId) {
-    if (Platform.OS === 'android' || Platform.OS === 'ios' || Platform.OS === 'windows') {
+    if (
+      Platform.OS === 'android' ||
```
Johan-dutoit: I'm started to think we should abstract this into a new call, considering it's being duplicated a fair amount.

Something along the lines of:

```
function isSupportedPlatforms() {
  return (
    Platform.OS === "android" ||
    Platform.OS === "ios" ||
    Platform.OS === "windows" ||
    Platform.OS === "web"
  );
}

if (isSupportedPlatforms) {
  //...
}
```

robertherber: Good idea. Just updated to simplify this.

mikehardy:
sorry this is sitting still guys - been busy with other things but it won't be forgotten!

PR #940, opened 2020-01-23
Align and update react-native dependency

friederbluemle:
Just a couple of minor `devDependency` updates and alignments:

- `react-native@0.61.5` uses `react@16.9.0`
- Updated `@types/react-native` to 0.61.2 (the latest released version)

`@types/react-native` _already_ has dependency on a matching version of `@types/react`. Removing the explicit declaration prevents potential build errors when versions mismatch.

Tested by successfully running `yarn analyze` before and after the updates.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,16 +70,15 @@
   "devDependencies": {
     "@react-native-community/bob": "^0.7.0",
     "@react-native-community/eslint-config": "0.0.5",
-    "@types/react": "^16.9.11",
-    "@types/react-native": "^0.60.22",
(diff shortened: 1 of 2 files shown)


mikehardy:
Seems fine to me. I'm not sure why using '~' isn't okay between react-native / react, seems overly cautious to me, but 🤷‍♂ - we're infrastructure in this module, fine being cautious 😄 . I have a nasty habit of just doing yarn --upgrade-latest and then PRing the broken things but that's not great advice. Thanks for helping keep things preened 'round these parts

Issue #912, opened 2019-12-12
Pass installerPackageName on Android


sattaman:
# Feature Request

Pass through installerPackageName on Android so the market the package came from can be identified.

## Why it is needed

We have an app which can be installed via the playstore, via an apk , and also via various stores in China. For analytics, and for our forced updates screen, we need to be able to determine which market place the app was installed from

## Possible implementation

Add a method `getInstallerPackageName()` which would be implemented on Android only, and would pass the result of https://developer.android.com/reference/android/content/pm/PackageManager.html#getInstallerPackageName(java.lang.String) through. 

### Code sample

`console.log("Device Model", DeviceInfo.getInstallerPackageName()); // e.g. com.android.vending`

mikehardy:
I would be willing to accept a PR like this - not sure how this will work out with the deprecation mentioned in #909 - may not be relevant but if it is then it should be noted in the accompanying change to README documenting the new feature

mikehardy:
@codecog code looks like exactly what I'd expect for Java, great! Seems the administrivia (README, example etc) will be most of the work per usual on this module (since the device APIs are usually trivial) :-). Thanks for doing this

codecog:
@mikehardy will write the docs for this with example usage as well and submit all in the same PR. 👍

codecog:
PR opened: https://github.com/react-native-community/react-native-device-info/pull/946

codecog:
@mikehardy think we can close this issue now.

codecog:
@sattaman reply to this issue once you've used the new method and verify it's what you're after.

mikehardy:
I think you're right - closing preliminarily but obviously we can chat more, cheers

PR #946, opened 2020-01-31
feat: add getInstallerPackageName for android

codecog:
Signed-off-by: Josh Kelly <josh@codecog.dev>

## Description

Adds request from Issue #912

`getInstallerPackageName()` provides the ability to fetch the location where an apk was installed from. Return values have been added to the docs.

## Compatibility

| OS            | Implemented |
| --------- | :---------: |
| iOS           |    ❌     |
| Android   |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+## 5.5.0
+
+feat: add getInstallerPackageName for android (#912) (thanks @codecog!)
+
 ## 5.4.4
 
 - feat: add Huawei P Smart devices to notch list (#945) (thanks @sanborN!)
--- a/README.md
+++ b/README.md
@@ -334,6 +334,7 @@ Every API returns a Promise but also has a corresponding API with 'Sync' on the 
 | [getHost()](#gethost)                                             | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getIpAddress()](#getipaddress)                                   | `Promise<string>`   |  ✅  |   ✅    |   ✅    | ❌ |
 | [getIncremental()](#getincremental)                               | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
+| [getInstallerPackageName()](#getinstallerpackagename)             | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getInstallReferrer()](#getinstallreferrer)                       | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ✅ |
 | [getInstanceId()](#getinstanceid)                                 | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getLastUpdateTime()](#getlastupdatetime)                         | `Promise<number>`   |  ❌  |   ✅    |   ❌    | ❌ |
@@ -781,6 +782,23 @@ The internal value used by the underlying source control to represent this build
 ```js
 DeviceInfo.getIncremental().then(incremental => {
   // "4820017"
+});
+```
+
+---
+
+### getInstallerPackageName()
+
+The internal value used by the underlying source control to represent this build.
+
+#### Examples
+
+```js
+DeviceInfo.getInstallerPackageName().then(installerPackageName => {
+  // Play Store: "com.android.vending"
+  // Amazon: "com.amazon.venezia"
+  // Samsung App Store: "com.sec.android.app.samsungapps"
+  // Developer, iOS: "unknown"
 });
 ```
 
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -504,6 +504,20 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     return getReactApplicationContext().getPackageManager().getPackageInfo(getReactApplicationContext().getPackageName(), 0);
   }
 
+  private String getInstallerPackageNameSync() {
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -786,6 +787,23 @@ DeviceInfo.getIncremental().then(incremental => {
...
+The internal value used by the underlying source control to represent this build.
+
+#### Examples
+
+```js
+DeviceInfo.getInstallerPackageName().then(installerPackageName => {
+  // Play Store: "com.android.vending"
+  // Amazon: "com.amazon.venezia"
+  // Samsung App Store: "com.sec.android.app.samsungapps"
+  // Developer, iOS: "null"
```
mikehardy:
```suggestion
  // Developer, iOS: "unknown"
```

(that's actually what you are already returning, and is in harmony with our stated "default" values for unsupported platforms)

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -504,6 +504,14 @@ private PackageInfo getPackageInfo() throws Exception {
     return getReactApplicationContext().getPackageManager().getPackageInfo(getReactApplicationContext().getPackageName(), 0);
   }
 
+  private String getInstallerPackageNameSync() {
+    String packageName = getReactApplicationContext().getPackageName();
+    return getReactApplicationContext().getPackageManager().getInstallerPackageName(packageName);
```
mikehardy: getInstallerPackageName is [documented to possibly return null](https://developer.android.com/reference/android/content/pm/PackageManager#getInstallerPackageName(java.lang.String)), instead of having a tri-state return value (something useful, "unknown", or null) I think it's a better dev experience if it's only 2 states (something useful or unknown). 

```suggestion
    String installerPackageName = getReactApplicationContext().getPackageManager().getInstallerPackageName(packageName);
    if (installerPackageName == null) {
      return "unknown";
    }
    return installerPackageName;
```

An alternative would be to check for null return in the javascript and mutate it to unknown there, but I think handling it in Java is fine. I just coded that in the comment text box so triple check it, if you like it, before committing :-)

example/App.js
```diff
@@ -159,6 +159,7 @@ export default class App extends Component {
       deviceJSON.userAgent = await DeviceInfo.getUserAgent();
       deviceJSON.instanceId = await DeviceInfo.getInstanceId();
       deviceJSON.installReferrer = await DeviceInfo.getInstallReferrer();
+      deviceJSON.installerPackageName = DeviceInfo.getInstallerPackageName();
```
mikehardy:
```suggestion
      deviceJSON.installerPackageName = await DeviceInfo.getInstallerPackageName();
```

src/index.ts
```diff
@@ -282,6 +282,19 @@ export function getBundleId() {
...
+  if (!installerPackageName) {
+    if (Platform.OS === 'android') {
+      installerPackageName = await RNDeviceInfo.getInstallerPackageName();
+    } else {
+      installerPackageName = 'unknown';
+    }
+  }
+
+  return installerPackageName;
+}
```
mikehardy: Can you add the sync variant as well? In general I like them at parity unless there is something inherently async about the API making it unreasonable (like a network access) or people have petitioned for it to be constant (like some fundamental to bootstrap values)

This is in the middle area where I think both variants are best to cover everyone as best as possible without it being a constant

codecog:
updated the PR with the requested changes.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+## 5.5.0
+
+feat: add getInstallerPackageName for android (#912)
```
mikehardy:
```suggestion
feat: add getInstallerPackageName for android (#912) (thanks @codecog!)
```

mikehardy: :-)

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -504,6 +504,20 @@ private PackageInfo getPackageInfo() throws Exception {
     return getReactApplicationContext().getPackageManager().getPackageInfo(getReactApplicationContext().getPackageName(), 0);
   }
 
+  private String getInstallerPackageNameSync() {
+    String packageName = getReactApplicationContext().getPackageName();
+    String installerPackageName = getReactApplicationContext().getPackageManager().getInstallerPackageName(packageName);
+
+    if (installerPackageName === null) {
```
mikehardy:
```suggestion
    if (installerPackageName == null) {
```

triple-equals is a javascript-ism, I don't think it will compile in Java?

codecog: You're right, i'm in JS land :D

example/App.js
```diff
@@ -159,6 +159,7 @@ export default class App extends Component {
       deviceJSON.userAgent = await DeviceInfo.getUserAgent();
       deviceJSON.instanceId = await DeviceInfo.getInstanceId();
       deviceJSON.installReferrer = await DeviceInfo.getInstallReferrer();
+      deviceJSON.installerPackageName = await DeviceInfo.getInstallerPackageName();
```
mikehardy: I just realized (as I was thinking through where I'll merge this then test it) that now that you added the sync variant (thank you!) you need to add a the same above in the sync section, can you do that :pray:  then I'll merge, do a quick local build+check and release? It's the only thing I can see now, everything else looks in great shape

codecog: not a problem, have added that example.

codecog:
Thanks, @mikehardy, likewise!

PR #947, opened 2020-02-01
install_referrer api over soon to be deprecated

skantus:
## Description

Use new `InstallReferrerClient` before to be deprecated by Google Play Store.

Fixed issue <https://github.com/react-native-community/react-native-device-info/issues/909>

Documentation:

- https://developer.android.com/google/play/installreferrer/library.html#java
- https://mvnrepository.com/artifact/com.android.installreferrer/installreferrer

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌  |
| Android |     ✅   |
| Windows |    ❌   |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
@@ -3,58 +3,71 @@ package com.learnium.RNDeviceInfo;
 import android.content.SharedPreferences;
 import android.content.Context;
 import android.os.RemoteException;
+import android.util.Log;
 
 import com.android.installreferrer.api.InstallReferrerClient;
 import com.android.installreferrer.api.InstallReferrerStateListener;
 import com.android.installreferrer.api.ReferrerDetails;
 
-import static com.android.installreferrer.api.InstallReferrerClient.InstallReferrerResponse;
-
-public class RNInstallReferrerClient implements InstallReferrerStateListener {
+public class RNInstallReferrerClient {
 
   private SharedPreferences sharedPreferences;
-  private InstallReferrerClient referrerClient;
+  private InstallReferrerClient mReferrerClient;
 
   RNInstallReferrerClient(Context context) {
     sharedPreferences = context.getSharedPreferences("react-native-device-info", Context.MODE_PRIVATE);
-    referrerClient = InstallReferrerClient.newBuilder(context).build();
-    referrerClient.startConnection(this);
+    mReferrerClient = InstallReferrerClient.newBuilder(context).build();
+    mReferrerClient.startConnection(installReferrerStateListener);
   }
 
   private String getInstallReferrer() {
     try {
-      return referrerClient
-        .getInstallReferrer()
-        .getInstallReferrer();
+      return mReferrerClient
+              .getInstallReferrer()
+              .getInstallReferrer();
     } catch (RemoteException e) {
       e.printStackTrace();
       return null;
     }
   }
 
-  referrerClient.startConnection(new InstallReferrerStateListener() {
-    @Override
-    public void onInstallReferrerSetupFinished(int responseCode) {
-      switch (responseCode) {
-        case InstallReferrerResponse.OK:
-          // Connection established
-          SharedPreferences.Editor editor = sharedPreferences.edit();
-          editor.putString("installReferrer", getInstallReferrer());
-          editor.apply();
-          break;
-        case InstallReferrerResponse.FEATURE_NOT_SUPPORTED:
-          // API not available on the current Play Store app
-          break;
-        case InstallReferrerResponse.SERVICE_UNAVAILABLE:
-          // Connection could not be established
-          break;
-      }
-    }
+  private InstallReferrerStateListener installReferrerStateListener =
+    new InstallReferrerStateListener() {
+      @Override public void onInstallReferrerSetupFinished(int responseCode) {
+        switch (responseCode) {
+          case InstallReferrerClient.InstallReferrerResponse.OK:
+            // Connection established
+            try {
+              if (BuildConfig.DEBUG) Log.d("InstallReferrerState", "OK");
+              ReferrerDetails response = mReferrerClient.getInstallReferrer();
+              response.getInstallReferrer();
+              response.getReferrerClickTimestampSeconds();
+              response.getInstallBeginTimestampSeconds();
 
-    @Override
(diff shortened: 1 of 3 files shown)


skantus:
@mikehardy I created a new PR because https://github.com/react-native-community/react-native-device-info/pull/913 never responded to the suggested comments re-consider moving forward with this new review.

Thanks.

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -0,0 +1,60 @@
...
+    public void onInstallReferrerSetupFinished(int responseCode) {
+      switch (responseCode) {
+        case InstallReferrerResponse.OK:
+          // Connection established
+          SharedPreferences.Editor editor = sharedPreferences.edit();
+          editor.putString("installReferrer", getInstallReferrer());
+          editor.apply();
+          break;
+        case InstallReferrerResponse.FEATURE_NOT_SUPPORTED:
+          // API not available on the current Play Store app
```
mikehardy: Everything looks great in this PR, with the one thought that it would be nice to expose the two states (not supported and unavailable) and persist them as well, simply as "not_supported" and "service_unavailable" perhaps?) plus to have a default ("invalid_response" perhaps?) just in case, and document those as possible return values. Seems like we're potentially losing information and may have a higher github issue support load in the future otherwise ("why isn't install referrer returning anything?")

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -0,0 +1,60 @@
...
+        case InstallReferrerResponse.SERVICE_UNAVAILABLE:
+          // Connection could not be established
+          break;
+      }
+    }
+
+    @Override
+    public void onInstallReferrerServiceDisconnected() {
+      // Try to restart the connection on the next request to
+      // Google Play by calling the startConnection() method.
```
mikehardy: would be nice to log something here just in case, using default react-native logging facilities and maybe info level to start (until we get a feel for whether that's right or it should be debug or maybe warn)

skantus:
> Thank you for doing this! The comments I left are mostly geared towards making sure we don't silently throw away any information that might help developers help themselves before needing to ask for help. I haven't checked all of the possible return values (do any of the APIs return null? do any of them throw RuntimeExceptions in the normal usage, that we should handle in order to avoid a crash?) but that will be the next phase of review - you could pre-empt it by verifying each API touched has been investigated for nulls and exceptions, and then I think we can merge pretty quickly. Thanks again

Ok @mikehardy I just did some modifications to make sure it works well, I have also done a couple of tests and added a couple of native logs.

Minishlink:
@mikehardy 

Shouldn't the leftover of the old way of retrieving install referrer be deleted? I'm thinking about these:

* https://github.com/react-native-community/react-native-device-info/blob/cb717537ec21da4c596d7598d59732a69de8b298/android/src/main/AndroidManifest.xml#L6
* https://github.com/react-native-community/react-native-device-info/blob/17aae94ccd19312bb74696ad57fbb4bf2b9a6f0e/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceReceiver.java#L12

mikehardy:
@Minishlink possibly? What's the impact on the installed base from a few perspectives:

1. the new implementation requires a dependency some are uncomfortable with from what I understand, delaying the move for them until there is no other way to get the information seems prudent? When is that? When the Android API levels sunset? Or sooner?
1. does the new implementation cover the entire Android API skew of react-native? (currently API16)
1. is the new implementation backwards compatible in results for the same app install source?

I think the answer to the third is easy and yes. The second I have not checked but is a maybe yes, the third - if I am correct and the old way of doing it will survive in the old Android API levels - makes me want to deprecate it here but keep it, until the APIs themselves are gone

Minishlink:
So apparently new versions in the PlayStore will stop working with the Intent, and it's not related to the Android API level. https://android-developers.googleblog.com/2019/11/still-using-installbroadcast-switch-to.html

It sunsets tomorrow. And the new API is apparently available since 2017. Since the Play Store updates itself, I'd think it's safe to remove.

I don't have the answer for this function being in rn-device-info or not: it depends on if you want to include Google Play specific info or not :)

Thanks

mikehardy:
@minishlink that sounds about as good to purge as anything thanks for checking that out. Happy to merge a PR if you want to claim the internet points 😂😁

Minishlink:
@mikehardy Haha, I'll probably do this in March :) To be clear: do you want to remove only the old API or the whole install referrer thing?

mikehardy:
Just the old API, if the play store app won't be responding to it, no reason for it to stay

eftikharEmad:
I just update the package to 5.5.3  I am getting this `InstallReferrerState:FEATURE_NOT_SUPPORTED` any I dea, using android 28 , react-native `0.61.3`

skantus:
> I just update the package to 5.5.3 I am getting this `InstallReferrerState:FEATURE_NOT_SUPPORTED` any Idea, using android 28 , react-native `0.61.3`, should I make any changes beside updating the package version ? thank you

Looks like the API not available on the current Play Store app?

You could see more here details:

https://developer.android.com/google/play/installreferrer/library

skantus:
> I just update the package to 5.5.3 I am getting this `InstallReferrerState:FEATURE_NOT_SUPPORTED` any Idea, using android 28 , react-native `0.61.3`, should I make any changes beside updating the package version ? thank you

Hi @eftikharEmad,

Looks like the API not available on the current Play Store app?

I mean this can happen because the user is handling an old version of the play store or you are trying  install the app in some old emulator old real device.

You could see more details here: https://developer.android.com/google/play/installreferrer/library

Hope it help.

eftikharEmad:
the emulator is android 9 , I already add the dependency but I am getting the same 
I am using only the `getUniqueId` from the package. will it effected hugely by this 
😕😕😕

skantus:
Did you try using `getAndroidId` instead?

https://github.com/react-native-community/react-native-device-info#getandroidid

eftikharEmad:
> Did you try using `getAndroidId` instead?
> 
> https://github.com/react-native-community/react-native-device-info#getandroidid

actually `getUniqueId` is working fine with me, I just did not understand this `InstallReferrerState:FEATURE_NOT_SUPPORTED` , so I we will postponed the update since we are `NOT` using the `Google Play Campaign Measurement` and the app is working 
 😅😅😅

PR #965, opened 2020-03-03
[Android] Remove deprecated way of retrieving INSTALL_REFERRER

Minishlink:
## Description

Related to #947 (see comments)

See https://android-developers.googleblog.com/2019/11/still-using-installbroadcast-switch-to.html

FYI, Google also removed that from Google App Measurements (used by Google Analytics)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* ~I added the documentation in `README.md`~
* ~I updated the typings files (`privateTypes.ts`, `types.ts`)~
* ~I added a sample use of the API (`example/App.js`)~


mikehardy:
I gave this a quick scan and it looks fine to start, thank you! I'll look a little more deeply when I have a minute (may take a day or three though - I just got back from travel) to make sure then merge and release.

mikehardy:
Sorry I let this sit so long! I just saw it again, my apologies.
I like it and will merge it but it will be a breaking change (no big deal but I take semantic versioning seriously) so the changelog will be for a 6.0. Also I have broken the PR with past merges - I just the conflicts but I have a question now:

The AndroidManifest, without the listener here, is empty. I think the PR could be updated to simply delete the AndroidManifest and it would still work

Can you try that and if it works commit a delete of that file?

Thank you!

mikehardy:
No response from submitter

PR #966, opened 2020-03-08
fix: memoization of boolean values.

jmheik:
## Description

Fixes memoization of boolean values. Previously memoization worked only when value was `true`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+## 5.5.4
+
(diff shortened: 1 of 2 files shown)


mikehardy:
This looks good! Sorry for the delay, things are a little crazy right now, no? I appreciate the changelog edit also, that helps me out. Hoping for a release shortly

PR #996, opened 2020-04-14
Fix issue where getModel on iOS would never return the generic names, only "unknown" if the real device name wasn't found.

TheAlmightyBob:
## Description

This appears to have been a simple typo in the iOS implementation of `getModel`. It was checking if `deviceName` was `nil`, and it if _was_, then it was continuing to parse `deviceName`. This was clearly not the intent.

This is such a small fix that I wasn't sure it was worth creating an issue for or adding to `CHANGELOG.md`, but happy to do so if others disagree.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

(this fix doesn't affect Android/Windows...)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - fix: memoization of static boolean return values (@jmheik)
 - fix: add mac catalyst compatibility to getCarrier (#973) (thanks @robertying!)
 - fix: add wider exception handling in install referrer (thanks @jmunozDevsu!)
+- fix: getModel on iOS now returns generic device type (e.g. "iPhone") if the specific model is unrecognized (@TheAlmightyBob)
 
 ## 5.5.3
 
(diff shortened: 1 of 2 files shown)


TheAlmightyBob:
@mikehardy I actually didn't have a use case that hit this, I was just looking at the code and thought it looked wrong. I tested it (before and after) by temporarily removing my test device from the list.

I'll update the changelog.

I don't know what's up w/ the Windows check? There's no way this is responsible for that...

mikehardy:
There is no way this is responsible for the windows check, no - that's a (forgive me) broken window at the moment and I just keep driving by it...we'll get to it eventually.

Test seems valid, thanks!

mikehardy:
v5.5.5 exists :-)

TheAlmightyBob:
You had me worried for a sec that the changelog would have the wrong version, thanks for updating that...

PR #1024, opened 2020-05-27
Fix: TypeError: Cannot assign to read only property 'exports' of object '#<Object>'

jamesst20:
Fix issue: TypeError: https://github.com/react-native-community/react-native-device-info/issues/950

```
Quoting: https://github.com/webpack/webpack/issues/4039

You can mix require and export. You can't mix import and module.exports.
```

I simply replaced the module.exports syntax with 'export const' and it does fix the error.

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -6,6 +6,18 @@ let batteryCharging = false,
   batteryLevel = -1,
   powerState = {};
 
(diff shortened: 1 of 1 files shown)


mikehardy:
v5.6.1 is out
https://github.com/react-native-community/react-native-device-info/releases/tag/v5.6.1

PR #1050, opened 2020-07-21
Fix: WKWebview is release

### LINKED ISSUE #774: Parallel calls to getUserAgent() cause a runtime exception
# Bug

Hello. I've used this package for years and I really like it. Thanks for your hard work!

Today I updated to the 3.0.0 version. I noticed some odd behavior, and have narrowed it down to the fact that parallel calls to the DeviceInfo.getUserAgent() function cause one of them to throw an exception.

I think I'm to a point where I'm no longer calling duplicate functions at the same time with the help of async await, so I've worked around the issue .... here's my end-result commit of going from 2.3.2 to 3.0.0 for reference : https://github.com/OurVoiceUSA/HelloVoter/commit/4cb41e3538f73932cfec411b7f03bd1e42bbb42e

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
    Memory: 374.36 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - /usr/local/bin/node
    npm: 6.11.3 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 27, 28, 29
      Build Tools: 28.0.3, 29.0.1, 29.0.2
      System Images: android-28 | Google Play Intel x86 Atom, android-29 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: ^16.9.0 => 16.9.0 
    react-native: ^0.60.5 => 0.60.5 
  npmGlobalPackages:
    react-native-cli: 2.0.1
```

Library version: 3.0.0

## Steps To Reproduce

Call the same function twice before the promise resolves, like this:

```bash
  componentDidMount() {
    DeviceInfo.getUserAgent().then(i => this.setState({UserAgent: i})).catch(e => console.warn(e));
    DeviceInfo.getUserAgent().then(i => this.setState({UserAgent: i})).catch(e => console.warn(e));
  }
```

Describe what you expected to happen:

No exception should be thrown.

## Reproducible sample code

You can init a new project like so 

```bash
react-native init AwesomeProject
cd AwesomeProject
npm install react-native-device-info
react-native link react-native-device-info
(cd ios && pod install)
```

Have `App.js` be the following:

```bash
import React, {PureComponent} from 'react';
import {
  View,
  Text,
} from 'react-native';
import DeviceInfo from 'react-native-device-info';

export default class App extends PureComponent {

  constructor(props) {
    super(props);
    this.state = {
      UserAgent: "unknown",
    };
  }

  componentDidMount() {
    DeviceInfo.getUserAgent().then(i => this.setState({UserAgent: i})).catch(e => console.warn(e));
    DeviceInfo.getUserAgent().then(i => this.setState({UserAgent: i})).catch(e => console.warn(e));
  }

  render() {
    const { UserAgent } = this.state;
    return (
      <View style={{flex: 1, margin: 50}}>
        <Text>Hello DeviceInfo! UserAgent => {UserAgent}</Text>
      </View>
    );
  }
};
```

And finally `react-native run-ios` and you'll see the warning from the caught exception. In this example it's complaining about WKErrorDomain, and at some point the message said something about thread safety, and I didn't capture the message in detail.

RoJoHub:
## Description

Fixes #774 
We can not use structures and @Property to setter . And must use __block
Otherwise，wkwebview might deallocate before we ask for user Agent

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |

## Checklist

* [x] I updated the iOS files (`RNDeviceInfo.m`)

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -37,9 +37,6 @@ typedef NS_ENUM(NSInteger, DeviceType) {
 
 @implementation RNDeviceInfo
 {
-    #if !(TARGET_OS_TV)
-    WKWebView *webView;
-    #endif
     bool hasListeners;
 }
 
@@ -724,16 +721,16 @@ RCT_EXPORT_METHOD(getUserAgent:(RCTPromiseResolveBlock)resolve rejecter:(RCTProm
         __strong RNDeviceInfo *strongSelf = weakSelf;
         if (strongSelf) {
             // Save WKWebView (it might deallocate before we ask for user Agent)
-            strongSelf->webView = [[WKWebView alloc] init];
+            __block WKWebView *webView = [[WKWebView alloc] init];
 
-            [strongSelf->webView evaluateJavaScript:@"window.navigator.userAgent;" completionHandler:^(id _Nullable result, NSError * _Nullable error) {
+            [webView evaluateJavaScript:@"window.navigator.userAgent;" completionHandler:^(id _Nullable result, NSError * _Nullable error) {
                 if (error) {
                     reject(@"getUserAgentError", error.localizedDescription, error);
-                    return;
+                }else{
+                    resolve([NSString stringWithFormat:@"%@", result]);
                 }
-                resolve([NSString stringWithFormat:@"%@", result]);
                 // Destroy the WKWebView after task is complete
-                strongSelf->webView = nil;
+                webView = nil;
             }];
         }
     });


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -724,16 +722,17 @@ - (unsigned long) getUsedMemory {
...
+            [webView evaluateJavaScript:@"window.navigator.userAgent;" completionHandler:^(id _Nullable result, NSError * _Nullable error) {
                 if (error) {
                     reject(@"getUserAgentError", error.localizedDescription, error);
+                    webView = nil
                     return;
                 }
                 resolve([NSString stringWithFormat:@"%@", result]);
                 // Destroy the WKWebView after task is complete
-                strongSelf->webView = nil;
+                webView = nil;
```
mikehardy, line 735: Interesting! My only comment is that now we are "cleaning up" (`webView = nil`) in two places, and I strongly prefer to clean up in one place only if possible.

I think the `webView = nil` could go on top of the `if (error) {` as we no longer need the webView anymore do we? Then there is just one `webView = nil`

RoJoHub: 😀
It's a good idea. 

----
In fact, before submitting pr, I also want to do this, but I want to try not to change the code
So.....

RoJoHub:
> Thanks for working on this @RoJoHub ! I found that I was able to reproduce this by simply altering the example app from this repo to have a few calls to getUserAgent in a row, were you able to reproduce this first, and then verify it still works and you no longer reproduce it afterwards?
> 
> My only suggested change is really small - just to consolidate the nil of webview to one spot after the function call, what do you think?

Hi, I created a demo to reproduce this problem.
Demo is a very extreme example, and it is like a Black-Box Testing.
In the demo, I checked the reference count and broke the `error`.
Only the setWebViewWithBlock method runs normally.

Please experience my demo.
https://github.com/RoJoHub/wkviewDemo

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -724,16 +721,16 @@ - (unsigned long) getUsedMemory {
...
                 if (error) {
                     reject(@"getUserAgentError", error.localizedDescription, error);
-                    return;
+                }else{
+                    resolve([NSString stringWithFormat:@"%@", result]);
                 }
-                resolve([NSString stringWithFormat:@"%@", result]);
                 // Destroy the WKWebView after task is complete
-                strongSelf->webView = nil;
+                webView = nil;
```
RoJoHub, line 733: New logic

mikehardy:
released as 5.6.3 🚀  - thanks again

RoJoHub:
> released as 5.6.3 🚀 - thanks again

WOW. Thanks

PR #1054, opened 2020-08-10
Fix Android devices total capacity calculation

gabriel-franzoni-pier:
## Description

Fixing issue while getting android total capacity. Sum up root directory and data directory to get total capacity.

Add `getTotalDiskCapacityNew` and `getFreeDiskStorageNew` - and their sync variants. Calculations now take into account both `Root` and `Data` file systems.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


gabriel-franzoni-pier:
Great points, @mikehardy. Since I haven't used the `getFreeDiskStorageSync` method I did not realize it existed.
I was reading the official Android docs and I found out the API used (`getExternalStorageDirectory`) is deprecated from android Q (API level 29) and beyond.

This implicates that we use `File`'s `getTotalSpace()` instead of `getExternalStorageDirectory()` which returns a `StatsFs`.

And yes, I agree with you - this would be a breaking change... IMHO this can be considered a bug, since for some devices it was returning a different number that the user would see at his Settings. We used to show users which device they had (Galaxy S5 32GB or 16GB?) and we had some weird results when not taking into account both file systems.

Do you think it's still ok to make that change and modify the `getFreeDiskStorageSync` method accordingly? I would also have to change the `README.md` :)


Source: [Android docs - deprecation](https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory())

gabriel-franzoni-pier:
Also I just realized the class is using `getBlockCountLong` and not supporting API < 18 with `getBlockCount`, but we have a predicate for comparison on `getTotalDiskCapacitySync` which enable the support for both `getAvailableBlocksLong` and `getAvailableBlocks`.

Since the min API level for the project is 16, should we add the support for this method, as well?

mikehardy:
Hey there!

- definitely want to maintain API16 support, this module is one of the more important in the RN ecosystem and I will track RN minSdkVersion (16) exactly, so if you could do both that would be perfect
- thinking about how to implement this, might be best to deprecate the existing methods and add new method names entirely, then it's not even breaking, but the implementation is free to be as modern as we want. What I mean specifically is that you can then be very opinionated about what is included and what is not when doing the calculation, and as long as it's documented and a new method name, no problem ?

gabriel-franzoni-pier:
Great suggestion, I will make those changes, then :)

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -329,6 +329,65 @@ public String getCarrierSync() {
   @ReactMethod
   public void getCarrier(Promise p) { p.resolve(getCarrierSync()); }
 
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  public double getTotalDiskCapacityNewSync() {
```
gabriel-franzoni-pier, line 333: Apparently there's a new API for API >= 30 (android 11) [getStorageDirectory](https://developer.android.com/reference/android/os/Environment#getStorageDirectory())

mikehardy: :laughing:  never a dull moment as the APIs march along. Only 14 API advances from now we'll be able to remove that version code check ;-)

README.md
```diff
@@ -330,6 +330,7 @@ Most APIs return a Promise but also have a corresponding API with `Sync` on the
 | [getFingerprint()](#getfingerprint)                               | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getFontScale()](#getfontscale)                                   | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ❌ |
 | [getFreeDiskStorage()](#getfreediskstorage)                       | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
+| [getFreeDiskStoragenNew()](#getfreediskstoragenew)                 | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
```
mikehardy, line 333:
```suggestion
| [getFreeDiskStorageNew()](#getfreediskstoragenew)                 | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
```

simple typo double-tap on that 'n'

README.md
```diff
@@ -355,6 +356,7 @@ Most APIs return a Promise but also have a corresponding API with `Sync` on the
 | [getTags()](#gettags)                                             | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getType()](#gettype)                                             | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌ |
 | [getTotalDiskCapacity()](#gettotaldiskcapacity)                   | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
+| [getTotalDiskCapacityNew()](#gettotaldiskcapacityNew)             | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
```
mikehardy, line 359:
```suggestion
| [getTotalDiskCapacityNew()](#gettotaldiskcapacitynew)             | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
```

I think the page anchors have to be all lower case?

src/index.ts
```diff
@@ -977,27 +977,55 @@ export async function getTotalDiskCapacity() {
   return -1;
 }
 
+export async function getTotalDiskCapacityNew() {
+  if (Platform.OS === 'android' || Platform.OS === 'ios' || Platform.OS === 'web') {
+    return RNDeviceInfo.getTotalDiskCapacityNew();
```
mikehardy, line 982: If there is no new code for ios or web, those need to delegate to the old method don't they? I think they will crash if it runs like this

src/index.ts
```diff
@@ -977,27 +977,55 @@ export async function getTotalDiskCapacity() {
...
 export function getFreeDiskStorageSync() {
   if (Platform.OS === 'android' || Platform.OS === 'ios' || Platform.OS === 'web') {
     return RNDeviceInfo.getFreeDiskStorageSync();
   }
   return -1;
 }
 
+export function getFreeDiskStorageNewSync() {
+  if (Platform.OS === 'android' || Platform.OS === 'ios' || Platform.OS === 'web') {
+    return RNDeviceInfo.getFreeDiskStorageNewSync();
```
mikehardy, line 1024: this is the last of this block of 4 but the same question ("will it crash on ios and web since those did not get a new native API?") apply to all 4

mikehardy:
This is looking really good thank you! I had a couple nit-picks on the API document table in the README, but one real question with regard to javascript delegating (or not) to the old native APIs on ios and web - I think we need to?

Otherwise this looks close to merge-able. I have one more item I want to think about as current manager of the packager in general which is that I might want to plan the obsolescence of the old API explicitly and do a breaking change instead by naming existing APIs xxxOld() with a deprecation tag, and making your APIs handle the main / original names. Don't make any related changes like that yet though - in fact if I do that I would probably just merge as is then re-shape the APIs. It is just something I need to think about - the long-term vision

gabriel-franzoni-pier:
Yup, I was wondering about the API naming since we don't want a breaking change for now. I guess we have to check the userbase reaction first and then make a breaking change in a future version, that might be the way to go.
I will make the requested changes, sorry for the sloppiness 😅

gabriel-franzoni-pier:
@mikehardy sorry for bother you, but what are your thoughts about this issue and a new release?
I am planning to release a new version of one of my apps - and I could include the newer version of this library, if it is ready until the release date

mikehardy:
@gabriel-franzoni-pier no bother - I really appreciate the PR, I have it on my list for review today, so will have more info in a couple hours

mikehardy:
This all still looks good to me - I believe since this will result in numbers that match what the settings app returns for many phones (most phones?) it's the right way to go for the "main" (not "xxxNew()") API names, and I'm going to re-name the old functions to xxxOld() for people that still need them, then if they still really really need them after they go away I'll ask for PRs that allow you to direct which filesystems to include if people really want that. But that's not for this PR

So I'll merge this, then I'll rename things really quickly, and we can do a major version so people are aware something's up

Thanks again for doing this - the numbers returned from those APIs were a sore point from before I became the current maintainer!

mikehardy:
Alright @gabriel-franzoni-pier if you want to depend on this commit hash https://github.com/react-native-community/react-native-device-info/commit/f617863017dce0daf623504e7987652561f4ee49 that's the version of this that will ship as 6.0.0 - I'm still waiting to hear on a separate breaking PR that I'd like to ship with 6.0 (#1059 ) so it might be a bit yet but this code is now ready to ship I think

mikehardy:
This is live now, enjoy! 

https://github.com/react-native-community/react-native-device-info/releases/tag/v6.0.0

PR #1062, opened 2020-08-22
[WIP] Added get screen dimensions

gh-clement-taboulot:
## Description

Added `getScreenDimensions()` that allows to get some screen specifications.

On iOS we can get screen dimensions in pt, px and mm.
On Android we can get screen dimensions in dp, px, mm. => **WIP**

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


mikehardy:
will close shortly in absence of feedback

taboulot:
@mikehardy 

Sorry for the lateness of my response I was off.

I understand your hesitation. But as react-native-device-info user I really see it as the place where I can find informations relative to my device specifications: model, macadress, etc.. 🧐

And I see screen dimensions as a device specification.

I asked to developers around me and they had the same feeling 😄

mikehardy:
I empathize with the point of view but the *safely useful* portion of the screen (which is the useful answer) is quite dynamic. It is not an intrinsic part of the device, it changes based on usage and current circumstance. I maintain react-native-safe-area-context is the correct home, and is actively developed. If they won't accept the functionality then we can see about it here - if that's the case in the end come back here and link the PR they rejected and we'll handle that as it may happen

mrousavy:
It's also worth mentioning that the **Dimensions** API in react-native is completely broken since a really long time and facebook didn't really care about that. See: https://github.com/react-native-community/discussions-and-proposals/issues/291

PR #1063, opened 2020-08-24
fix: fix getInstanceId() wrong link

KingAmo:
## Description

Fixes wrong link of `getInstanceId()` in readme

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1654,7 +1654,7 @@ When installing or using `react-native-device-info`, you may encounter the follo
 <details>
   <summary>[android] - Unable to merge dex / Multiple dex files / Problems with `com.google.android.gms`</summary>
 
-`react-native-device-info` uses `com.google.android.gms:play-services-gcm` to provide [getInstance()][#getinstance].
+`react-native-device-info` uses `com.google.android.gms:play-services-gcm` to provide [getInstanceId()][#getinstanceid].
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1654,7 +1654,7 @@ When installing or using `react-native-device-info`, you may encounter the follo
 <details>
   <summary>[android] - Unable to merge dex / Multiple dex files / Problems with `com.google.android.gms`</summary>
 
-`react-native-device-info` uses `com.google.android.gms:play-services-gcm` to provide [getInstance()][#getinstance].
+`react-native-device-info` uses `com.google.android.gms:play-services-gcm` to provide [getInstanceId()][#getinstanceid].
```
mikehardy, line 1657: Strangely, in the details section none of the links worked. So my suggestion here is just to describe it instead of linking. It was definitely broken before though, good catch! What do you think of this?

```suggestion
`react-native-device-info` uses `com.google.android.gms:play-services-gcm` to provide the implementation for Instance ID fetching.
```

KingAmo: sorry,there is another mistake, it should be '()' not '[]', i should have check it

Issue #554, opened 2018-12-25
wish set android:exported="false" , OK??


suwu150:
## Summary
```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
          package="com.learnium.RNDeviceInfo">
          
    <application>
        <receiver
            android:name="com.learnium.RNDeviceInfo.RNDeviceReceiver"
            android:enabled="true"
            android:exported="false"
        >
            <intent-filter>
                <action android:name="com.android.vending.INSTALL_REFERRER" />
            </intent-filter>
        </receiver>
    </application>

</manifest>
```

the device is android,i wish set android:exported="false".

## Current behavior
```
android:exported="true".
```
## Expected behavior

```
android:exported="false"
```

machour:
Why?

mayurssoni2456:
I have the same question - do we need exported to be "true"? any thoughts

mikehardy:
I believe it must be true yes. The docs indicate it has to be true to receive broadcasts from outside the app package, and I believe this is the installer sending a broadcast so you'd only receive if it were true.

But feel free to try it either way, I could be wrong about the installer package specifically?

https://developer.android.com/guide/topics/manifest/receiver-element

hpelitebook745G2:
@mikehardy my app was previously flagged vulnerable with low risk rating specifically on the AndroidManifest.xml since the receiver was `exported="true"`, can I set it to false since I'm only using the library to get the basic info like getVersion()?

mikehardy:
You may do whatever you like.

PR #1065, opened 2020-08-24
Protect exported broadcast receiver with install packages permission.

### LINKED ISSUE #452: MobSF issue (com.learnium.RNDeviceInfo.RNDeviceReceiver) is not Protected.
Did some check of my apk in MobSF
and got
(com.learnium.RNDeviceInfo.RNDeviceReceiver) is not Protected. [android:exported=true]

A Broadcast Receiver is found to be shared with other apps on the device therefore leaving it accessible to any other application on the device.
screen shot 2018-07-19 at 5 10 14 pm

leoek:
## Description

protects the exported broadcast receiver which listens for app installation intents with the INSTALL_PACKAGES permission to limit it's usage to broadcasters which are able to install packages (usually only the Play Store). This avoids warnings of scanners which check for publicly exported broadcast receivers and restricts that broadcast receiver as much as possible. With the permission set, the users has to explicitly grant an app the install packages permission for it to use the broadcast receiver. (Most users probably won't have any app other than the play store with that permission.) Unfortunately it will still show up as false-positive in MobSF, but with a longer explanation and note about permissions.

Fixes #452 and #554 

## Compatibility

This is android only and therefore doesn't require any changes on ios or windows.

| OS      | Implemented |
| ------- | :---------: |
| iOS     |   ❌     |
| Android |    ✅     |
| Windows |   ❌     |

## Checklist

Imo there is no need for sample code or README changes. It might deserve a mention in the changelog though.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -3,8 +3,11 @@
     package="com.learnium.RNDeviceInfo">
 
   <application>
-    <receiver android:name="com.learnium.RNDeviceInfo.RNDeviceReceiver" android:enabled="true" android:exported="true"
-        tools:ignore="ExportedReceiver">
+    <receiver android:name="com.learnium.RNDeviceInfo.RNDeviceReceiver"
+        android:enabled="true"
+        android:exported="true"
+        tools:ignore="ExportedReceiver"
+        android:permission="android.permission.INSTALL_PACKAGES">
       <intent-filter>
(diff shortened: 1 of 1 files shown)


leoek:
Thanks for the fast merge!

mattijsf:
Thanks for the update. It might be good to also rename the receiver into something like `RNDeviceInstallReferrerReceiver` to emphasize that it's a single purpose receiver. Might create a PR for this...

PR #1072, opened 2020-09-03
[Windows] Fix vNext issue where module did not work on C++ app

tero-paananen:
Tested with C# and C++ app


mikehardy (approved):
I will trust you completely :-)

PR #1075, opened 2020-09-17
fix: Xcode 12 compatibility

radko93:
## Description

Latest Xcode 12 might fail to build when a library does not depend on `React-Core` directly hence this change is necessary for native modules on iOS. This change requires to React Native 0.60.2 or newer and it's not breaking change. For more details please check: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

I have not tested this library with Xcode 12 yet, but most native modules fail so I'm submitting this PR as it's thas not change anything besides changing minimum RN version to 0.60.2.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist


mikehardy (approved):
Thanks @radko93 !

PR #1079, opened 2020-09-23
[Documentation] getApiLevel returns -1 in iOS

abumalick:
## Description
Added `-1` as a return value for iOS for the `getAPILevel` function

The `?` in the documentation is a bit misleading. I thought I would get `null` or `undefined`. When I checked it was `-1`. I think it is good to document it in the API documentation if iOS has consistent return value for this function.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ Gets the API level.
 
 ```js
 DeviceInfo.getApiLevel().then(apiLevel => {
-  // iOS: ?
+  // iOS: -1
   // Android: 25
   // Windows: ?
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
I prefer not to do this as all APIs could be platform-specific, and the values returned for "unknown" are documented clearly at the top, -1 is what you get for an answer to an integer API when not supported on the platform

abumalick:
oh, I didn't see this. Thank you for clarification.

PR #1083, opened 2020-10-02
useManufacturer

schie:
Added `useManufacturer()` that wraps the async function, `getManufacturer()`, in a hook.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅ ❌     |

## Checklist

  * I wasn't able to confirm on a Windows box, but was able to do so on iOS and Android. I assume it would work because the core, `getManufacturer()`, is marked as being able to work on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1660,10 +1660,28 @@ const { loading, result } = useIsEmulator(); // { loading: true, result: false }
 <Text>{loading ? 'loading...' : result}</Text>;
 ```
 
+---
+
+### useManufacturer
+
+Gets the device manufacturer.
+
+#### Example
+
+```jsx
+import { useManufacturer } from 'react-native-device-info';
+
+const { loading, result } = useManufacturer(); // { loading: true, result: "Apple"}
+
+<Text>{loading ? 'loading...' : result}</Text>;
+```
+
 =======
+
 ## Native interoperatibily
 
 If you need to check for device type from the native side, you can use the following:
+
 ```java
 import com.learnium.resolver.DeviceTypeResolver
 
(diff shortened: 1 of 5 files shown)


mikehardy (approved):
This looks great! thank you

My only thought, and it's unrelated really, is that I almost wish we had a code generator! Every device-info piece of data should really be available sync, async, and via hook, it's a shame it is not easy to make that happen.

But for this change of course, it's a valid hook and a good implementation, thank you!

schie:
@mikehardy, great idea!

Off the top of your head, do you know if this lib follows the following motif through out?
- `RNDeviceInfo.getSystemManufacturer()` for asynchronous 
- `RNDeviceInfo.getSystemManufacturerSync()` for synchronous

If so, I think it can be done.

Regardless, I'll look into it.

mikehardy:
Yes! It follows that dogmatically when possible, with the possible exception of the constant values (not sure they have a matching pair since they are always sync ?) and a couple of APIs where sync is not possible or really not wanted (user agent I think? Not sure of others?). So if done there would probably need to be a small lookup table of APIs to not emit sync for or not emit async for.

schie:
Awesome, sounds good. I'll see what I can do.

schie:
Tried to crack this, but it's super complicated. 

Went down the following route:

**a function that returns an object with the asynchronous, synchronous, and hook that could be spread (...) out in the index file**

I gave up on this because it ended up being a ridiculous callback hell for handling custom logic (OS checking and such)



<hr/>

Since I can't find the silver bullet, my next plan is to make a convenience function to try to DRY-up existing functions with the pattern:

```js
if (Platform.OS == 'android' || Platform.OS == 'ios') {
// return this
}

// return default 
```

This may help with just the overall long-windedness of the index file.

PR #569, opened 2019-02-05
Fixes #568

andreifilip123:
## Description

Fixed issue #568 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [X] I have tested this on a device/simulator for each compatible OS
* [ ] I added the documentation in `README.md`.
* [ ] I mentioned this change in `CHANGELOG.md`.

PR #1084, opened 2020-10-04
adding useIsHeadphonesConnected

### LINKED ISSUE #1013: add listener for updates to DeviceInfo.isHeadphonesConnected()
# Feature Request

Add way to listen for changes to the **isHeadphonesConnected**() value. 

## Why it is needed

Currently, if the user plugs in/unplugs their wired headset or turns off/on their BT headset after the component has loaded, no event is fired. 

The value of **DeviceInfo.isHeadphonesConnected()** only updates on the next render of the component.

## Possible implementation

Ideally this would be a new hook.

### Code sample

```ts
const headphonesConnected = useIsHeadphonesConnected();
```

schie:
#569 
 Description

Fixes #1013

- Added hook, `useIsHeadphonesConnected()` , that tells if headphones are connected to the device. 
- Added a new event, `RNDeviceInfo_headphoneConnectionDidChange`,  that is fired by iOS and Android
- Added a convenience hook, `useOnEvent` , that
  - is used by `useIsHeadphonesConnected`
  - can be used to update existing event hooks to DRY them up
  - can help 'simplify' future event hooks

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |
| Windows |    ❌     |

## Checklist

### TL;DR

use-is-headphones-connected mov

### DIFF
--- a/README.md
+++ b/README.md
@@ -1676,6 +1676,24 @@ const { loading, result } = useManufacturer(); // { loading: true, result: "Appl
 <Text>{loading ? 'loading...' : result}</Text>;
 ```
 
+---
+
+### useIsHeadphonesConnected
+
+Tells if the device is connected to wired headset or bluetooth headphones.
+
+This hook subscribes to the event, `RNDeviceInfo_headphoneConnectionDidChange` , and updates the `result` field accordingly.
+
+#### Example
+
+```jsx
+import { useIsHeadphonesConnected } from 'react-native-device-info';
+
+const { loading, result } = useIsHeadphonesConnected(); // { loading: true, result: false}
+
+<Text>{loading ? 'loading...' : result}</Text>;
+```
+
 =======
 
 ## Native interoperatibily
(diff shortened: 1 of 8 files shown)


mikehardy (approved):
That's fantastic @schie - well-coded it seems, great atomic commits as well, makes my review very easy.

mikehardy:
v6.2.0 is out! and I tangentially discovered that the example app was still set up for RN0.63 even though we pinned it to RN0.62 when bringing in react-native-windows, so it wasn't working :facepalm: - fixed that, everything here checked out. Thanks

schie:
@mikehardy,
awesome! I had switch to react-native@0.63 in my local, but I didn't want to change anything because I have zero-experience w/ react-native-windows or anyway to confirm any fixes worked, so I left it be.

mikehardy:
Yeah, I'm not really up on RNWindows but I know that because facebook has a stance of only supporting android + iOS in core it has to be a fork / module and consequently lags a little bit. When we merged support they weren't ready for 0.63 yet and I'm not sure where they are now but it definitely needs to be 0.62 for us at the moment owing to that.

Issue #1085, opened 2020-10-05
DRYing up index code base


schie:
# Feature request
Reduce long-windedness and redundancy of the source code in the index file.

## Why it is needed
There is a recurring pattern in the source code that can be be made less verbose.
While this is not _needed_ per se, it is very tedious, long-winded, and redundant logic.
There is also a small discussion about this issue [here](https://github.com/react-native-community/react-native-device-info/pull/1083).

The below example represents the vast majority of the getters in the file.

```ts
let someMemoValue;
export async function getSomething() {
  if (Platform.OS == 'x' || Platform.OS == 'y' || ...) {
    someMemoValue = await RNDeviceInfo.getSomething();
  } else {
    someMemoValue = SOME_DEFAULT_VALUE;
  }
  return someMemoValue;
}

export function getSomethingSync() {
  if (Platform.OS == 'x' || Platform.OS == 'y' || ...) {
    someMemoValue = RNDeviceInfo.getSomethingSync();
  } else {
    someMemoValue = SOME_DEFAULT_VALUE;
  }
  return someMemoValue;
}

```

## Possible implementation

To somewhat solve this tedium and redundancy, I propose a convenience function that generates the Async and Sync functions.

### Code sample

The following snippet shows how the previous code sample could be rewritten w/out any breaking changes.

```ts
export const [getSomething, getSomethingSync] = getSupportedPlatformInfoFunctions({
  getter: () => RNDeviceInfo.getSomething(),
  syncGetter: () => RNDeviceInfo.getSomethingSync(),
  defaultValue: SOME_DEFAULT_VALUE,
  supportedPlatforms: ['android', 'ios', ...],
});
```

_Note: I went with passing in `getter` and `syncGetter` functions to_
- avoid having to pass around more strings
- avoid being overly terse
- avoid being, possibly, more error-prone

The function, `getSupportedPlatformInfoFunctions` , could also bet expanded to include the `useOnMount` hooks, but that's a bit out-of-scope for what I want to do, for meow. 

<hr />
<br/>
If we're cool with this implementation or with some minor tweaks, I'll make a PR soon that'll include a rewrite of one or two sets of the existing getters.
<br/>
<br/>
Thoughts?

mikehardy:
[...]

PR #1086, opened 2020-10-06
Adding getSupportedPlatformInfoFunctions

schie:
## Description

re #1085

- Added a helper/convenience function, `getSupportedPlatformInfoFunctions(...)` that gets rid of some of the repetitive logic around supported platforms and memoization
- Added helper functions, `getSupportedPlatformInfoSync` and `getSupportedPlatformInfoAsync` , that can be used to explicitly get info sync or async if one desires to do so

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

  - unable to verify on Windows, but did verify for iOS and Android

### Screen Shots
Below are screen shots of iOS and Android and the updated `getDeviceName*` functions (no difference)
image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -440,28 +440,13 @@ export function getBootloaderSync() {
   return bootloader;
 }
 
-let device: string;
-export async function getDevice() {
-  if (!device) {
-    if (Platform.OS === 'android') {
-      device = await RNDeviceInfo.getDevice();
-    } else {
-      device = 'unknown';
-    }
-  }
-  return device;
-}
-
-export function getDeviceSync() {
-  if (!device) {
-    if (Platform.OS === 'android') {
-      device = RNDeviceInfo.getDeviceSync();
-    } else {
-      device = 'unknown';
-    }
-  }
-  return device;
-}
+export const [getDevice, getDeviceSync] = getSupportedPlatformInfoFunctions({
+  getter: () => RNDeviceInfo.getDeviceName(),
+  syncGetter: () => RNDeviceInfo.getDeviceNameSync(),
+  defaultValue: 'unknown',
+  memoKey: 'device',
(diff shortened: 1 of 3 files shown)


schie:
The usage of this function can probably be used with a lookup table to make the index file more terse, but that's a future thing if desired — me thinks.

src/index.ts
```diff
@@ -440,28 +440,13 @@ export function getBootloaderSync() {
...
-      device = RNDeviceInfo.getDeviceSync();
-    } else {
-      device = 'unknown';
-    }
-  }
-  return device;
-}
+export const [getDevice, getDeviceSync] = getSupportedPlatformInfoFunctions({
+  getter: () => RNDeviceInfo.getDeviceName(),
+  syncGetter: () => RNDeviceInfo.getDeviceNameSync(),
```
mikehardy, line 445: I think you intend these native getters for this one?

```suggestion
  getter: () => RNDeviceInfo.getDevice(),
  syncGetter: () => RNDeviceInfo.getDeviceSync(),
```

mikehardy (changes requested):
This looks excellent to me - just one comment that appears to be a trivial think-o

src/index.ts
schie: good catch!

mikehardy (approved):
Excellent! I'll merge this but as it's just a refactor and not even a bugfix I'll not do a release, just FYI (no user benefit, seems like a net negative) but this will be a big improvement for development.

If it works for you and you're interested in doing it, normalizing the rest of the file like this would work fine by me :-) :pray:

schie:
Totally fine with me.

I'll knock out a few conversions later today.

PR #1102, opened 2020-10-17
adding mocks

### LINKED ISSUE #1101: Adding Jest Mocks
## Description

Through-out the various apps I work on, we always end up writing jest mocks for this lib.

Is there any opposition to adding file like `<root>/jest/react-native-device-info-mock.js` that can be imported into an app's jest setup ?

There are a few other libs that follow a similar approach:
- [react-native-netinfo](https://github.com/react-native-netinfo/react-native-netinfo/blob/master/jest/netinfo-mock.js)
- [react-native-async-storage](https://github.com/react-native-async-storage/async-storage/blob/master/jest/async-storage-mock.js)

## Use Case

As mentioned in various other issues related to mocks, this would greatly improve `Developer Experience` .

<hr />

If we're cool with it, I can make a PR adding it.

schie:
## Description

resolves #1101

Added a mock file simplifies the mocking/testing process for devs

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

  - tested on iOS and Android, but unable to do so on Windows.
  - sample usage in example jest files


mikehardy (approved):
fantastic, thank you!

schie:
NP!

htbkoo:
First of all, thanks a lot @schie for adding the mocks which is already very useful 👍 

A quick question - I can see that [there is a `TODO` reminder](https://github.com/react-native-device-info/react-native-device-info/pull/1102/files#diff-0b65441e97fba0d74e342347404bde1a89ad2fb20b5d9af94f84177aa2887249R136) about `hook mocks`, and I wonder if there are any suggestions or guidelines on mocking the hooks for now.

This is what I currently have at my `jest.setup.ts` file but I am not sure if this is the best way:
```
jest.mock("react-native-device-info", () => ({
  ...mockRNDeviceInfo,
  useBatteryLevel: mockRNDeviceInfo.getBatteryLevel, // only mocking `useBatteryLevel` because this is the only hook that I am using currently
}));
``` 

CC @mikehardy in case you would like to share your thoughts as well

schie:
Hi @htbkoo ,

Thanks for the kind words. 

For the hook you're using, yes, that should be fine. 

You should definitely make a PR adding that if you have time!

htbkoo:
Thanks a lot @schie for creating #1132 and apologizes I did not get time to create the PR to add the mocks in the last week......

PR #1107, opened 2020-10-20
add new devices: Pixel 3a and Redmi 7 with notch

tronin:
## Description

Added Redmi 7 and Pixel 3a support

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Release Notes
 
+### 7.0.1
+ * fix(android): fix missing Redmi 7 and Pixel 3a model name in devicesWithNotch.ts (@tronin)
+
 ### 7.0.0
 
  * BREAKING CHANGE: Increase iOS minimum to iOS 10. Stay on 6.x if you need to support iOS 9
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
looks good to me, thanks!

will do a release in a few minutes

systemride:
I'm curious why the Pixel 3a was added to the list? I have a Pixel 3a, and it does not have a notch.

tronin:
@systemride  Yes, but he has no frame and is behaving like a notch. Notch in this case applies to the the SafeArea activation. The bottom indent of the swipe up controller appears as well as the upper indent for the StatusBar.

tronin:
It may be that the naming is not quite correct, but hasNotch function must be activated for all phones with notch or without a frame.

systemride:
I see what you're saying. But yes I agree, perhaps it should be considered to change the name of the function.

mikehardy:
A name change would be a breaking change for everyone, that's pain for not much gain
A docs PR that explains more clearly the intent would be no pain and maybe gain though?

PR #1113, opened 2020-10-26
refactor(windows): port C# Windows implementation to C++

### LINKED ISSUE #1073: Port C# Windows implementation to C++
# Feature request

This PR recently added Windows support:
https://github.com/react-native-community/react-native-device-info/pull/1059

First I just wanted to say thanks so much for doing this!  It's great to see so much enthusiasm for Windows support in this space.

I have a suggestion here, to port this code over to C++.  See next section for rationale.

## Why it is needed

There are a couple reasons to consider porting the module to C++:

**Performance:**  
The core of react-native-windows is written in C++ and the standard template is a C++ app.  This is for performance reasons.  Bringing in a C# native module adds the overhead of the CLR to the app which is significant, and isn't strictly necessary.

**Issue https://github.com/microsoft/react-native-windows/issues/5858**:  
This currently prevents apps from using C# native modules from the standard C++ based template app, which is the default.  We are talking within Microsoft on how to address this, and that process will take time.  In the mean-time there are workarounds but they aren't easy for customers to understand, so we have a short to medium term problem, which could be avoided altogether if the module was converted to C++
## Possible implementation

The react-native-windows website has a guide on how to author a C++ native module.    
Start here:   
 https://microsoft.github.io/react-native-windows/docs/native-modules  
See section called "Sample Native Module (C++)"

### Code sample

For a working example of a C++ native module, see:
https://github.com/react-native-community/async-storage/tree/master/windows/ReactNativeAsyncStorage

jaimecbernardo:
## Description

Port Windows implementation from C# to C++.
Port Windows code in the example app to use a C++/winrt implementation.
Remove the C# implementation.
Add some Windows tests to the example app, and run those as the Windows CI, replacing old Windows test in GitHub workflows.

Many thanks to the authors of the C# implementation and to those who have been updating/maintaining it, since this work is based on their work.

Fixes #1073 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Windows |    ✅     |

## Checklist

* [X] I have tested this on a device/simulator for Windows, testing on react-native versions 0.61, 0.62 and 0.63.


windows/code/RNDeviceInfoCPP.h
```diff
@@ -0,0 +1,442 @@
+#pragma once
```
mikehardy, line 1: Am I strange for thinking a .h file should be a header file and the .cpp would normally contain the implementation 🤔  ?

mikehardy:
Oooh exciting :-), thank you @jaimecbernardo hopefully not too distasteful despite the Sync method signatures being objectively kind of awful (and yet removed at anyone's peril)...

I read through and other than a question about .h vs .cpp, going back to the related issue, I don't have a windows rig I trust to run this so may I politely ask that you confirm this appears to run fine for you?

You worked through the example as well so I'm assuming yes, but have to ask :-)

mikehardy:
Oh - one other item, we were pinned in the example on 0.62.x because of react-native-windows, it appears that the example could be bumped to 0.63.x ? https://github.com/react-native-device-info/react-native-device-info/blob/7f1d2a66acc2b828ec5b68330682ac6f89fd7227/example/package.json#L15

Related, I would ❤️  for this script to be usable again https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh - as a maintainer I feel that tracking current stable in the example is both important and the biggest waste of my time ever, but I can't refresh safely right now

windows/code/RNDeviceInfoCPP.h
jaimecbernardo: Yes, that'd be accurate for classes you'd like to include in order to keep the API in the header, without code.
In this case it can be either way, since the API won't be consumed directly by other C++ code, but rather compiled as a module.
I've chosen to follow this sample when choosing this: https://github.com/microsoft/react-native-windows-samples/blob/master/samples/NativeModuleSample/cppwinrt/windows/NativeModuleSample/DataMarshallingExamples.h

jaimecbernardo: I can divide it into a .h and .cpp if you prefer.

mikehardy: I don't feel that strongly about it - I feel documentation is frequently harder to do well than code, and a coordinated change across both is very difficult, so if there is existing documentation in the form of an example then I would lean towards conforming with it - TL;DR: - LGTM ;-)

jaimecbernardo:
Thank you for the review :)

> I read through and other than a question about .h vs .cpp, going back to the related issue, I don't have a windows rig I trust to run this so may I politely ask that you confirm this appears to run fine for you?

Yes, I've added a github workflow that builds and runs the example app on Windows and checks if each tab in the example app is correctly loaded. You can check the reasults in the PR's checks.

> Oh - one other item, we were pinned in the example on 0.62.x because of react-native-windows, it appears that the example could be bumped to 0.63.x ?

Indeed it could be bumped.

> Related, I would ❤️ for this script to be usable again https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh - as a maintainer I feel that tracking current stable in the example is both important and the biggest waste of my time ever, but I can't refresh safely right now

Interesting, and I've gone and added some more test code in there 😅 I'll take a look at the script and see what else needs to be copied around/ran. With auto-linking on Windows since 0.63, I suppose it'll be easier.
You have to run it in macOS, though, because of `pod install`. Is this accurate?

mikehardy:
Yeah, it is not going to be easy to make that work across all three platforms (win/mac/linux) that it runs on now - I primarily run it on macOS because of the pod install yes. It may be an intractable problem - I want full multi-platform support more than I want the script so I'm prepared to accept that but I will be sad. Updating examples is normally the same time it takes to merge 1-2 large PRs and I don't typically have a lot more time than that regularly

jaimecbernardo:
I'll give it a try at adding some instructions to the script pertaining to the Windows code.
I see there are some other changes that are not yet contemplated in the scripts, though, such as a non Windows test and an "app-extension" for iOS.

mikehardy:
Indeed - entropy takes hold quickly!

jaimecbernardo:
Hi @mikehardy 

I've updated `refresh-example.sh` with the windows specific commands needed.
I've also updated the example to `react-native 0.63.3`.

Regarding the "app-extension" for iOS, all I've done is manually cherry-pick this commit https://github.com/react-native-device-info/react-native-device-info/commit/2e8d47c9c863e1795089e93b188cd65f1e8f3886 and resolve the conflicts. The app still runs, but I have no idea what the "app-extension" is supposed to do.

After updating to `react-native 0.63.3`, I've tested it successfully on a Windows device, an Android device and an iOS emulator.

mikehardy:
Sweet! The app extension is just there to make sure getFontScale works in app extensions, it was a bug and recent pr

jaimecbernardo:
I suppose the app building and running for iOS is a sufficient test, then? :)

jaimecbernardo:
The test should've run successfully. It seems to have failed while getting a NuGet package.
I've re-run it in my repo to make sure it should run fine.
Could you please have the action re-run, @mikehardy ?

.github/workflows/windows-app-test.yml
```diff
@@ -0,0 +1,56 @@
...
+    - uses: actions/checkout@v2
+      name: Checkout Code
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.9.1'
+
+    - name: Setup MSBuild
+      uses: microsoft/setup-msbuild@v1.0.0
```
mikehardy, line 19: out of curiosity, why pin the patch version even?
```suggestion
      uses: microsoft/setup-msbuild@v1
```

should be sufficient and allow bugfixes?

mikehardy:
Thanks! Yes, if the iOS example app builds the extension should be fine, I believe that is sufficient

I re-ran the action and it appears to be reproducibly not working:

```
Build failed with message D:\a\react-native-device-info\react-native-device-info\example\node_modules\react-native-windows\ReactWindowsCore\ReactWindowsCore-Items.vcxitems(231,5): error : This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is D:\a\react-native-device-info\react-native-device-info\example\windows\packages\boost.1.68.0.0\build\boost.targets. [D:\a\react-native-device-info\react-native-device-info\example\node_modules\react-native-windows\ReactWindowsCore\ReactWindowsCore.vcxproj]. Check your build configuration.
```

I have literally zero experience with building windows items so even the stack overflow that comes up https://stackoverflow.com/questions/22909506/this-project-references-nuget-packages-that-are-missing-on-this-computer is mysterious but this appears to be semi-common and something in the csproj may have bearing

jaimecbernardo:
I've re-ran it on my repo: https://github.com/jaimecbernardo/react-native-device-info/pull/2/checks?check_run_id=1321362933
And it worked successfully.

So my theory is that the github action is saving some state between runs in the same PR. It has ran before when react-native was 0.62 . Now it's running after updating to 0.63. My guess is that it's trying to use the NuGet packages that are there from 0.62.

I'll look into it on my repo and try to replicate these conditions.
I'll add a commit to fix once I understand what's the issue here and am able to fix it.

exotexot:
Works for me on react-native 0.62 and react-native-windows 0.62.14. Thanks @jaimecbernardo !

mikehardy:
Nice, thanks for the success report @exotexot

.github/workflows/windows-app-test.yml
jaimecbernardo: Thank you. No issues here.
I've added the suggested change to the latest commit.

jaimecbernardo:
Hi @mikehardy ,

Ended up removing the caching of node_modules between CI runs, since that cache was big and contained build artifacts for the modules, which were causing the build failures on following runs.

It passes now 😄

.github/workflows/windows-app-test.yml
```diff
@@ -16,20 +16,11 @@ jobs:
...
+      uses: microsoft/setup-msbuild@v1
       with:
         vs-version: 16.5
 
-    - name: Check node modules cache
-      uses: actions/cache@v1
-      id: yarn-cache
-      with:
-        path: ./example/node_modules
-        key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
```
mikehardy: so strange this does not work, as the key should miss and thus not restore
I have good luck with this incantation:

https://github.com/invertase/react-native-firebase/blob/63b822cd404892a61403c1c4c10d91a8ec4f13cc/.github/workflows/tests_e2e.yml#L56-L65

I think the caching should be a net positive, not least for speed, but because it will by definition fetch from network less and my largest sources of CI flakiness in the repos I manage are transient network failures fetching remote resources needed to do the CI.

Because of that, may I ask you to try the way we do it on react-native-firebase to see if it works? I guess this won't be easy to reproduce though it seemed that it relies on a 0.62 artifact to be built and only then fails when building 0.63?

I'm prepared to accept this if you don't have a good / easy way to reproduce it to check, and will just re-investigate if it turns into a source of CI flakiness

jaimecbernardo: I think I see the original issue now. It's caching the `./example/node_modules` path but is using the `./yarn.lock` file instead of the `./example/yarn.lock` file.
Funny enough, I don't think this gains much speed. On previous runs I noticed that cache downloading and modules installing both took ~5minutes. The CI flakiness point is very valid though.
I'm going to split into two caches to solve the issue.

jaimecbernardo: Nevermind, there's time saving after all, from my tests.

jaimecbernardo:
I've added a commit so that the workflow uses two caches, one for `./node_modules` and another for `./example/node_modules`. This fixes the CI issue this PR was having, while still keeping a cache.

mikehardy:
Fantastic! I don't think I had anything else then, let me re-scan this fresh but I believe it is just on me to go "yep" and merge. Really fantastic stuff @jaimecbernardo ! I'm excited to have an up to date example again as a side benefit and (I think - I will test) my handy refresh script back to keep it up to date :)

mikehardy:
@jaimecbernardo with apologies, I'm about to head out for vacation for 1 week starting tomorrow, and I've been hard up against other deadlines (forward porting react-native-firebase to all the breaking changes released in concert with the Firebase Summit that just happened...). I'll have to shelve this till I got back but it is still my expectation that I just need to review and it's good to go. Thanks for your patience

jaimecbernardo:
Thank you for the heads-up. Have a nice vacation! 😄

jaimecbernardo:
Hi @mikehardy ,
Any news on this? 😄 
Would love to get it in so I can properly open https://github.com/react-native-device-info/react-native-device-info/pull/1122

mikehardy (approved):
Your patience has been magnificent - thanks for pinging me on this one. The work looks fantastic on re-read just now, I really appreciate the mostly-resurrection of the refresh-example script. I'll figure out how to get the app extension handled in that thing, definitely not in scope here. Let's do this

Issue #1108, opened 2020-10-20
Adding general tests


schie:
## Description
Scope: Adding tests of the TS code
- [x] setup jest
- [x] add tests for internal functions
  - [x] asyncHookWrappers.ts
  - [x] supported-platform-info.ts
- [ ] add lib tests
  - _'ensure that exported `getX` calls expected Native Module function'_
  - [x]  getUserAgent
  - [x] getUserAgentSync
  - [ ] getSystemName
  - [ ] getReadableVersion
  - [ ] hasNotch
  - [ ] getTotalDiskCapacityOld
  - [ ] getTotalDiskCapacityOldSync
  - [ ] getFreeDiskStorageOld
  - [ ] getFreeDiskStorageOldSync
  - [ ] syncUniqueId
  - [ ] hasSystemFeature
  - [ ] hasSystemFeatureSync
  - [x] getPowerState
  - [x] getAvailableLocationProviders
  - [ ]  hasGms
  - [ ] hasGmsSync
  - [ ]  hasHms
  - [ ] hasHmsSync
  - Hooks
    - [ ]  useBatteryLevel
    - [ ] useBatteryLevelIsLow
    - [ ] useDeviceName
    - [ ] useFirstInstallTime
    - [ ] useHasSystemFeature
    - [ ] useIsEmulator
    - [ ] usePowerState
    - [ ] useManufacturer
    - [ ] useIsHeadphonesConnected
- [x] run tests through GitHub Action

## Use Case

Adding the above should help the review process.

**_note_**: after this is complete, I plan to work on Native Code tests, but I might need some guidance on that.

schie:
@mikehardy,

Any experience testing `NativeEventEmitter` ? 

I'm having trouble testing emitting events in jest, so I'm a little stuck.

 If you have time, I can push up what I have so far to take a look in a bit

mikehardy:
I'm sorry, I do not, I'm pretty much mocked to empty stubs for RNDI in my work project

schie:
Gah! I was hoping others might help. Writing tests are the worst!

I'll be working on this a bit more soon.

PR #1116, opened 2020-11-01
more tests and a memoization bug fix

### LINKED ISSUE #1117: falsy memoization bug
## Summary

|             |     |
| ----------- | --- |
| Version     | 6.2.1+   |
| Affected OS | YES   |
| OS Version  | ?   |

####  `getSupportedPlatformInfoSync`
https://github.com/react-native-device-info/react-native-device-info/blob/5003f1e36fa2625939029a65decaad65c7831c3c/src/internal/supported-platform-info.ts#L50

#### `getSupportedPlatformInfoAsync`
https://github.com/react-native-device-info/react-native-device-info/blob/5003f1e36fa2625939029a65decaad65c7831c3c/src/internal/supported-platform-info.ts#L71

This bug seems to have been there long before the introduction of `getSupportedPlatformInfoFunctions` , but it has now been consolidated to that function after the refactoring.

## Current behavior

Right now, the functions used by `getSupportedPlatformInfoFunctions` perform a falsy-check on `memo[memoKey]` instead of checking if it's undefined. Since falsy values like, `false`, `0`, `null`, and `''`, are _valid_, they should prevent subsequent getters from being called.

## Expected behavior

if a getter responds back with a falsy value (other than `undefined`) , it should prevent subsequent getters to be called in favor of the memoized value.

schie:
## Description

re #1108
resolves  #1117

adds tests for the following groups of getters
- non-memoized boolean
- non-memoized number
- memoized array
- memoized boolean
- memoized number

fixes internal mock
- moves `getApiLevel` to `numberFnNames` (jest.setup.ts)

fixes a memoization bug related to falsy values in `getSupportedPlatformInfoFunctions`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

<hr />

The `non-memoized string` are a bit more complicated ('tedious' is probably a better word), so I'll have that in a later PR, or hopefully, others can take on some. :)

### DIFF
--- a/__tests__/getters.test.ts
+++ b/__tests__/getters.test.ts
@@ -99,6 +99,13 @@ const memoizedNumberGetters = [
   'getTotalMemory',
   'getMaxMemory',
 ].map(makeTable);
+const nonMemoizedNumberGetters = [
+  'getUsedMemory',
+  'getFontScale',
+  'getFreeDiskStorage',
+  'getBatteryLevel',
+  'getTotalDiskCapacity',
+].map(makeTable);
 
 describe('number getters', () => {
   describe.each(memoizedNumberGetters)(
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
The things you find when actually testing... :facepalm:

schie:
_'ha... yeah...'_ , he says in a self-deprecating melancholy tone

PR #1128, opened 2020-11-16
Add isCatalyst

### LINKED ISSUE #1127: In Catalyst, app using getSystemName() reports as "iOS" on Big Sur
In Catalyst, app using getSystemName() reports as "iOS" on Big Sur. Probably be an easy fix though.

marcosrdz:
## Description

Fixes #1127

Added `isCatalyst()` that allows the developer to know whether the app is running on iOS or Mac Catalyst

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,7 @@ Most APIs return a Promise but also have a corresponding API with `Sync` on the 
 | [hasSystemFeature()](#hassystemfeaturefeature)                    | `Promise<boolean>`  |  ❌  |   ✅    |   ❌    | ❌ |
 | [isAirplaneMode()](#isairplanemode)                               | `Promise<boolean>`  |  ❌  |   ✅    |   ❌    | ✅ |
 | [isBatteryCharging()](#isbatterycharging)                         | `Promise<boolean>`  |  ✅  |   ✅    |   ❌    | ✅ |
+| [isCatalyst()](#iscatalyst)                                       | `boolean`           |  ✅  |   ❌    |   ❌    | ❌ |
 | [isEmulator()](#isemulator)                                       | `Promise<boolean>`  |  ✅  |   ✅    |   ✅    | ❌ |
 | [isLandscape()](#isLandscape)                                     | `Promise<boolean>`  |  ✅  |   ✅    |   ✅    | ❌ |
 | [isLocationEnabled()](#isLocationEnabled)                         | `Promise<boolean>`  |  ✅  |   ✅    |   ❌    | ✅ |
@@ -1308,6 +1309,18 @@ Tells if the battery is currently charging.
 DeviceInfo.isBatteryCharging().then(isCharging => {
   // true or false
 });
+```
+
+---
+
+### isCatalyst()
+
+Tells if the application is running on iOS or Mac Catalyst.
+
+#### Examples
+
+```js
+let isCatalyst = DeviceInfo.isCatalyst();
 ```
 
 ---
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -66,6 +66,7 @@ RCT_EXPORT_MODULE();
          @"brand": @"Apple",
          @"model": [self getModel],
          @"deviceType": [self getDeviceTypeName],
+         @"isCatalyst": @([self isCatalyst]),
      };
 }
 
@@ -508,6 +509,15 @@ RCT_EXPORT_METHOD(getFreeDiskStorage:(RCTPromiseResolveBlock)resolve rejecter:(R
 - (NSString *) getDeviceTypeName {
     return [DeviceTypeValues objectAtIndex: [self getDeviceType]];
 }
+
+- (BOOL) isCatalyst {
+#if TARGET_OS_MACCATALYST
+    return YES;
+#else
+    return NO;
+#endif
+}
+
 
 - (NSArray *) getSupportedAbis {
     /* https://stackoverflow.com/questions/19859388/how-can-i-get-the-ios-device-cpu-architecture-in-runtime */
--- a/src/index.ts
+++ b/src/index.ts
@@ -371,6 +371,13 @@ export const [isEmulator, isEmulatorSync] = getSupportedPlatformInfoFunctions({
   defaultValue: false,
 });
 
+export const isCatalyst = getSupportedPlatformInfoSync({
+  defaultValue: false,
+  memoKey: 'catalyst',
+  supportedPlatforms: ['ios', 'macos', 'windows'],
+  getter: () => RNDeviceInfo.isCatalyst(),
+});
+
 export const isTablet = () =>
   getSupportedPlatformInfoSync({
     defaultValue: false,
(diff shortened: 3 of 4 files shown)


mikehardy:
Hi there! I certainly like the idea of an API that signals we are working on catalyst, but I am not sure if this is the way.

Here is my doubt: should https://github.com/react-native-device-info/react-native-device-info#getsystemname actually be where this is done, and return macOS ? (and also perhaps tvOS, iPadOS, depending) ? I think that could work

In fact, checking the Flutter device_info code, we do the equivalent thing: 
- react-native: https://github.com/react-native-device-info/react-native-device-info/blob/55108da452fbe17231da968d2016011ad402d8ae/ios/RNDeviceInfo/RNDeviceInfo.m#L60
- flutter: https://github.com/flutter/plugins/blob/06530f9d3da4e80a818c3d3c0ec34fe2084b4abf/packages/device_info/device_info/ios/Classes/FLTDeviceInfoPlugin.m#L25

What I do not want to see is isCatalyst, isTvOS etc for all the fracturing Apple types, so even if getSystemName is not the place to do it exactly, I would want a single API like getSystemName


Perhaps their extra information chunk provides the information:
https://github.com/flutter/plugins/blob/06530f9d3da4e80a818c3d3c0ec34fe2084b4abf/packages/device_info/device_info/ios/Classes/FLTDeviceInfoPlugin.m#L31-L37

marcosrdz:
> Hi there! I certainly like the idea of an API that signals we are working on catalyst, but I am not sure if this is the way.
> 
> 
> 
> Here is my doubt: should https://github.com/react-native-device-info/react-native-device-info#getsystemname actually be where this is done, and return macOS ? (and also perhaps tvOS, iPadOS, depending) ? I think that could work
> 
> 
> 
> In fact, checking the Flutter device_info code, we do the equivalent thing: 
> 
> - react-native: https://github.com/react-native-device-info/react-native-device-info/blob/55108da452fbe17231da968d2016011ad402d8ae/ios/RNDeviceInfo/RNDeviceInfo.m#L60
> 
> - flutter: https://github.com/flutter/plugins/blob/06530f9d3da4e80a818c3d3c0ec34fe2084b4abf/packages/device_info/device_info/ios/Classes/FLTDeviceInfoPlugin.m#L25
> 
> 
> 
> What I do not want to see is isCatalyst, isTvOS etc for all the fracturing Apple types, so even if getSystemName is not the place to do it exactly, I would want a single API like getSystemName
> 
> 
> 
> 
> 
> Perhaps their extra information chunk provides the information:
> 
> https://github.com/flutter/plugins/blob/06530f9d3da4e80a818c3d3c0ec34fe2084b4abf/packages/device_info/device_info/ios/Classes/FLTDeviceInfoPlugin.m#L31-L37

This was the way we could determine if the app was running on iOS or macOS. On Catalina, it returns "Mac OS X", but on Big Sur it returns "iOS". Which means we can't tell where the app is running if the app is on Big Sur.

mikehardy:
Your direct experience is invaluable, thanks for providing the extra details.

I think the normal word for these sorts of things is "flavor"
perhaps a getSystemFlavor ? and it could start by just returning "Catalyst" for catalyst builds but could also hold the rest of the fragments (iPadOS, tvOS, watchOS, etc) if that became a thing?

I don't want to have a bunch of independent system flags though

So - a simple name change to that and I'd merge and release today? Would that work for you?

marcosrdz:
My concern this mixing those up is future-proof-ability. What if Catalyst extends to other Apple platforms? refactoring to separate tvOS from getSystemFlavor would be required and so on.

mikehardy:
This is sort of a general cross platform problem. It's hard to guess the future but it's possible with a general interface to be more precise in what "flavor" returns. Perhaps it could be "Catalyst-implantOS" or "Catalyst-vrOS" or whatever if that happens?

exotexot:
> What I do not want to see is isCatalyst, isTvOS etc for all the fracturing Apple types, so even if getSystemName is not the place to do it exactly, I would want a single API like getSystemName

I think a isCatalyst flag could be useful as it is more of a special edge case in apples ecosystem, but than again a popular one in the pre Apple Silicon era.

> This was the way we could determine if the app was running on iOS or macOS. On Catalina, it returns "Mac OS X", but on Big Sur it returns "iOS". Which means we can't tell where the app is running if the app is on Big Sur.

I also want to mention that getSystemVersion reports 14.2 (iOS) instead of (11.0) on BigSur. To detect BigSur than could be achieved via isCatalyst && getSystemVersion() > 11.

mikehardy:
It seems there are a bunch of special edge cases for the Apple ecosystem though, and as mentioned I do not want a bunch of flags for them. Your detection hypothesis for detecting BigSur is useful but would be even better if the module actually did it instead of everyone having to implement it, and it could be behind a specific flavor response as well

I really do not want a bunch of flags, is there a way to make this work with a more general purpose getSystemFlavor type API?

marcosrdz:
> > What I do not want to see is isCatalyst, isTvOS etc for all the fracturing Apple types, so even if getSystemName is not the place to do it exactly, I would want a single API like getSystemName
> 
> 
> 
> I think a isCatalyst flag could be useful as it is more of a special edge case in apples ecosystem, but than again a popular one in the pre Apple Silicon era.
> 
> 
> 
> > This was the way we could determine if the app was running on iOS or macOS. On Catalina, it returns "Mac OS X", but on Big Sur it returns "iOS". Which means we can't tell where the app is running if the app is on Big Sur.
> 
> 
> 
> I also want to mention that getSystemVersion reports 14.2 (iOS) instead of (11.0) on BigSur. To detect BigSur than could be achieved via isCatalyst && getSystemVersion() > 11.

I find isCatalyst a good approach given that its a special case that will definitely grow in the platform. Having "BigSur" if isCatalyst and > 11 will lead to the need to refactoring (especially after every release...also what about 11.1? Will it also say BigSur instead of whatever name they give it?). This approach wont work if the app isnt running with catalyst (rn macos for example).

mikehardy (changes requested):
Okay, but please add it to the mocks, to the example app, I think one of the internal/PrivateTypes needs to be isCatalystSync ? (even though it's the same, yes), and ideally the changelog, then we can merge this and release

marcosrdz:
> Okay, but please add it to the mocks, to the example app, I think one of the internal/PrivateTypes needs to be isCatalystSync ? (even though it's the same, yes), and ideally the changelog, then we can merge this and release

Will do. I'll aim to get that ready tonight.

exotexot:
Alternatively, one could extended https://github.com/react-native-device-info/react-native-device-info#getDeviceType

Maybe it could also return "desktop" on windows and catalyst

mikehardy:
I would be a lot happier with that - interesting thought @exotexot - and I think it conveys the information we need

mikehardy:
I appreciate the work here and the collaboration - I went with a merge on adding "desktop" to getDeviceType as a preference to adding this specific flavor. I *believe* that will still allow the precision people want (ios + desktop? that's catalyst) and I *hope* it is more flexible then just a catalyst specific flag, but the discussion here was where we figured that out and that was valuable. We can revisit later if this decision turns out to be in error, as mentioned I hope "desktop" as a device type is flexible while still being specific enough but in a year+ we might decide it was bad choice.

it'll be in the next release shortly

Thanks!

PR #1129, opened 2020-11-17
Added hasGms and hasHms functions

craig95:
## Description

Added `hasGms()` and `hasHms()` that tells you if the Android device supports Google Mobile Services or Huawei Managed Services.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Release Notes
 
+### Next
+ * feat(hasHms): added `hasHms` to indicate if Android device supports Huawei Mobile Services.
+ * feat(hasGms): added `hasGms` to indicate if Android device supports Google Mobile Services.
+
 ### 7.0.2
  * fix(windows): remove use of Object.fromEntries (#1110, thanks @jaimecbernardo!)
 
--- a/README.md
+++ b/README.md
@@ -366,6 +366,8 @@ Most APIs return a Promise but also have a corresponding API with `Sync` on the 
 | [getUsedMemory()](#getusedmemory)                                 | `Promise<number>`   |  ✅  |   ✅    |   ❌    | ✅ |
 | [getUserAgent()](#getuseragent)                                   | `Promise<string>`   |  ✅  |   ✅    |   ❌    | ✅ |
 | [getVersion()](#getversion)                                       | `string`            |  ✅  |   ✅    |   ✅    | ❌ |
+| [hasGms()](#hasGms)                                               | `boolean`           |  ❌  |   ✅    |   ❌    | ❌ |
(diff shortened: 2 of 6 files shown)


mikehardy (changes requested):
Interesting - I would have expected this to be available as a key in getSystemAvailableFeatures but I don't see it - so there's not any other way to get this information.

This looks like the start of Android system fragmentation though, 1 type is a flag, 2 types is a trend I fear, avoid proliferation could you reshape the API to "getManagedServicesFlavor" returning undefined / "HMS" / "GMS" ? 

The implementation looks great though - we already have the GMS build dep and you've avoided the HMS dependency nicely, which I appreciate

craig95:
I also checked `getSystemAvailableFeatures` and was also surprised not to find any indicators for GMS or HMS. 

I did consider a getManagedServicesFlavor type implementation for the same reasons you mentioned. The reason I decided to split it for each flavour was:

1. It is a simpler API to use. If getManagedServicesFlavor returns an array of strings you will need to loop through the array to check if the flavour you are interested in is present. 
2. I also considered the Android system fragmentation argument but in my opinion this won't be common. Huawei was forced by the US government to find an alternative to GMS, and building an alternative is no easy task. Huawei is the second largest smart phone producer in the world and in my opinion if other companies are also forced away from GMS they will first consider using HMS. 

I am not opposed to a getManagedServicesFlavor but for the reasons mentioned above I decided on the `hasGms` and `hasHms` versions. 

So if you would prefer a getManagedServicesFlavor type function I'll refactor it to that?

mikehardy (approved):
Mmm as I was typing a response that included a train of thought where Xioami also does MMS (MIUI Managed Services) I realized the next step of fragmentation would actually be multiple vendor ecosystems installed on the same device. And that this would already be desired by users of HMS (as they'd probably like GMS if they could get it, although I understand the nationalistic implications of the whole thing - I've been following it as a react-native-firebase maintainer, since it relies on GMS)

That implies that it would be an array as you mention, not the simple strings I mentioned, at which point it is an irritating API to consume.

So I've convinced myself alongside your convincing. This LGTM, at least until some 3rd vendor decides to go for the data+power-grab and build their own, then an array

Extra bonus points for getting the example and docs and changelog all in the PR, that helps me a lot

mikehardy:
Windows CI check is bitrot from GitHub's change to environment files, and it's in a GitHub Action even, while we are working separately on the Windows port, so is unrelated + acceptable

craig95:
Awesome, thanks @mikehardy.

mikehardy:
published as 7.1.0 - thanks!

schie:
@mikehardy 

no jest tests!?! I'm hurt...

💔

lol

PR #1131, opened 2020-11-18
1119: Change keychain access control

### LINKED ISSUE #1119: Weak Keychain Protection Class kSecAttrAccessibleAlways
This package uses [`kSecAttrAccessibleAlways` keychain security attribute](https://github.com/react-native-device-info/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L94) which is [deprecated in iOS 9](https://developer.apple.com/documentation/security/ksecattraccessiblealways) and not recommended for application use. 

Essentially this is a security risk which allows data in the keychain be accessible regardless if a device is locked.

dmitry-bluescape:
## Description

Fixes #1119

Could you help me test it?

### DIFF
--- a/ios/RNDeviceInfo/DeviceUID.m
+++ b/ios/RNDeviceInfo/DeviceUID.m
@@ -91,7 +91,7 @@ NSString * const UIDKey = @"deviceUID";
 + (NSMutableDictionary *)keychainItemForKey:(NSString *)key service:(NSString *)service {
     NSMutableDictionary *keychainItem = [[NSMutableDictionary alloc] init];
     keychainItem[(__bridge id)kSecClass] = (__bridge id)kSecClassGenericPassword;
-    keychainItem[(__bridge id)kSecAttrAccessible] = (__bridge id)kSecAttrAccessibleAlways;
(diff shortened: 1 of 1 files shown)


dmitry-bluescape:
The change is for iOS only, I'm not sure why windows tests fail

mikehardy:
No worries on the windows CI failure, we have a separate MASSIVE batch of windows changes pending that I need to merge, and it will take care of that - I'll put this in the review queue - thanks for taking the time to make a PR !

mikehardy (approved):
Yeah - combining your comment explaining the history and the options, I think this is best - looks good to me, I will test and do a release as soon as I can, in the meantime https://github.com/ds300/patch-package can be used if needed

weiningtry7:
change kSecAttrAccessibleAlways to kSecAttrAccessibleAfterFirstUnlock, the (OSStatus) status = -25300 it cause that uninstall the app, thd uid is always changed，because keychain can not access.

mikehardy:
@weiningtry7 do you have a PR you could propose? we merge PRs pretty well here when they're explained and the author tests them etc

PR #1133, opened 2020-11-26
Update build.gradle

Dimon70007:
update to use latest firebase version with bom

## Description

Fixes #<issue-number>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -63,6 +63,6 @@ dependencies {
       implementation "com.google.android.gms:play-services-iid:$iidVersion"
   }
 
-  testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.2'
-  testImplementation "org.mockito:mockito-core:3.3.3"
+  testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
+  testImplementation "org.mockito:mockito-core:3.6.44"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,8 +7,10 @@ buildscript {
             google()
             jcenter()
         }
+        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '3.5.1'
+
         dependencies {
-            classpath 'com.android.tools.build:gradle:3.6.3'
+            classpath "com.android.tools.build:gradle:$buildGradleVersion"
         }
     }
 }
@@ -20,7 +22,7 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
-    compileSdkVersion safeExtGet('compileSdkVersion', 28)
+    compileSdkVersion safeExtGet('compileSdkVersion', 29)
 
     defaultConfig {
         minSdkVersion safeExtGet('minSdkVersion', 16)
@@ -51,7 +53,8 @@ dependencies {
     implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1.2')}"
     def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
     if(firebaseIidVersion){
-        implementation "com.google.firebase:firebase-iid:$firebaseIidVersion"
+        implementation platform("com.google.firebase:firebase-bom:${safeExtGet("firebaseBomVersion", "25.12.2")}")
+        implementation "com.google.firebase:firebase-iid:${safeExtGet("firebaseIidVersion", "20.2.3")}"
     }else{
         def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
         implementation "com.google.android.gms:play-services-iid:$iidVersion"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -49,17 +49,20 @@ repositories {
 }
 
 dependencies {
-    implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1.2')}"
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -20,7 +22,7 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
-    compileSdkVersion safeExtGet('compileSdkVersion', 28)
+    compileSdkVersion safeExtGet('compileSdkVersion', 29)
```
mikehardy, line 25:
```suggestion
    compileSdkVersion safeExtGet('compileSdkVersion', 30)
```

we should be able to compile against 30 now?

android/build.gradle
```diff
@@ -7,8 +7,10 @@ buildscript {
             google()
             jcenter()
         }
+        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '3.5.1'
```
mikehardy, line 10:
```suggestion
        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '4.1.2'
```

should be current if we are changing it

android/build.gradle
```diff
@@ -51,7 +53,8 @@ dependencies {
     implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1.2')}"
     def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
     if(firebaseIidVersion){
-        implementation "com.google.firebase:firebase-iid:$firebaseIidVersion"
+        implementation platform("com.google.firebase:firebase-bom:${safeExtGet("firebaseBomVersion", "25.12.2")}")
+        implementation "com.google.firebase:firebase-iid:${safeExtGet("firebaseIidVersion", "20.2.3")}"
     }else{
         def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
         implementation "com.google.android.gms:play-services-iid:$iidVersion"
```
mikehardy, line 60: This part won't work well. The IID library was dropped from the BoM for a while, so depending on what BoM you are using for other reasons (e.g., react-native-firebase) you may need to specify a BoM that does *not* have an IID version, at which point we really need to use the firebaseIidVersion the user specifies.

Here you are saying "do you have a firebaseIidVersion?" then doing BoM work when you have it

The point of the BoM is that *if* you have it, then you don't put a version for the IID dependency at all, so this does not really help anything, it just gives us new external gradle config surface area (the firebaseBoMVersion) that we aren't using, and can't use very well

https://github.com/firebase/firebase-android-sdk/issues/1077#issuecomment-657011199

mikehardy (changes requested):
Hi there!
I think parts of this can work, but I find the BoM part problematic for the reasons I note in the associated comment

android/build.gradle
Dimon70007: yes, if this configuration works with defaults in your project

android/build.gradle
Dimon70007: applies any version, because user have ability to specify his buildGradlePluginVersion if he wants

android/build.gradle
Dimon70007: in my project with many libraries the firebase-bom does not works and makes dependencies conflict. that's why i have chosen specifying firebaseIidVersion. Anyway, i did not try to specify firebaseIidVersion with firebaseBomVersion together

Dimon70007: for example i have react-native-push-notification, react-native-firebase (crashlytics and messaging) and react-native-device-info and all this libs depends on firebase IID

Dimon70007:
Hi @mikehardy . I added firebaseBomVersion as standalone option from firebaseIIdVersion

android/build.gradle
```diff
@@ -7,8 +7,10 @@ buildscript {
             google()
             jcenter()
         }
+        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '4.1.2'
```
mikehardy, line 10:
```suggestion
        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '4.1.1'
```

This was my mistake, current version is still 4.1.1, there is not (and probably will not be?) a 4.1.2 - I'll suggest this but then also commit it so this can move forward, sorry

android/build.gradle
```diff
@@ -47,16 +49,20 @@ repositories {
...
+      implementation "com.google.firebase:firebase-iid:${firebaseIidVersion}"
+  }else{
+      def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
+      implementation "com.google.android.gms:play-services-iid:$iidVersion"
+  }
 
-    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.2'
-    testImplementation "org.mockito:mockito-core:3.3.3"
+  testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.2'
+  testImplementation "org.mockito:mockito-core:3.3.3"
```
mikehardy, line 67:
```suggestion
  testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
  testImplementation "org.mockito:mockito-core:3.6.44"
```

bumping these to current version - same as gradle plugin I'll suggest these in github web ui but then commit so this may move forward

mikehardy (approved):
Looks good to me - thanks for working through it with me
I'll give it a good testing and adjust as needed but should be fine
Working towards a release today

Dimon70007:
Happy to help.

PR #1137, opened 2020-12-04
Added DeviceTypeDesktop to getDeviceType()

exotexot:
As discussed here: https://github.com/react-native-device-info/react-native-device-info/issues/1127 and here: https://github.com/react-native-device-info/react-native-device-info/pull/1128

I've extended getDeviceType() to return "Desktop" on Catalyst and Windows apps. I'm not sure my solution for the windows version is desired, but it is tested and works.


mikehardy (approved):
I like these - and I think catalyst is getting closer to standard usage, just ran up against it in another project. Still not working out of the box on react-native core which is why most people probably haven't noticed this yet. Either way, with this device info will be ready. Thanks!

mikehardy:
v7.4.0 is out just now with this, thanks!

PR #1140, opened 2020-12-09
Redmi 9 has notch

esgudnason:
## Description

Fixes Redmi 9 has notch

## Checklist


mikehardy (approved):
Thanks! No way I would have gotten that model number correct since I don't have the device ;-). I'll do a release shortly

SarvaDev-123:
but right now It's not detected Redmi 9 has notch

mikehardy:
@SarvaDev-123 if you discover there is a problem, and post a PR to solve it, I'll happily merge it. Cheers

Issue #1149, opened 2021-01-07
Support building without com.android.installreferrer


saghul:
## Description

Would it be acceptable to support building this library without com.android.installreferrer support? I can work on a PR if that is ok.

## Use Case

It's not open source and thus any app using this library cannot be released on F-Droid.

mikehardy:
Very good point, and I'm an F-Droid fan, I'd love to do anything necessary to make sure we support it.
I just merged a change related to install referrer (removing the previous / deprecated style) so that area of code should be free+clear for development, I'd welcome a PR

Thanks!

saghul:
Very nice to hear! I'll try to work on a PR for this tomorrow.

saghul:
✨  https://github.com/react-native-device-info/react-native-device-info/pull/1153 ✨

PR #1153, opened 2021-01-08
feat: Add ability to run without com.android.installreferrer

saghul:
Fixes:
https://github.com/react-native-device-info/react-native-device-info/issues/1149

## Description

Fixes #<issue-number>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
@@ -2,84 +2,123 @@ package com.learnium.RNDeviceInfo;
 
 import android.content.SharedPreferences;
 import android.content.Context;
-import android.os.RemoteException;
 import android.util.Log;
 
-import com.android.installreferrer.api.InstallReferrerClient;
-import com.android.installreferrer.api.InstallReferrerStateListener;
-import com.android.installreferrer.api.ReferrerDetails;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+
 
 public class RNInstallReferrerClient {
+  private static Class<?> InstallReferrerClientClazz;
+  private static Class<?> InstallReferrerStateListenerClazz;
+  private static Class<?> ReferrerDetailsClazz;
 
-  private SharedPreferences sharedPreferences;
-  private InstallReferrerClient mReferrerClient;
+  static {
+    try {
+      InstallReferrerClientClazz = Class.forName("com.android.installreferrer.api.InstallReferrerClient");
+      InstallReferrerStateListenerClazz = Class.forName("com.android.installreferrer.api.InstallReferrerStateListener");
+      ReferrerDetailsClazz = Class.forName("com.android.installreferrer.api.ReferrerDetails");
+    } catch (Exception e) {
+      System.err.println("RNInstallReferrerClient exception. 'installreferrer' APIs are unavailable.");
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -2,84 +2,123 @@
...
+  // From InstallReferrerClient.InstallReferrerResponse
+  private static final int R_RESPONSE_OK                    = 0;
+  private static final int R_RESPONSE_SERVICE_UNAVAILABLE   = 1;
+  private static final int R_RESPONSE_FEATURE_NOT_SUPPORTED = 2;
+
+  RNInstallReferrerClient(Context context) {
+    sharedPreferences = context.getSharedPreferences("react-native-device-info", Context.MODE_PRIVATE);
+
+    if (InstallReferrerClientClazz == null || InstallReferrerStateListenerClazz == null || ReferrerDetailsClazz == null) {
+      return;
```
mikehardy, line 40: Just in case, could you throw in an info log explaining "The install referrer library was not found, so install referrer functionality will not be available" in here? silently bailing out of documented functionality is the way to cut this functionality out, but it might violate principle of least surprise for some un-wary dev

saghul: That log line is already done in the static block which sets those.

saghul: Line 23

mikehardy (changes requested):
one request for a log message, otherwise I think this is good to go

I checked the build.gradle and https://github.com/react-native-device-info/react-native-device-info/blob/9f286d724be1041ec4816c677ff2fff713115f07/android/build.gradle#L53 indicates implementation dependency so I don't think it needs modification in order to clean the app up for F-Droid inclusion?

mikehardy (approved):
Oh - we crossed paths :-) - I see what you mean, line 23 log is sufficient

mikehardy:
v7.4.0 is out just now with this, thanks!

saghul:
😱😍 That was fast, cheers!

mikehardy:
No reason to sit on stuff that's done, right? Thanks for the PR, happy to enable pure FOSS stuff whenever possible, same reason I work on AnkiDroid :)

saghul:
For posterity, here is how we make the"libre" flavor of Jitsi Meet which is published on F-Droid:

```
        implementation(project(':react-native-device-info')) {
            exclude group: 'com.google.firebase'
            exclude group: 'com.google.android.gms'
            exclude group: 'com.android.installreferrer'
        }
```

android/src/main/java/com/learnium/RNDeviceInfo/RNInstallReferrerClient.java
```diff
@@ -2,84 +2,123 @@
...
-        Log.d("RNInstallReferrerClient", "InstallReferrerService disconnected");
+      return null;
+    }
+
+    public void onInstallReferrerSetupFinished(int responseCode) {
+      switch (responseCode) {
+        case R_RESPONSE_OK:
+          // Connection established
+          try {
+            //if (BuildConfig.DEBUG)
```
tmoldovan8x8, line 87: Should we remove these ifs?

PR #1159, opened 2021-01-11
hasGms & hasHms doc fix

magrinj:
## Description

Small doc fix for `hasGms` and `hasHms` functions


mikehardy (approved):
Indeed I think your change is more accurate, thank you!

PR #1163, opened 2021-01-20
Update Gradle Wrapper to 6.5

friederbluemle:
Small Gradle Wrapper update from 6.3 to 6.5.

I saw the following error in the `android/` folder:

```
Minimum supported Gradle version is 6.5. Current version is 6.3.
```


mikehardy:
why not 6.8? (it's current, works for my project FWIW)

friederbluemle:
Sure, I updated it to 6.8 :+1:
The reason why I went with 6.5 is simply because that's the version used with the Android Gradle plugin/Android Studio 4.1.1. Unless there is a specific reason/benefit to further update Gradle Wrapper, I usually don't bother and just keep these versions aligned with AGP for all my Android projects.

mikehardy (approved):

mikehardy:
Haha but you know me already. I'm usually uninterested in alignment and more interested in discovering (early if possible) if there is a problem with integration of "current stable" whatever that is. My stuff is typically unaligned but is "current stable". At least for this one (unlike my frequent "oops I took react past react-native's react dep recommendation) doesn't have a reason behind it so we shouldn't be in trouble

Cheers @friederbluemle

PR #1165, opened 2021-01-21
add new calcul freeDiskStorage ios 11+

### LINKED ISSUE #1148: getFreeDiskStorage invalide value
## Summary
real device ios 14.3 iphone 11pro 
|             |     |
| ----------- | --- |
| Version     |  7.3.1   | 
| Affected OS | ios  |
| OS Version  |  14.3   |

## Current behavior

Value returned if not valid. Systeme os return 23go free space or, method getFreeDiskStorage return 6.9 go
## Expected behavior

i Would like return the real system free disk storage

gh-pro:
## Description

Fixes #1148

Update `getFreeDiskStorage()` that allows ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -488,15 +488,28 @@ RCT_EXPORT_METHOD(getTotalDiskCapacity:(RCTPromiseResolveBlock)resolve rejecter:
     resolve(@(self.getTotalDiskCapacity));
 }
 
-- (double) getFreeDiskStorage {
-    uint64_t freeSpace = 0;
-    NSDictionary *storage = [self getStorageDictionary];
-
-    if (storage) {
-        NSNumber *freeFileSystemSizeInBytes = [storage objectForKey: NSFileSystemFreeSize];
-        freeSpace = [freeFileSystemSizeInBytes unsignedLongLongValue];
+- (uint64_t) getFreeDiskStorage {
+    uint64_t totalFreeSpace = 0;
+    if (@available(iOS 11.0, *)) {
+        NSURL *fileURL = [[NSURL alloc] initFileURLWithPath:NSHomeDirectory()];
+        NSError *error = nil;
+        NSDictionary *results = [fileURL resourceValuesForKeys:@[NSURLVolumeAvailableCapacityForImportantUsageKey] error:&error];
+	    
+        if (results) {
+            totalFreeSpace = ((NSString *)results[NSURLVolumeAvailableCapacityForImportantUsageKey]).longLongValue;
+        }
+    } else {
+        NSError *error = nil;
+        NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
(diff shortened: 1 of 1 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -488,15 +488,28 @@ - (double) getTotalDiskCapacity {
...
+        NSDictionary *dictionary = [[NSFileManager defaultManager] attributesOfFileSystemForPath:[paths lastObject] error: &error];
+        
+        if (dictionary) {
+            NSNumber *freeFileSystemSizeInBytes = [dictionary objectForKey:NSFileSystemFreeSize];
+            totalFreeSpace = [freeFileSystemSizeInBytes unsignedLongLongValue];
+        }
+        
     }
-    return (double) freeSpace;
+    return (uint64_t) totalFreeSpace;
```
mikehardy, line 512: I'm concerned about the types:

1- previous code was getting an `unsignedLongLongValue` and returning that, cast to `(double)` in the return
2- new code ios11+ is getting a `longLongValue` (is it unsigned or not? could be a difference from before) and returning cast to `(uint64_t)` instead of `(double)` - so different return type and there might be some implicit conversion going on from longLongValue to uint64_t
3- new code not-ios11+ is getting `unsignedLongLongValue` (like old code before this change) but is casting to uint64_t instead of double

My thoughts are:
1- keep double as API return value and cast during return statement, why? because react-native is IEEE floating point only for numbers since javascript only has one "number" and it is IEEE floating point (meaning: double)
2- carefully examine if unsignedLongLongValue and longLongValue are actually equivalent and convert to double in same way when cast directly to double. If there is possibility for overflow, then there should be a test against whatever max value is available prior to overflow, and that value should be returned (as opposed to going extremely negative via wrapping around the number space)

Unrelated thought: we are initializing the same error value twice for use in API calls. Should be initialized once, but most importantly: we aren't doing anything with the error value. Why not? Should at least log the error I think? Better if we rejected the promise with the error contents though

mikehardy (changes requested):
Thanks for this! I left comments regarding error handling and some possible subtle type issues I'm concerned about, but I'm positive in general on the change, and with some thought and collaboration I think we can get this merged and released

ios/RNDeviceInfo/RNDeviceInfo.m
gh-pro: thanks for reply, I'm not expert to language natif ios, I just found this code in web, and optimized this, so any suggestion to help is welcome

mikehardy: I'm not an expert either. My comments were pretty thorough though - I need you to investigate and either confirm it's okay after reading about the various data type properties, or take the suggestions I already provided

VikinoCavalier (approved):

mikehardy:
@VikinoCavalier we don't need "approval" we need the comments I made specifically addressed. Do you have sufficient objective-c skills to do that? That would advance the PR

gh-pro:
Hy, sorry i have don't free time to tested and validate the feature, who can tested this ?

gustavoabel:
hey @mikehardy and @rndi-bot, this didn't go forward?

PR #1175, opened 2021-03-10
Use Partial instead of '| {}'

vvscode:
## Description

Using union type `PowerState | {}` doesn't allow to pick properties directly from related-to-power methods results. 

you can check it [here](https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=14&pc=1#code/KYDwDg9gTgLgBDAnmYcBCBDGNhUQZRi1QF44ByAVwDsBraiAd2vLgB8KawAbSgcz7AAJqw7kAxgAsMUPgEtqfURQBmlbt3IBuAFA7QkWHAU4oKjONQAFJrkLE4Abx1w4AIyynEAGWAA3YG4ALjhqSgBbN1xdVw9sXAIiHBDMeLx7HBi4biYbRlwAWQghYBC3CAhuYAxqLIBtWmBEEIBnGCgFPgBdEJrEXQBfPSr4EBC8uyTUDkcBuDJZ3R0QADo4r18A7i04AHpduCsoCBRYRAp1hM3A1iEIYBbQiHhQOTa4CGoEZFRyWfI9EA)
```ts
export type BatteryState = 'unknown' | 'unplugged' | 'charging' | 'full';

export interface PowerState {
  batteryLevel: number;
  batteryState: BatteryState;
  lowPowerMode: boolean;
  [key: string]: any;
}

let x: PowerState | {} = {};

x.batteryLevel; // Property 'batteryLevel' does not exist on type '{}'
```

as soon as `let y: Partial<PowerState> = {};` is a valid expression from the perspective of the type, and it allows to read props from the `PowerState` interface I made this proposal

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅   |
| Windows |    ✅    |

## Checklist


vvscode:
seems I missed `flow` usage

mikehardy (approved):
I like this change, I can see what you mean about the type not being internally addressable if it is a union with `{}`
I hate to say this for Flow users, but the last few of them need to admit Typescript won and migrate. Most of the projects I work on have purged flow types at this point and they must be maintained in the external flow type repository for interested parties. We will do the same at some point, they should be thought of as deprecated and I won't make guarantees on them going forward

PR #1180, opened 2021-03-16
arm64 iOS simulators support

### LINKED ISSUE #1179: Unknown device model in arm64 iOS Simulator on M1 Mac
RNDeviceInfo.m methods are using these checks to detect whether app is running in iOS Simulator:

`if ([deviceId isEqualToString:@"i386"] || [deviceId isEqualToString:@"x86_64"]`

It works for Intel Macs, but not for Apple Silicon Macs, where iOS Simulator runs on the same arm64 architecture as on real iOS devices.

## Summary

|             |     |
| ----------- | --- |
| Version     | 8.0.5   |
| Affected OS | iOS Simulator (M1 Mac)  |
| OS Version  | 14.4   |

## Current behavior

**DeviceInfo.getModel()** and other model-related methods return wrong values (unknown, null, false, etc).

## Expected behavior

As a workaround, if I use `#if TARGET_IPHONE_SIMULATOR` macro in **getDeviceId** and **isEmulator** methods in RNDeviceInfo.m, everything works as it should. So, it would be great to modify those simulator checks.

Arkkeeper:
## Description

Fixes #1179 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -356,24 +356,19 @@ RCT_EXPORT_METHOD(syncUniqueId:(RCTPromiseResolveBlock)resolve rejecter:(RCTProm
     uname(&systemInfo);
     NSString* deviceId = [NSString stringWithCString:systemInfo.machine
                                             encoding:NSUTF8StringEncoding];
-    if ([deviceId isEqualToString:@"i386"] || [deviceId isEqualToString:@"x86_64"] ) {
+    #if TARGET_IPHONE_SIMULATOR
         deviceId = [NSString stringWithFormat:@"%s", getenv("SIMULATOR_MODEL_IDENTIFIER")];
-    }
+    #endif
     return deviceId;
 }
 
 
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
This looks good to me - I'm familiar with that build macro
Just to confirm before merge+release: you have tested this on arm64 and x86 simulator and the return values are correct (i.e., no regression for existing users, and you fixed the problem)

I'm always guessing yes but I have to ask :-). Cheers

mikehardy:
pending question on testing

Arkkeeper:
> you have tested this on arm64 and x86 simulator
yes, it has been tested both on arm64 and x86_64 simulators

mikehardy:
magnificent :-), I have one other PR about to land but I'll merge this now and then release both later today

Thank you!

mikehardy:
https://github.com/react-native-device-info/react-native-device-info/releases/tag/v8.0.6 it's out :rocket:

Issue #1178, opened 2021-03-14
Unable to compile with Mac Catalyst  13


marcosrdz:
https://github.com/react-native-device-info/react-native-device-info/blob/c50b1b93621e6241331e5422c28ec774845a304a/ios/RNDeviceInfo/RNDeviceInfo.m#L116

This UIIdiom is only available in native macOS and Mac Catalyst 14. This breaks the ability to compile with Mac Catalyst 13.

mikehardy:
PRs happily merged!

mikehardy:
Hi there - after some investigation and testing on #1181 it appears that Xcode 12, which is the minimum for react-native these days (a reasonable stance, it's needed for new APIs) won't let you target catalyst 13 anymore, so requiring catalyst 14 seems reasonable

PR #1181, opened 2021-03-17
fix(ios, deviceType): support iOS-compiled app running on macOS

user-none:
## Description

Expansion of PR #1137. That PR only checked if the app is an iOS app built for macOS not if it is an iOS app running on macOS. See https://developer.apple.com/documentation/xcode/creating_a_mac_version_of_your_ipad_app for details about how Catalyst is for creating a macOS native type package from an iOS app. #1137 uses a compile time define which is only present when creating a macOS version.

This adds a runtime check for running unmodified iOS apps on macOS, https://developer.apple.com/documentation/apple-silicon/running-your-ios-apps-on-macos . This allows proper detection of the underlying device type the iOS app is running on.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

* [ X] I have tested this on a device/simulator for each compatible OS

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -111,7 +111,11 @@ RCT_EXPORT_MODULE();
 {
     switch ([[UIDevice currentDevice] userInterfaceIdiom]) {
         case UIUserInterfaceIdiomPhone: return DeviceTypeHandset;
-        case UIUserInterfaceIdiomPad: return TARGET_OS_MACCATALYST ? DeviceTypeDesktop : DeviceTypeTablet;
+        case UIUserInterfaceIdiomPad:
+            if (TARGET_OS_MACCATALYST || (@available(iOS 14.0, *) && [NSProcessInfo processInfo].isiOSAppOnMac)) {
+                return DeviceTypeDesktop;
+            }
+            return DeviceTypeTablet;
         case UIUserInterfaceIdiomTV: return DeviceTypeTv;
         case UIUserInterfaceIdiomMac: return DeviceTypeDesktop;
         default: return DeviceTypeUnknown;
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -112,8 +112,13 @@ RCT_EXPORT_MODULE();
     switch ([[UIDevice currentDevice] userInterfaceIdiom]) {
         case UIUserInterfaceIdiomPhone: return DeviceTypeHandset;
         case UIUserInterfaceIdiomPad:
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Clever! This looks great to me. Out of curiosity - since you're in the area, do you see anything easy to do for #1178 ? Perhaps just an `@available` check ? I think the same issue could affect your change, but I'm not sure how concerned to be - I don't know the relative adoption of various catalyst versions

mikehardy:
@user-none - just pending a question about being even more compatible than existing code already is, nothing wrong with the code as is I like this solution

user-none:
@mikehardy #1178 is easy enough to fix. Give me a bit and I'll update this PR with the code for it.

mikehardy:
Champion! Seriously, thank you

user-none:
I found a bug in my code while looking at the other ticket. I've pushed up a fix but I'm still working on the other one.

user-none:
The best I can come up with is:

```objc
#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 140000
    case UIUserInterfaceIdiomMac: return DeviceTypeDesktop;
#endif
```

This should work because Catalyst tracks iOS versions. The header for `enum UIUserInterfaceIdiom` is an iOS header and specifically sets UIUserInterfaceIdiomMac as available in iOS 14. I said should work because I can't test it. The current version of Xcode ships with Catalyst 14 and when I create a new project it doesn't allow targeting for Catalyst 13.

I can commit that if you don't want, or you can make 14 a min dependency going forward.

mikehardy:
I'm inclined to just say Catalyst 14 is a minimum version, Xcode 12 is already a minimum version and I am fine in general with the statement "You must run current stable Xcode to develop things for Apple products", then enforce minimums based on that.

At the end of the day if it's a really serious issue, that's why patch-package exists, and why we all use open source. You can just fix it, and it's not even that hard.

Thanks for looking into this I do appreciate it

So then I think (please confirm) there is actually nothing else actionable here, and I should close the other issue re v13 catalyst while merging and releasing this. I think that should match your thinking?

user-none:
> At the end of the day if it's a really serious issue ...

The code to work around it is in this conversations so if it needs to be dealt with, we know what to do.

> So then I think (please confirm) there is actually nothing else actionable here, and I should close the other issue re v13 catalyst while merging and releasing this. I think that should match your thinking?

Nothing else actionable. It's ready for merging. Closing out #1178 is appropriate since the 14 is the minimum required catalyst version.

mikehardy:
https://github.com/react-native-device-info/react-native-device-info/releases/tag/v8.0.6 it's out :rocket:

PR #1184, opened 2021-03-23
adding semantic release

schie:
## Description

- Replaces `np` with `semantic-release`
- adds a GitHub action workflow to auto-release on push into master or manual trigger (if semantic-release deems a new version is warranted)
  - we can remove the auto-release on push if need be

@mikehardy, 

In order to get it working in GH Actions, the following secrets will need to be set in the repo's settings:
- `GITHUB_TOKEN` — [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) to be able to make commits to update changelog, create and push tags, make releases, etc...
- `NPM_TOKEN` — [access token](https://docs.npmjs.com/about-access-tokens) to be able to push to NPM

Also, to perform a dry-run to test, you can run `yarn shipit`. Semantic Release defaults to `dry-run mode` if it isn't in a detected CI environment, but you can also pass the flag [`--dry-run`](https://semantic-release.gitbook.io/semantic-release/usage/configuration#dryrun)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |  N/A     |
| Windows |    N/A     |

## Checklist

image


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
mikehardy, line 23: - setup-node is on v2 
- shouldn't we use node-version 14 these days? I don't feel strongly though I have a preference for "latest stable" unless there is a reason not to

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install dependencies
+        run: rm -rf node_modules && yarn install --frozen-lockfile
```
mikehardy, line 25: I'd prefer this be a run script in package.json so it's easier for developers to do the same steps when troubleshooting CI things, something like `yarn ci:install` or similar (i..e, a `ci` prefix for ci-related run scripts then name it something reasonable for the CI step?)

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
mikehardy, line 29: I just sent you invites to the github org (so implies repo access, and can add secrets at org or repo level) and to the npmjs.com package

Should have done that sooner it has been on my todo list for weeks! then you may implement this and test it. I'm not concerned at all about attribution, but I know that setting these things up is best done with access and testing so please have at it with my blessing :-)

mikehardy (approved):
This is a real step forward for the package, thank you sincerely!

I added a couple quick comments, only one of which I think is an actual "please change this", and the other is related to access so you can actually test this and know it's good.

I approve in general though so please feel free to set up tokens and merge at will when you're satisfied with it, following that there is a small PR from a community member that might result in a great first exercise of the system?

For the avoidance of doubt since you now have power to do releases and I'm the only releaser at the moment: I never expect perfect from anyone I work with, on the contrary I expect errors to happen because we are human. My expectations are focused on responsiveness and best effort and fixes when the things do go wrong :-). That's the standard I hold myself to anyway.

Cheers

.github/workflows/release.yml
schie: the `v1` was probably a fat-finger move. I'll update along w/ `node-version: 14`

.github/workflows/release.yml
schie: agreed. I'll update

schie:
it looks like GitHub automatically adds an access token, so I just created one for NPM and added it to the organization. 

if semantic release fails to write to the repo, I'll add a Personal access token and re-run it.

mikehardy (approved):
looks super clean :-)

mikehardy:
:popcorn:

schie:
@mikehardy, looks like it failed as part of the `clean` script. 

Is there a reason why cleaning of the example project occurs before deployment?

mikehardy:
@schie I don't remember a specific reason :thinking: - I remember a vague reason before our .gitignores and .npmignore's and package.json files were cleaned up #872 where a lot of manual cleaning was done prior to shipping (and it was ineffective anyway) so it might just be old useless stuff

Issue #1056, opened 2020-08-11
getSerialNumberSync api error?


M45ter:
# Bug report
when android SDK < 26, getSerialNumberSync() always return "unknown"?

> public String getSerialNumberSync() {
>     try {
>       if (Build.VERSION.SDK_INT >= 26) {
>         if (getReactApplicationContext().checkCallingOrSelfPermission(Manifest.permission.READ_PHONE_STATE) == PackageManager.PERMISSION_GRANTED) {
>           return Build.getSerial();
>         }
>       }
>     } catch (Exception e) {
>       // This is almost always a PermissionException. We will log it but return unknown
>       System.err.println("getSerialNumber failed, it probably should not be used: " + e.getMessage());
>     }
>     return "unknown";
>   }
> 

## Environment info

`react-native info` output:

```bash
# "react-native": "0.62.2",
# "react-native-device-info": "^5.5.7",
```

mikehardy:
I fixed your code block so it was readable - triple-backticks are your friend https://www.markdownguide.org/extended-syntax#syntax-highlighting

The API is only available from API26 on. https://developer.android.com/reference/android/os/Build#getSerial()

M45ter:
this is better?
```java
if (Build.VERSION.SDK_INT >= 26) {
...
} else {
     return Build.SERIAL;
}
```

mikehardy:
Please read the docs for the API you propose: https://developer.android.com/reference/android/os/Build#SERIAL

> A hardware serial number, if available. Alphanumeric only, case-insensitive. This field is always set to Build#UNKNOWN.

So, no, that won't help. Getting hardware identifiers is a losing proposition, you're fighting the operating systems now. They don't want to give you hardware identifiers. My best advice is to stop trying. Use something like Firebase Instance ID or similar

M45ter:
ok, thanks

stewartmcgown:
We're deploying a lot of Sunmi V2 devices, which allow access to the serial number via BUILD.SERIAL. I don't see a reason to not have the else statement and return BUILD.SERIAL if it isn't unknown?

stewartmcgown:
In case someone stumbles across this from Google and is in the same very specific sitation:

Add this to`patches/react-native-device-info+7.0.2.patch`:
```
--- a/node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
[...]

PR #1185, opened 2021-03-24
fix: get legacy serial number for old APIs (#1056)

stewartmcgown:
## Description

Following the disussion in #1056 this will allow users of old MDM devices like the Sunmi V2 and others to access serial numbers via their privileged system access permissions assigned by the MDM.

This shouldn't cause any problems as the Build.SERIAL will be populated with 'unkown' regardless of permissions.

Been using this via patch-package for months with live Sunmi V2 devices. Works like a charm.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -688,6 +688,8 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
         if (getReactApplicationContext().checkCallingOrSelfPermission(Manifest.permission.READ_PHONE_STATE) == PackageManager.PERMISSION_GRANTED) {
           return Build.getSerial();
         }
+      } else {
+        return Build.SERIAL;
       }
     } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Oh cool - thanks @stewartmcgown 

yeah, this seems like a reasonable stance and I understand the utility, those old devices will be around for a while and it was permitted then, I don't think we are in violation of any expectations here.

@schie this is the PR I mentioned that might make a good exercise of #1184 - so I'm appproving and am +1 on merge+release at will but won't merge it yet so it can serve as the first semantic-release release

schie (approved):
This looks good to me. 

We'll have to keep an eye on it since `Build.SERIAL` was deprecated in API 29. Our `minSdkVersion` is 16, so it should be fine for now.

mikehardy:
Dang, it's always something - release action failure:

>     ' ! [remote rejected] HEAD -> master (protected branch hook declined)\n' +

I'm okay with whatever exception is necessary to make it go, even including removing master protection if needed

schie:
@mikehardy, semantic-release is good-to-go!

PR #1187, opened 2021-03-24
ci(.github/workflows/release.yml): using `GH_TOKEN` as env and secret

schie:
## Description

Regarding the action, `Release` ,

replaces `GITHUB_TOKEN` environment variable w/ `GH_TOKEN` and updates the value to use the secret of the same name. `GITHUB_TOKEN` secret is a special secret that can't be set, so if you're writing into a protected branch, you'll need to add a PAT of a user that can write to said protected branches.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    N/A     |
| Windows |    N/A     |

## Checklist

-- 

_it sucks that there isn't a good way to test actions locally_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,6 @@ jobs:
         run: yarn ci:install
       - name: Release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: yarn ci:shipit


schie:
I always forget to add the `[skip ci]` jazz

mikehardy:
> so if you're writing into a protected branch, you'll need to add a PAT of a user that can write to said protected branches.

Yep, that's my experience

> it sucks that there isn't a good way to test actions locally

Yep, that's my experience ;-)

There is actually a project to do local runners and stuff, but it's a total yak shave I think vs the typical tasks + time taken to just bull through it on master, so I just bull through it on master as well. And, it's okay. Not awesome, but okay, and at the end everything works and is then awesome :-)

PR #1190, opened 2021-03-30
fix(ios): Remove IPv4-only method use to prevent App Store warnings

### LINKED ISSUE #1135: Support ipv6 addresses
## Summary

https://developer.apple.com/support/ipv6/

Apple requires (but hasn't really enforced?) supporting ipv6, and the usage of `inet_ntoa` (an ipv4 specific method) causes some validators to flag this as not supporting ipv6 as required by the Apple App Store rules:

https://github.com/react-native-device-info/react-native-device-info/blob/master/ios/RNDeviceInfo/RNDeviceInfo.m#L548

|             |     |
| ----------- | --- |
| Version     |  7.2.1   |
| Affected OS | iOS  |
| OS Version  | 14.2  |

## Current behavior

Usage of `inet_ntoa` is causing a security validator to flag RNDeviceInfo as non-compliant with App Store rules.

## Expected behavior

Support ipv6 addresses by calling `inet_ntop`?

gtfunes:
## Description

Fixes #1135 by replacing the use of `inet_ntoa` with `inet_ntop`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -541,7 +541,9 @@ RCT_EXPORT_METHOD(getSupportedAbis:(RCTPromiseResolveBlock)resolve rejecter:(RCT
         // Loop through linked list of interfaces
         temp_addr = interfaces;
         while(temp_addr != NULL) {
-            if(temp_addr->ifa_addr->sa_family == AF_INET) {
+            sa_family_t addr_family = temp_addr->ifa_addr->sa_family;
+            // Check for IPv4 or IPv6-only interfaces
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Campeón! :trophy: 
Thank you - this all looks good to me, but having not pulled it and run it - to verify, you tested this on a real iphone and it worked well for you?

I imagine the answer is "of course", but I have to check :-)

mikehardy:
@schie you are magic :mage: - thank you so much for setting up semantic-release
@gtfunes your fix is out there, buen provecho :-)

Issue #1216, opened 2021-04-10
Feature suggestion: is app installed from store (App Store/Google Play)


dimaportenko:
I had a task to separate analytics in the app installed from the store (App Store/Google Play) and all other builds from Testflight, direct APKs, in development, etc. 

I've implemented it with native modules
iOS - [iOS stackoverflow answer](https://stackoverflow.com/questions/18282326/how-can-i-detect-if-the-currently-running-app-was-installed-from-the-app-store/18619270#18619270) and [App Center SDK solution](https://github.com/microsoft/appcenter-sdk-apple/blob/develop/AppCenter/AppCenter/Internals/Util/MSACUtility%2BEnvironment.m)    
Android - [Android stackoverflow solution](https://stackoverflow.com/questions/37539949/detect-if-an-app-is-installed-from-play-store/37540163#37540163)

I would like to share my solution with the community and I think this library could be a good place to do it. 

Please let me know if this feature suits this library I'll do PR then.

mikehardy:
It would definitely fit, we already have it for Android and Windows actually https://github.com/react-native-device-info/react-native-device-info#getinstallerpackagename

If any other platforms are missing or the current platforms may have better implementations please post a PR, we'll happily merge it

PR #1217, opened 2021-04-10
Add getcurrentappenvironment

dimaportenko:
## Description

Related #1216

add `getCurrentAppEnvironment` function for ios to determine whether app was downloaded from `AppStore`,`TestFlight` or `Other`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ The example app in this repository shows an example usage of every single API, c
 | [getHost()](#gethost)                                             | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌  |
 | [getIpAddress()](#getipaddress)                                   | `Promise<string>`   |  ✅  |   ✅    |   ✅    | ❌  |
 | [getIncremental()](#getincremental)                               | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌  |
-| [getInstallerPackageName()](#getinstallerpackagename)             | `Promise<string>`   |  ❌  |   ✅    |   ✅    | ❌  |
+| [getInstallerPackageName()](#getinstallerpackagename)             | `Promise<string>`   |  ✅  |   ✅    |   ✅    | ❌  |
 | [getInstallReferrer()](#getinstallreferrer)                       | `Promise<string>`   |  ❌  |   ✅    |   ✅    | ✅  |
 | [getInstanceId()](#getinstanceid)                                 | `Promise<string>`   |  ❌  |   ✅    |   ❌    | ❌  |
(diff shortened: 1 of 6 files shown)


src/index.js.flow
```diff
@@ -44,6 +46,8 @@ declare module.exports: {
   getCarrierSync: () => string,
   getCodename: () => Promise<string>,
   getCodenameSync: () => string,
+  getCurrentAppEnvironment: () => Promise<string>,
+  getCurrentAppEnvironmentSync: () => string,
```
schie, line 50: Wouldn't these return the `AppEnvironment` type?

src/internal/privateTypes.ts
```diff
@@ -55,6 +55,8 @@ interface ExposedNativeMethods {
   getCarrierSync: () => string;
   getCodename: () => Promise<string>;
   getCodenameSync: () => string;
+  getCurrentAppEnvironment: () => Promise<string>;
+  getCurrentAppEnvironmentSync: () => string;
```
schie, line 59: Shouldn't these return the `AppEnvironment` type?

mikehardy:
I would like for this interface to be unified with the existing similar interface for android that (and windows) that returns package installer. Yes the types will be different but I assume they are different for windows and android as well (they should be...) yet at heart they are all strings, and if the strings are formally specified (i.e. constrained to a small set as the values you mention above) that can still be unioned in type-wise

I don't think this requires a completely new API

so to be specific - please just make this the iOS implementation of the existing API which is not currently supported on iOS and I think we're close. I love that seeing the app install method for iOS will be possible with this though! that can b very useful for feature toggles etc (like, showing a bug report / feedback link or not, turning on screen recording during usage or not etc)

dimaportenko:
@mikehardy so you would like instead of the new API function it'll be iOS implementation of `getInstallerPackageName()`, is it correct?

mikehardy:
Exactly that, and thank you! As I said I can imagine this being really useful for people

dimaportenko:
@mikehardy ok, it makes sense to me as well. I'll do this change.

dimaportenko:
I changed the implementation. It's `getInstallerPackageName()` for iOS now. Please take a look.

schie:
@dimaportenko can you update the README to reflect your changes? 

Otherwise, it looks good to me.

dimaportenko:
@schie done

mikehardy (approved):
Nice this looks excellent to me as well, thank you @dimaportenko !

schie:
hmm.. looking now

schie:
I need to look more into RNDI's permissions

PR #1220, opened 2021-04-18
Adding mavenCentral() as jcenter() is shutting

maheshwarimrinal:
Adding mavenCentral() as jcenter() is shutting down

## Description

Fixes #1220

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -6,7 +6,6 @@ buildscript {
         repositories {
             mavenCentral()
             google()
-            jcenter()
         }
         def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '4.1.1'
 
@@ -47,7 +46,6 @@ repositories {
         // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
         url "$rootDir/../node_modules/react-native/android"
     }
-    jcenter()
 }
 
 dependencies {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,6 +4,7 @@ buildscript {
     // module dependency in an application project.
     if (project == rootProject) {
         repositories {
+            mavenCentral()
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
Oh thank you! I'm just working through this on all my projects but had not gotten here yet. Adding mavenCentral is great, but I don't think it is useful by itself. Can you remove jcenter as well? If not, I really want to know we cannot yet + which dependencies block it

mikehardy:
I checked locally to see if the example could be cleaned of jcenter/bintray and unfortunately react-native is not quite ready yet.

Failures are: 
- flipper (but they have resolved upstream - earliest tag published to mavenCentral appears to be [0.75.x](https://mvnrepository.com/artifact/com.facebook.flipper/flipper-network-plugin) for which I requested a cherry-pick to react-native 0.64)
- com.facebook.yoga:android-annotations - work tracked here https://github.com/facebook/yoga/issues/1065
- com.facebook.fbjni:fbjni-java-only - [newer versions are on mavencentral](https://mvnrepository.com/artifact/com.facebook.fbjni/fbjni-java-only) but the dependency bump is transitive in flipper and is not until flipper 0.79.1

[Yoga android annotations appear to be used in flipper as well](https://github.com/facebook/yoga/issues/1065#issuecomment-822023614), so whenever yoga android annotations *finally* publishes to mavenCentral

mikehardy (approved):
This looks great to me - thank you for taking it on!
The comment I left on purging the example of jcenter was just context as I researched it for publication to broader react-native community as well as the yoga repository itself, there is nothing more we can do here

mikehardy:
@schie not urgent at all - but still an unhappy release bot :thinking: https://github.com/react-native-device-info/react-native-device-info/runs/2375121728?check_suite_focus=true#step:5:83

schie:
looking into it now

PR #1225, opened 2021-05-04
Fix init if navigator is not defined

janicduplessis:
## Description

Add a check to see if navigator is defined. This fixes server side rendering in node where navigator is not defined. Since init is called when the module is required this causes an issue, even if not using any of the library methods.

## Compatibility

web only fix

## Checklist


mikehardy (approved):
LGTM - hi @janicduplessis :wave:

Issue #922, opened 2019-12-20
DeviceInfo.getUniqueID() is same in multiple iPads registered with same Apple IDs


kmvkrish:
# Bug

The method `DeviceInfo.getUniqueID()` returns the same value for iPads that use the same Apple Account and have iCloud sync enabled.

## Environment info

React native info output:

```bash
React-Native: 0.59.10
ReactJS: 16.8.3
React-native-Device-Info: 0.24.0
```

Library version: **0.24.0**

## Steps To Reproduce

1. Register all the iPads with the same Apple IDs.
2. Enable iCloud Sync on all the devices. Enable Keychain Sync also.
3. Check for the `DeviceInfo.getUniqueID()` return value.
...

Describe what you expected to happen:

1. The UniqueID (UUID) should not be synced across multiple devices using the same Apple ID.

mikehardy:
We're just exposing what the native API gives us. Have you consulted the upstream documentation to see how it behaves?

https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L37

Alternatively you may be content with using async-storage and UUIDv4 packages in an easy combo to uniquely identify installs if that's the goal

kmvkrish:
Hi @mikehardy 

> https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L37

In this line, the code checks for the UniqueID in the following sequence:
1. Keychain.
2. If not found in Keychain, checks in UserDefaults.
3. If not found in UserDefaults, checks for AppleIFV - `UIDevice.current.identifierForVendor`.
4. If not found, then returns a new UUID generated.

From iOS 6,` identifierForVendor` will always be returned and that is unique per device per vendor. Due to this, the AppleIFV is always being returned when queried for the first time and the same is being saved in Keychain and UserDefaults - [Persist UID to NSUserDefaults and Keychain, if not yet saved](https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L56)

I have also seen the code for saving the UniqueID for iOS in Keychain [https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L71](https://github.com/react-native-community/react-native-device-info/blob/master/ios/RNDeviceInfo/DeviceUID.m#L71).  In the code DeviceUID.m, the `save` method didn't use this attribute.

[...]

PR #1226, opened 2021-05-05
fix(iOS): syncUniqueId doesn't create a new random ID if there is no access to IDFV

InkOut:
## Description

Fixed syncUniqueId so that it doesn't create a new random ID if there is no access to IDFV

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [+] I have tested this on a device/simulator for each compatible OS
* [+] I added the documentation in `README.md`
* [-] I mentioned this change in `CHANGELOG.md`
* [+] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [+] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -980,7 +980,9 @@ This is a constant and may be referenced directly
 
 Gets the device unique ID.
 On Android it is currently identical to `getAndroidId()` in this module.
+
 On iOS it uses the `DeviceUID` uid identifier or generates random UUID if IDFV is not available. On iOS 14.5 or earlier, you must request IDFA access. For example use [react-native-permissions](https://github.com/zoontek/react-native-permissions).
+
 On Windows it uses `Windows.Security.ExchangeActiveSyncProvisioning.EasClientDeviceInformation.id`.
 
 #### Examples
--- a/README.md
+++ b/README.md
@@ -1004,7 +1004,7 @@ let uniqueId = DeviceInfo.getUniqueId();
 
 This method is intended for iOS.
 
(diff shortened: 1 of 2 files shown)


mikehardy (changes requested):
I'm +1 on this change but the documentation in the comment above the method, and the method description in the readme are now incorrect, and that's just as important as the code change. Can you update those to indicate they now delegate to getUniqueId on iOS if IDFV is not available?

InkOut:
@mikehardy thank you for your comment. I think I've fixed readme.

mikehardy (approved):
Ok - LGTM - thanks - Final question - this is just about how to release it - to me this is actually a point release - semantic-version patch. I don't believe this changes anything for installed users - is that correct? I am perfectly fine releasing a breaking change / semver major if this changes things for installed users but I can't see how it would.

What do you think?

InkOut:
Yes, i'm +1 on semantic-version patch:
1. I would say that it's more a bug fix, not a breaking change. Sync should sync without others updates
2. I don't think there are more than 10 users using this function

InkOut:
temporary closing

mikehardy:
Even if 1 person is using semantic version says if it's breaking, it's breaking - which is fine! I'll just bump the right part of the version number
But is sounds like you agree that existing users will not be affected, which means patch release yes

Bardiamist:
Hey, `syncUniqueId` was added for [another purpose](https://github.com/react-native-device-info/react-native-device-info/issues/922#issuecomment-585027771). After this PR logic be another and will not help after restore second device from iCloud backup if no access to IDFV. Not sure why you need this change but if you need it then consider to add new method instead change `syncUniqueId`.

PR #1231, opened 2021-05-07
build(android): gradle plugin 4.2 ignores submodule gradle.properties

mikehardy:
## Description

Android Gradle Plugin 4.2 ignore submodule gradle.properties, so this file is unused

remove it - tested it via removing it from the installed location in
node_modules/react-native-device-info/android/gradle.properties and it
was fine

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |         |
| Android |    ✅     |
| Windows |        |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -7,7 +7,7 @@ buildscript {
             mavenCentral()
             google()
         }
-        def buildGradleVersion = ext.has('buildGradlePluginVersion') ? ext.get('buildGradlePluginVersion') : '4.1.1'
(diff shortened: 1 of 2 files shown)


mikehardy:
Mmm good question but I think for `build` no? It's not really a release-worthy thing so I tagged it build and wasn't expecting a release anyway. That tag is I think just a sort of more specialized `chore` (like `lint` in my mind)

schie:
Yeah, agreed. Just wanted to make sure all is fine.

PR #1238, opened 2021-05-13
feat(windows, isTabletMode, isKeyboardConnected, isMouseConnected, DeviceType GamingConsole) and fix(windows, isTablet): revise condition

### LINKED ISSUE #1237: Windows is reporting IsTablet incorrectly
Windows is reporting IsTablet incorrectly

## Summary

|             |     |
| ----------- | --- |
| Version     |  8.1.3 |
| Affected OS | Windows |
| OS Version  | Windows 10   |

It returns true when it should return false. And I'd imagine the opposote.

I will be doing a PR here soon to address this.

namrog84:
Added isTabletMode, isKeyboardConnected, and isMouseConnected
Fixed IsTablet returning true when we are on a desktop windows (i.e. not on a tablet).

## Description

* Window's has devices that can have 2 interaction modes, tablet mode and desktop mode. Added a new API to expose this API to RN apps.
* Added is Mouse and Keyboard connected api.  
* Fixes bug #1237 

TODO for future.  
* Add equivalent for ios/android where applicable.
* Add event hooks where applicable.

Food for Thought:
* What is a "Tablet" hardware in the world of Windows?  
  * Is a Microsoft Surface (tablet)(2015-2020), actually a "tablet" or more like a hybrid, or something else? Most Windows apps do not look or behave differently in this context. 
* Could/Should a Xbox Consoles or Surface Hub, or Microsoft Surface Table(2012) considered a tablet? 
* **Should IsTablet just always be false for windows? And should there be other react-native-device-info API to help better differentiate the devices?** 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Windows |    ✅     |

## Checklist

  * I've tested it on a Windows Desktop Computer. I do not have other Windows devices. 


namrog84:
Having started to reflect on this, after spending some time investigating and testing my changes. I now am seriously considering if IsTablet should just always return false on windows. 
Since even in the rare instances of a Tablet for windows, most things do not distinguish based upon "Tablet" vs not, but more likely to distinguish based upon Sku, ScreenSize, IsKeyboard, etc...

windows/code/RNDeviceInfoCPP.h
```diff
@@ -44,7 +44,29 @@ namespace winrt::RNDeviceInfoCPP
...
+      auto deviceFamily = AnalyticsInfo::VersionInfo().DeviceFamily();
+
+      bool isTabletByAnalytics = deviceForm == L"Tablet" || deviceForm == L"Mobile" || deviceFamily == L"Windows.Mobile";
+      bool isTabletByOsRegex = std::regex_match(os, std::regex(".*windowsphone.*", std::regex_constants::icase));
+
+      if (isTabletByAnalytics || isTabletByOsRegex) {
+          return true;
+      }
+
+      // If there is no keyboard, is that considered functionally equivalent to a 'tablet' in the Windows OS world?
```
mikehardy, line 63: Mmm - that's a tough one. I think you're right *if* it isn't a `Handset` device type. No keyboard means it's definitely not a Desktop. Still might be a TV?

I think the difference between phone and tablet is screen size, it's used to determine layout style. Keyboard is used to determine interaction and should not be conflated I think except as some sort of backup

That reminds me that "Desktop" is actually a valid `getDeviceType` return type - added to support macOS in #1137 but not documented until just now. You could probably add that in this PR since it's so intimately related - "isTablet" should arguably just take do an internal 

```c++
// terrible untested typed in to a web text area code from a java dev follows
if (getDeviceType() == L"Tablet") { 
  return true; 
} 
return false;
```

mikehardy (approved):
I think this could go in right now as an improvement, and I'm a fan of incremental positive change so if this fixes you right now and you're out of time, and it passes CI (I just approved the run) then I can merge it

But my comment on maybe refactoring it to be based on a fuller getDeviceType implementation stands and would make for a richer / better implementation I think. Could table that for now / forever, could do it really quickly if you like in this PR, or could do it as a follow on - I'll say "implementor's choice" - let me know what you prefer

mikehardy:
Just saw you edited your comment - that is the heart of it right? what type of device is each device? and microsoft definitely loves crossover devices which makes it harder. Maybe a getDeviceTypeDetail API that exposes the complexity by containing the full set of the types you inspected for tablet determination, then a curated "getDeviceType" return where you try to return reasonable things but explain for windows if people are getting specific, use the detail API ?

namrog84:
I am on the fence about it.  I will probably leave as-is for right now.  It works and unblocks on the short term. And I think it's more versatile than it was. 

1. I still think IsTablet might want to always be false on Windows. 
2. I think it might be worthwhile to add a "IsKeyboardConnected", or something kind of API, which could also be used on android/ios with potential external peripherals? And help windows apps better inform app behavior.
3. I feel inclined to leave the code in, to help give ideas and good reference for person who then tackles implementing DeviceType on windows. Right now it's hardcoded to "Desktop" for windows.  
  * I can think more on this and will implement DeviceType to help differentiate between Mobile, Desktop, TV, Console, and others for Windows. I will try and do that soon in a follow up PR.

mikehardy:
That sounds like a reasonable approach - incremental improvement now (yes :fire:!) and maybe some more detailed APIs in the future that help designers decide what sort of UI to present. Clearly Windows isn't always "Desktop" :laughing: but at least now it'll say tablet mostly-correctly which let's it be a touch UI vs always Desktop

Ci is chewing on it now

mikehardy (approved):

mikehardy:
Quick question to verify: you have run this locally and it seems sensible to you? All works fine?

namrog84:
I was copy/pasting around a bit between projects, and had some bad copy/paste and with additional last minutes changes that weren't as well tested. 

I can do a 1 final full proper test on these final changes tomorrow (Friday/Sat). I appreciate it! :thumbsup:

mikehardy:
no problem - CI passed, for what that's worth :-), it actually does execute quite a bit but it's nothing like actually looking at the example app and seeing the result correct on device. Just @ me when you're ready, I'm watching for it

Cheers!

asklar:
following up offline internally first :)

mikehardy:
@asklar no problem, I'll move as fast or slow as you / your group wants :-), I truly appreciate the help on the windows side of the code base and hopefully it's clear I'm open to new ideas on the API if it helps people use the module. Philosophically I really like exposing all the details of a device without opinion if possible (more like "hasKeyboard", "hasTouchScreen", etc - although I think window dimensions are supported well by another module?) and then if there is an "opinionated" API (like "isTablet") having it be a best effort but with details available for apps with fewer choices to make. Anything like that is +1 for me

namrog84:
Talked offline.  Thanks Asklar! 

Was pointed to this API
https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.uiviewsettings.userinteractionmode?view=winrt-19041

"Some devices (PC, laptop, tablet) support both a Desktop (mouse-optimized) and a Tablet (touch-optimized) mode.
Users can switch between running in Tablet mode and Desktop mode by going to Settings > System > Tablet mode and setting Make Windows more touch-friendly when using your device as a tablet."


I think this is the most appropriate usage of what is considered isTablet in a Windows environment.  Updated functions to reflect this. 

Some things to do still in the future.  Right now both isTablet and deviceType are provided as constants.  And realistically should longer term only be consumed as a react_method call, and possibly even hook up an event with some windows API for a callback if the user changes mode.  But will leave that for another day. 

Ultimately just wanting to fix the bug in that isTablet is reporting incorrectly today.

windows/code/RNDeviceInfoCPP.h
```diff
@@ -35,16 +36,30 @@ namespace winrt::RNDeviceInfoCPP
...
     bool isEmulatorHelper(std::string model)
     {
       return std::regex_match(model, std::regex(".*virtual.*", std::regex_constants::icase));
     }
 
-    bool isTabletHelper(std::string os)
+    bool isTabletHelper()
     {
-      return !std::regex_match(os, std::regex(".*windowsphone.*", std::regex_constants::icase));
+      auto view = winrt::Windows::UI::ViewManagement::UIViewSettings::GetForCurrentView();
```
namrog84, line 48: Still need to move to UI thread before completing PR.
Else it intermittently fails for being on wrong thread.

namrog84: It 100% needs to be on UI thread.  Not a huge fan of this approach, it does work, but I am open to suggested changes.

namrog84: FYI Mike, would like to give until Monday or Tuesday to make sure a few people can get eyes on most recent changes.

mikehardy: Sure thing - and I had never thought about hybrids actually being dynamic with respect to whether they should have a touch-oriented or Keyboard/Mouse type interface put up, but honestly even iPadOS behaves similar now IIRC? And Android may as well. If isTablet is intended to be a packaged-up heuristic it should be dynamic with an event/hook as well but I agree that's a bit of overengineering for what I imagine is an immediate problem.

I would still lobby for either APIs per typical input device (hasKeyboard, hasMouse or similar) or perhaps like react-native-netinfo has done more of an "InputDevice" object where all available input devices and output devices were available. That idea could be extended to "Displays" and "Modems" etc for a more general purpose tree of information vs the current approach with an approaches-infinity number of endpoints.

All to say I'm open to re-shaping things to be useful as react-native itself expands to include more than just "handsets and maybe 1 or 2 info tidbits about other things". This is a foundational library and I want it to really support all devices well

Cheers

namrog84: I did a patch-package for our local problem, so I am in no rush for this PR to go thru.  And blocking on the JS thread isn't great either (As per my most recent change).   I do agree with everything you've said and I do think it's a tough situation.  What are good short term and long term considerations, with what level of change required. 

I agree that longer term, having more dynamic keyboard, mouse, and the dynamic event hook nature is definitely a great idea.

For the isTablet today, perhaps I should just revert back to an earlier PR changeset and leave these proposed changes and dynamic async for a future event/api reshaping PR? Then maybe work on a 2nd PR that adds in the methods for for keyboard and mouse and event hook for dynamicTabletMode? 


I will leave this PR open and also gather some more insights/thoughts this coming week.  

Obviously you get the final say on things, so if there is anything you'd personally specifically want, do let me know and I can do that too.

namrog84:
More time to ponder and thinking is always good.  

I reworked the whole PR a bit to better align with what I think are the better long term goals.

* Added new APIs per our conversation around keyboard and mouse. IsMouseConnected and IsKeyboardConnected
    * Probably worth adding hooks in the future if possible to detect connected/unconnected events.
* Added new api for IsTabletMode to better align with 'dynamic nature' of switching modes
    * In the future, should add an event/hook for this, but out of scope for what I wanted as a simple PR.
* Reverted some things around isTablet to better align with 'hardware' and less about modes, keyboards, or mouse. Per your comments. Also since it was async only call due to the APIs, only added the async nature. 
    * In the future, should add an event/hook for this, if it gets changed.

windows/code/RNDeviceInfoCPP.h
namrog84: Resolving this comment thread and going back to parent PR for further conversation. I think I was able to address most of the concerns with something I know I am happier with than some of my earlier iterations.

windows/code/RNDeviceInfoCPP.h
```diff
@@ -35,16 +43,37 @@ namespace winrt::RNDeviceInfoCPP
...
+      // AnalyticsInfo is considered flakey or unreliable by most sources to be used, but no recommended alternatives were discovered.
+      // DeviceForm potential but not inclusive results:
+      // [Mobile, Tablet, Television, Car, Watch, VirtualReality, Desktop, Unknown]
+      // DeviceFamily potential but not inclusive results:
+      // [Windows.Desktop, Windows.Mobile, Windows.Xbox, Windows.Holographic, Windows.Team, Windows.IoT]
+      auto deviceForm = AnalyticsInfo::DeviceForm();
+      auto deviceFamily = AnalyticsInfo::VersionInfo().DeviceFamily();
+      
+      bool isTabletByAnalytics = deviceForm == L"Tablet" || deviceForm == L"Mobile" || deviceFamily == L"Windows.Mobile";
+      bool isTabletByOsRegex = std::regex_match(osName, std::regex(".*windowsphone.*", std::regex_constants::icase));
```
asklar, line 70: we don't support running on windows phone afaik (RNW requires RS3 IIRC) so I think you can remove the regex and EAS stuff.
FWIW AnalyticsInfo is not flakey or unreliable :) it might return values that don't match what you'd expect but that is by design (e.g. for quirked apps or SKUs), so we should always use it when determining whether to behave one way or another to be in line with other behavior that depends on the quirks.

windows/code/RNDeviceInfoCPP.h
```diff
@@ -75,6 +121,53 @@ namespace winrt::RNDeviceInfoCPP
...
+    void isMouseConnected(ReactPromise<bool> promise) noexcept
+    {
+      promise.Resolve(isMouseConnectedSync());
+    }
+
+    REACT_METHOD(isTabletMode);
+    void isTabletMode(ReactPromise<bool> promise) noexcept
+    {
+      m_reactContext.UIDispatcher().Post([promise]() {
+        auto view = winrt::Windows::UI::ViewManagement::UIViewSettings::GetForCurrentView();
```
asklar, line 154: you'll want to caveat this with "if not running inside an islands app" since this will crash outside UWP. In those cases, you'll want to use the APIs I mentioned offline.

windows/code/RNDeviceInfoCPP.h
namrog84: I didn't add windowsphone originally, I was only fixing the incorrect regex return value, but good to know and I will remove it completely.

Me calling it flakey and unrelabie, is exactly what I mean by not getting values you'd expect by design. But I will update the comment to say that instead.

windows/code/RNDeviceInfoCPP.h
namrog84: What is an Islands app, and I will look into adding those others for isTabletMode as well.  
When is RNW not running inside UWP?

asklar: XAML Islands (aka Desktop bridge) is a set of technologies that allow win32 apps (i.e. non-appcontainer) to have XAML UI in them; see [this link](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/xaml-islands).
For RNW, it is important that we support this scenario for certain consumers (as in, things might dim down but should not crash), ideally we respond to the right win32-style events via window messages instead, for example. There are helpers to subscribe to window messages from a native module if you `#include <DesktopWindowBridge.h>`  in RNW, there's an example of it in the inbox [DeviceInfo module](https://github.com/microsoft/react-native-windows/blob/fe6d6a12c143267eb2a84a89cc417b1c43221006/vnext/Microsoft.ReactNative/Modules/DeviceInfoModule.cpp#L69-L84)

We have a Win32/RNW island app in the repo in `\packages\playground\windows\playground-win32.sln`

namrog84: So all of this is completely new and foreign to me, first time hearing of any of it.

So what I need to do here is add a 
`if (IsWinUI3Island())` and this particular UI:ViewManagement I can't use in an island?
And then use the other APIs otherwise?

asklar: yes that's right. There are helpers to detect when you're in an island, see [helpers.h](https://github.com/microsoft/react-native-windows/blob/fe6d6a12c143267eb2a84a89cc417b1c43221006/vnext/Microsoft.ReactNative/Utils/Helpers.h#L33)

namrog84:
PR is in an incomplete/broken state as of this comment. Will be updating in coming days.

namrog84:
Made some more tweaks.
Everything seems to be working fine now in the /example app. But per asklar's suggestion, I will also try it out in the playground-win32app as well tomorrow.

namrog84:
I have officially tested my PR in the included example app,  our own internal Win10 UWP app, the RNW XamlIsland app, and a internal Xbox Console app.  

I am good to go with PR in it's current state.


SideNote: Discovered a few other existing functions don't work in XamlIsland app, but now with the IsXamlIsland(), it'd be easier to address those when those relevant folks want to make those changes in the future.

mikehardy:
Fantastic! Let's test the release bot 😁

mikehardy (approved):
Lgtm - even included docs and example addition thank you!

windows/code/RNDeviceInfoCPP.h
```diff
@@ -75,6 +153,79 @@ namespace winrt::RNDeviceInfoCPP
...
+      promise.Resolve(isMouseConnectedSync());
+    }
+
+    // Helper function, should eventually be removed/replaced when it's exposed through RNW's public API.
+    bool IsXamlIsland() {
+      AppPolicyWindowingModel e;
+      // This value comes from an inline function GetCurrentThreadEffectiveToken() from processthreadsapi.h
+      auto token = (HANDLE)(LONG_PTR) -6;
+      auto windowingModel = CALL_INDIRECT(L"Api-ms-win-appmodel-runtime-l1-1-2.dll", AppPolicyGetWindowingModel, token, &e);
+      if (FAILED(windowingModel) || e == AppPolicyWindowingModel_ClassicDesktop) {
```
asklar, line 188: can't do FAILED here since AppPolicyGetWindowingModel doesn't return an HRESULT sadly
Check for != ERROR_SUCCESS instead

windows/code/RNDeviceInfoCPP.h
```diff
@@ -75,6 +153,79 @@ namespace winrt::RNDeviceInfoCPP
...
+      return false;
+    }
+
+    REACT_METHOD(isTabletMode);
+    void isTabletMode(ReactPromise<bool> promise) noexcept
+    {
+      if (IsXamlIsland()) {
+        // If the function succeeds, the return value is the requested system metric or configuration setting.
+        // If the function fails, the return value is 0. GetLastError does not provide extended error information.
+        auto ret = CALL_INDIRECT(L"ext-ms-win-ntuser-sysparams-ext-l1-1-0.dll", GetSystemMetrics, SM_TABLETPC);
```
asklar, line 200: maybe we want SM_DIGITIZER here, I think SM_TABLETPC might always be false for win10 (I don't have a tablet to try this on at the moment)

asklar (approved):
thanks Adam for all the hard work on this!
left a couple of comments but otherwise lgtm

mikehardy:
Will wait for those but otherwise it's just ci for me after

windows/code/RNDeviceInfoCPP.h
namrog84: Updated!

windows/code/RNDeviceInfoCPP.h
namrog84: Updated!

namrog84:
Updated the 2 last PR suggestions.  Reran example, and all looks good to me now.

asklar (approved):

mikehardy:
Mmm  looks like the `yarn install` step is failing while building transpiling the typescript from the added types - it looks like something fundamental (so all CI items fail) but also small - can you take a look when you get a chance @namrog84 ? Thanks!

namrog84:
I've been working out of modifying node_modules/react-native-device-info and copying back into the src as I tested things. And forgot to copy back some index.ts change. oops!


Why you might ask, work in reverse?  The current dev flow has you use dev-sync which uses `cp -r` which isn't a valid windows copy command.  Early on, I had built some custom command for a an equivalent windows dev-sync, and then just forgot and found it easier to work the other way around. (node_modules back to root). Then computer rebooted and lost it :| 

I've yet to come across any project that handles the dev sync between root module project and example/app project work extremely well and gracefully, and I've seen a bunch of different approaches :|

mikehardy:
This could be a whole lot better dev experience. react-native-firebase uses a metro.config.js to shim at the bundler file load layer via a remap of it's module file paths back up to source and it works well on unix but I'm not sure for windows? Not sure if react-native-windows uses metro or not even!

I give that a try just to learn what we can - I have a windows host,

There are javascript analogs that could be cross-platform for dev-sync at least though, having a unix-ism in there for something so simple as a cp -r is a bit embarrassing on our side. That's something I can easily look at short term

mikehardy:
It's still failing in CI though unfortunately - shame I have to hit the button each time on first contributors, so I just invited you as a contributor in the hope that will allow running PR Actions now more iteratively to get this across the line

asklar:
yes, we use metro in windows too : )
another way to do this would be to yarn link the example app's dependency (`yarn link ` in the package folder, then `yarn link react-native-device-info` in the app's folder)

windows/code/RNDeviceInfoCPP.h
```diff
@@ -58,6 +119,23 @@ namespace winrt::RNDeviceInfoCPP
         return false;
       }
     }
+	
+    REACT_SYNC_METHOD(getDeviceTypeSync);
+    std::string getDeviceTypeSync() noexcept
```
lvalentine, line 124: Since you're in here, how about adding a device type for console?

lvalentine: if (winrt::Windows::System::Profile::AnalyticsInfo::VersionInfo().DeviceFamily() == L"Windows.Xbox")
{
   return "TV";
}

namrog84: Since I'll need to add this to
`export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'unknown';`

Should this be a bit more generic like `Console` or `GameConsole`.

Since using "Xbox" here doesn't really feel like it aligns with the other device types?

namrog84: Talked offline, updated to GameConsole.

src/internal/types.ts
```diff
@@ -1,4 +1,4 @@
-export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'unknown';
+export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'GameConsole' | 'unknown';
```
asklar, line 1:
```suggestion
export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'GamingConsole' | 'unknown';
```

Gaming instead of game?

namrog84: That works for me. Updated!

windows/code/RNDeviceInfoCPP.h
mikehardy: I like that. It's not really TV it does imply a specific controller type, and these types are the high-level sort-of-bundle of expected features so seems right

src/internal/types.ts
```diff
@@ -1,4 +1,4 @@
-export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'unknown';
+export type DeviceType = 'Handset' | 'Tablet' | 'Tv' | 'Desktop' | 'GamingConsole' | 'unknown';
```
mikehardy, line 1: Adding that type means an update to README as well for return type though. Ugh - shame I don't have typedoc set up on this project. Another day - simple update

mikehardy (approved):
still +1 on everything but now with note to add new GamingConsole possible type to getDeviceType return value in README

src/internal/types.ts
namrog84: Updated readme

mikehardy:
Somewhere in this commit train Windows CI started failing 🤔 
https://github.com/react-native-device-info/react-native-device-info/actions/workflows/windows-app-test.yml

namrog84:
I think I might have a local repro,  for some reason I get some weird ReactNative invariant violation with 2x RCTView views being registered in some situations on 'first launch'.  Seems completely unrelated to my changes, but ill give it a closer look this coming week.

mikehardy:
It's the PR that just won't go down quietly ;-). I really appreciate it - I think it's just the windows side of this house hasn't seen a lot of attention so many areas of it may be bit-rotten

asklar:
> It's the PR that just won't go down quietly ;-). I really appreciate it - I think it's just the windows side of this house hasn't seen a lot of attention so many areas of it may be bit-rotten

The "Window has been unexpectedly closed" kind of error usually happens if the app crashes.
@namrog84 you can set up to capture crash dumps using the LocalDumps regkey, see how we do it for RNW tests in [SetupLocalDumps.cmd](https://github.com/microsoft/react-native-windows/blob/master/.ado/scripts/SetupLocalDumps.cmd).
Then you can upload the crash dump as an artifact in github actions so you can download it once it fails and inspect what went wrong.

mikehardy:
@asklar just curious - this sounds like it should go in as a standard on pretty much any RNWindows Github CI run, no?

Enable here https://github.com/react-native-device-info/react-native-device-info/blob/ff8f672cb08fa39a887567d6e23e2f08778e8340/.github/workflows/windows-app-test.yml#L67

And then a final stanza that does something like https://github.com/invertase/react-native-firebase/blob/3fef777f1e08c2dacfb21f4ef2c27b71e8b973f4/.github/workflows/tests_e2e_android.yml#L201-L211 ?

(note the compression may be pointless, in fact in that specific workflow it results in a zipped zip which is silly, but haven't fixed it yet because in practice it's not so bothersome)

asklar:
> @asklar just curious - this sounds like it should go in as a standard on pretty much any RNWindows Github CI run, no?
> 
> Enable here
> 
> https://github.com/react-native-device-info/react-native-device-info/blob/ff8f672cb08fa39a887567d6e23e2f08778e8340/.github/workflows/windows-app-test.yml#L67
> 
> And then a final stanza that does something like https://github.com/invertase/react-native-firebase/blob/3fef777f1e08c2dacfb21f4ef2c27b71e8b973f4/.github/workflows/tests_e2e_android.yml#L201-L211 ?
> 
> (note the compression may be pointless, in fact in that specific workflow it results in a zipped zip which is silly, but haven't fixed it yet because in practice it's not so bothersome)

this can be applied to any windows app, but yes, it'd be useful for RNW test app CI scenarios :)

namrog84:
Just wanted to post an update, I haven't forgotten about this PR. I've just have had other work prioritized, will circle back to investigating the build break when I have time.

mikehardy:
Thanks for the update - I was curious :-). I had out-of-town visitors then an SSD failure so was out of action myself but I'm back now and can collaborate if you need any help

namrog84:
Sorry, definitely going to still finish this.  Just lots of other things have taken higher priority and haven't had a chance to circle back to this yet.

namrog84:
Looks like the 2 failures as the same as another PR.
https://github.com/react-native-device-info/react-native-device-info/pull/1311

Which has to do with some other things unrelated to either PR.  (e.g. windows cert expiry?)

namrog84:
I never touched anything android related and now it seems to be missing a toolchain or something?
I am at a loss on this one. 

The last this this particular build gate passed was 9 days ago. I might just wait and see if someone else's PR/gate is hitting this too and figure out if its a my branch problem or a build machine issue.




FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:stripDebugDebugSymbols'.
> No toolchains found in the NDK toolchains folder for ABI with prefix: arm-linux-androideabi

* Try:
Run with --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:stripDebugDebugSymbols'.

Caused by: java.lang.RuntimeException: No toolchains found in the NDK toolchains folder for ABI with prefix: arm-linux-androideabi

namrog84:
https://github.com/actions/virtual-environments/issues/4144

namrog84:
Aha! Finally got all the gates passing @mikehardy !

Looks like now we can wait on more approvals. 

Hi @machour @GantMan @Johan-dutoit,  I am ready for y'all to take a look since it looks like I need your approves. Let me know if you have any questions or comments.

Thanks!

mikehardy:
I'm actually sufficient to merge+release, looks like it was a real grind for unrelated reasons (crypto key on windows build, that entropic Android NDK failure) - I really appreciate your shoring those up to get it working. Putting this on my review queue and if it's good we can go!

namrog84:
Ah thanks! didn't know that! Apologies to those that got pinged unnecessarily! 

No problem! Let me know if you have any questions or comments! :)

windows/code/RNDeviceInfoCPP.h
```diff
@@ -58,6 +83,27 @@ namespace winrt::RNDeviceInfoCPP
...
+    {
+      if (isTabletHelper()) {
+        return "Tablet";
+      }
+      else if (winrt::Windows::System::Profile::AnalyticsInfo::VersionInfo().DeviceFamily() == L"Windows.Xbox")
+      {
+        return "GamingConsole";
```
mikehardy, line 95: I'm not sure if it's possible now or ever will be, but it would be cool to know what the app was if this code actually shipped in a context where getDeviceType === "GamingConsole" :-) :video_game:

mikehardy (approved):
I think this looks great. I have been meaning to bring up  react-native-windows VM on my main dev rig and I used this as an excuse to do so. :zap: 

image


Everything worked

mikehardy:
This repo auto-releases so the version with this should ship shortly, I want to say (again I think) that I really appreciate that the windows side of things has a working example + CI again now, I've never done react-native-windows but even with my inexperience I was able to get it set up and running pretty well.

The only thing I stumbled on was our "dev-sync" script doesn't work well, the easy way out might be a "dev-sync-windows" that uses a series of xcopy commands to do lib and windows and package.json ?

namrog84:
@mikehardy It is sorta possible now! 
These are the public facing versions of an internal app that could be used in used in both places.  
Console App mentioned here:
https://news.xbox.com/en-us/2021/09/28/xbox-cloud-gaming-beta-on-console-begins-releasing-to-insiders-starting-today/
Windows PC App mentioned here:
https://www.xbox.com/en-US/apps/xbox-app-for-pc

for more info:
https://www.xbox.com/en-US/xbox-game-pass/cloud-gaming
Top 1 is browser, but bottom 3 native apps are powered by cross platform code and react-native. https://i.imgur.com/8YYflIe.png

namrog84:
Also, for the dev-sync.  I had looked around a little at some cross platform npm packages, didn't find anything I particularly liked a lot, but ended up just installing cygwin (https://www.cygwin.com/)(adds the 'cp' and other unix commands support to windows).

But it would be great to have something that works a bit better cross platform that doesn't require developers to install anything additional.  I will keep an eye out for any good cross-platform solutions.  


Worse case: the dev-sync could always be moved into a helper script that does a platform check and does cp on mac/linux and something else on windows too.

asklar:
> Also, for the dev-sync. I had looked around a little at some cross platform npm packages, didn't find anything I particularly liked a lot, but ended up just installing cygwin (https://www.cygwin.com/)(adds the 'cp' and other unix commands support to windows).
> 
> But it would be great to have something that works a bit better cross platform that doesn't require developers to install anything additional. I will keep an eye out for any good cross-platform solutions.
> 
> Worse case: the dev-sync could always be moved into a helper script that does a platform check and does cp on mac/linux and something else on windows too. I know I've seen a bunch of different approaches, and it's a lot my team's preferred way of handling it. To not rely on multiple package.json commands and not rely on misc npm cross plat packages. (e.g. `"dev-sync": "node scripts/dev-sync.js"` ) and scripts/dev-sync.js have something like `const isWindows = process.platform === 'win32';` that decides to use appropriate commands.

FWIW powershell works cross-plat :)

namrog84:
We do personally use a lot of powershell as well for crossplat, but unfortunately it falls into the 'requires non windows' machines to have installed PowerShell Core (or is it just called PowerShell 7 now?) for that they might not already have installed.   

So we've actually been migrating away from it for that reason :(  Even though I personally prefer it over node js script code.

mikehardy:
there actually is a copyfiles javscript module that looked serviceable, but I was focusing mostly on "run the thing" vs "perfect the package.json" so I didn't suss it out too far. Given this a javascript package (in the main, with native bits but the point is javascript...) finding a javascript solution seems best

mikehardy:
I also know there are ways you can fake things and intermediate etc in react-native-config.js + metro config + package.json to just directly reference the files but I must admit I find that even more distasteful then a file copy. I consider that style "fancy" and though I work with it in a couple other repos I maintain and it *does* work it is easily broken, very very few people understand it (because no normal apps do it) and it makes the example apps even more bespoke then they are so my goal of programmatic version refresh on examples is even harder.

Kind of funny that this is the sort of problem (having an example app reference module code under development) becomes a sticking point

PR #1263, opened 2021-07-12
fix: make getUsedMemory more inline with iOS #895

### LINKED ISSUE #895: Improve used memory reporting
# Feature Request

Both iOS and Android used memory seem quite far away from their real values. On Android, it looks like it is only counting the JS used memory. On iOS, it is not that bad; however, it doesn't seem to be counting memory from Core Foundation (explanation below).

We've just discovered a memory leak due to some CF objects not being released. Oddly enough, this library was reporting "normal" memory usage, and it would even decrease the higher the memory leak was! This lead us to think memory was not an issue initially.

## Why it is needed

More accurate values could be used for improved debugging.

## Possible implementation
Here's a Swift answer that claims to be "better" than the accepted one: https://stackoverflow.com/questions/787160/programmatically-retrieve-memory-usage-on-iphone

### Code sample
-

tbezman:
## Description

My company is logging some metrics around memory usage and we've found there is a large discrepancy between iOS and Android. Android is about an order of magnitude lower. 

This is more inline with what's happening in iOS. Android was previously only getting the used bytes from the RN Runtime,
not the entire app.

The Android example app went from reporting ~4mb => ~80mb which is way more inline with what I'd expect.

**Do I need to bump the version in package json or will you take care of that on the next release?**

Fixes #895 

## Checklist

Some of the checklist items don't apply since I didn't change any APIs.

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -20,6 +20,8 @@ import android.os.Environment;
 import android.os.PowerManager;
 import android.os.StatFs;
 import android.os.BatteryManager;
+import android.os.Debug;
+import android.os.Process;
 import android.provider.Settings;
 import android.webkit.WebSettings;
 import android.telephony.TelephonyManager;
@@ -454,11 +456,26 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   public void isBatteryCharging(Promise p) { p.resolve(isBatteryChargingSync()); }
 
   @ReactMethod(isBlockingSynchronousMethod = true)
-  public int getUsedMemorySync() {
-    Runtime rt = Runtime.getRuntime();
-    long usedMemory = rt.totalMemory() - rt.freeMemory();
-    return (int)usedMemory;
+  public double getUsedMemorySync() {
+    ActivityManager actMgr = (ActivityManager) getReactApplicationContext().getSystemService(Context.ACTIVITY_SERVICE);
+    if (actMgr != null) {
+      int pid = android.os.Process.myPid();
+      android.os.Debug.MemoryInfo[] memInfos = actMgr.getProcessMemoryInfo(new int[]{pid});
+
+      if(memInfos.length != 1) {
+        System.err.println("Unable to getProcessMemoryInfo. getProcessMemoryInfo did not return any info for the PID");
+        return -1;
+      }
+
+      android.os.Debug.MemoryInfo memInfo = memInfos[0];
+
+      return memInfo.getTotalPss() * 1024D;
+    } else {
+      System.err.println("Unable to getProcessMemoryInfo. ActivityManager was null");
+      return -1;
+    }
   }
+
   @ReactMethod
   public void getUsedMemory(Promise p) { p.resolve(getUsedMemorySync()); }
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -454,11 +455,13 @@ public boolean isBatteryChargingSync(){
   public void isBatteryCharging(Promise p) { p.resolve(isBatteryChargingSync()); }
 
   @ReactMethod(isBlockingSynchronousMethod = true)
-  public int getUsedMemorySync() {
-    Runtime rt = Runtime.getRuntime();
-    long usedMemory = rt.totalMemory() - rt.freeMemory();
-    return (int)usedMemory;
+  public double getUsedMemorySync() {
+    Debug.MemoryInfo memInfo = new Debug.MemoryInfo();
```
mikehardy, line 459: Interesting! First, thanks for looking into this, I was unaware there was a discrepancy in here, no one has mentioned it before!

Have you considered the warning here? https://developer.android.com/reference/android/os/Debug#getMemoryInfo(android.os.Debug.MemoryInfo)

> Note: this method directly retrieves memory information for the given process from low-level data available to it. It may not be able to retrieve information about some protected allocations, such as graphics. If you want to be sure you can see all information about allocations by the process, use ActivityManager.getProcessMemoryInfo(int[]) instead.

It appears this is the recommended way to make sure you really get accurate info: https://developer.android.com/reference/android/app/ActivityManager#getProcessMemoryInfo(int[])

Appears you can easily get the PID like so https://developer.android.com/reference/android/os/Process#myPid() ?

mikehardy (commented):
Hi there - I have one comment on API choice, and in answer to your question we have semantic release here so as long as the commit is conventional (something I can actually do on merge via a squash-merge) it will be version-bumped, changelog-generated and released immediately

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
tbezman: @mikehardy I'm not 100% if that API is better or not. It notes some restrictions around rate limits and how it's only intended for debugging 

> Note: this method is only intended for debugging or building a user-facing process management UI.

I feel like the API I'm using should be sufficient.

I'm very open to feedback and it is your repo so say the magic words and I will swap it out. Thanks for the feedback!

mikehardy: I see what you mean... And yet I think graphics is a frequent consumer of lots of memory. Also I don't think people will call this much? I usually lean towards exposing all the info possible so I think the more thorough API is better, but maybe a Readme update with a link to the android API and the warning text? So developers know there may be a real cost.

mikehardy: @tbezman This is not meant to pressure: I'm just sweeping my mailbox of all notifications and I'd saved this one. Just to be clear this is waiting on swapping out the API for the one that contains graphic memory as well, along with maybe a readme note containing the admonition from API docs and a pointer to them so devs can make their tradeoff decisions. Then I'll merge it and it'll auto-release and be amazing :sweat_smile:

tbezman: Hey @mikehardy. Just now getting back to this. I'll have a commit up soon!

mikehardy (approved):
LGTM! pending CI

mikehardy:
Fingers crossed, semantic-release should do it's thing and this will exist as a release in a moment.

PR #1297, opened 2021-08-27
provide constants for deviceType

### LINKED ISSUE #1296: convert some strings into enums
## Description

Right now `DeviceType` and `BatteryState` are just strings, it would be nice to have these as enums or even just provide constants linked to the current strings. This would let us replace something like:

```
switch (getDeviceType()) {
  case 'Desktop':
    // desktop stuff here.
    break
  case 'Handset':
    // handset stuff here.
    break
  case 'Tablet':
    // tablet stuff here.
}
```

with:

```
switch (getDeviceType()) {
  case DeviceType.DESKTOP:
    // desktop stuff here.
    break
  case DeviceType.HANDSET:
    // handset stuff here.
    break
  case DeviceType.TABLET:
    // tablet stuff here.
}
```

## Use Case

This would increase type safety and reduce errors in case someone fat-fingers these strings in their code.

chrisdrackett:
## Description

Fixes #1296

This is a quick pass as I don't have the time right now to figure out how this package is built. Hopefully I'll have some free time next week to look into it if nobody else is able to take a look.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -563,8 +563,18 @@ export const [isAirplaneMode, isAirplaneModeSync] = getSupportedPlatformInfoFunc
   defaultValue: false,
 });
 
-export const getDeviceType = () => {
-  if (Platform.OS === 'windows') return 'Desktop';
+export const deviceType = {
+  DESKTOP: 'Desktop',
+  HANDSET: 'Handset',
+  TABLET: 'Tablet',
+  TV: 'Tv',
+  UNKNOWN: 'unknown'
+} as const
+
+type DeviceTypeReturnType = typeof deviceType[keyof typeof deviceType]
+
+export const getDeviceType = (): DeviceTypeReturnType => {
+  if (Platform.OS === 'windows') return deviceType.DESKTOP;
 
   return getSupportedPlatformInfoSync({
     memoKey: 'deviceType',
@@ -574,8 +584,8 @@ export const getDeviceType = () => {
   });
 };
 
-export const getDeviceTypeSync = () => {
-  if (Platform.OS === 'windows') return 'Desktop';
+export const getDeviceTypeSync = (): DeviceTypeReturnType => {
+  if (Platform.OS === 'windows') return deviceType.DESKTOP;
 
   return getSupportedPlatformInfoSync({
     memoKey: 'deviceType',


src/index.ts
```diff
@@ -563,8 +563,18 @@ export const [isAirplaneMode, isAirplaneModeSync] = getSupportedPlatformInfoFunc
...
-  if (Platform.OS === 'windows') return 'Desktop';
+export const deviceType = {
+  DESKTOP: 'Desktop',
+  HANDSET: 'Handset',
+  TABLET: 'Tablet',
+  TV: 'Tv',
+  UNKNOWN: 'unknown'
+} as const
+
+type DeviceTypeReturnType = typeof deviceType[keyof typeof deviceType]
```
schie, line 574: Try something like this:

```ts
export enum DeviceTypeReturnType {
    DESKTOP = 'Desktop',
    HANDSET = 'Handset',
    TABLET = 'Tablet',
    TV = 'Tv',
    UNKNOWN=  'unknown'
}
```

personally, it's a lot cleaner / easier to understand

schie (changes requested):

src/index.ts
mikehardy: Oh my - I had not realized that this had gone stale, my apologies @chrisdrackett - I agree in general with @schie's comment, and just wanted to add that index.ts has a conflict now but it should be easy to resolve, I think it's just the addition of windows support for device type, and the new possible 'GamingConsole' return value (which should be added to this PR's set of possible values) from #1238

chrisdrackett: I totally forgot about this, I'm all for adding an enum as well, so hopefully I can find time to update this if someone else does not get to it first (we're in the middle of a pretty big launch so things like this aren't on my radar atm.)

schie: @mikehardy would you consider this a breaking change?

schie: I think "no" because, I believe you can still compare the strings

mikehardy: @schie I could only definitively say with testing, sorry - while I basically insist on using typescript where ever I can, I'm not an expert (yet?)

chrisdrackett: this should not be a breaking change as outside of TS these are still the same strings

mikehardy: if someone's typescript build breaks, it's a breaking change
(note: breaking changes aren't evil, we can release them no problem, we just want to make sure the version increment  + changelog correctly communicates the changes developers need to look out for)

mikehardy:
doing some janitor work - can reopen if interested, but appears abandoned, or just slow communication (happy to reopen if so)

PR #1302, opened 2021-09-10
docs(README): add proguard rule note for hasGms android release build issue

### LINKED ISSUE #1300: hasGms returns as false on Huawei devices which have GMS pre-installed
## Summary

|             |     |
| ----------- | --- |
| Device     | Huawei Mate 10 |
| Version     | 8.0.0   |
| Affected OS | Android   |
| OS Version  | 10 |

## Current behavior
hasGms returned as false

## Expected behavior
hasGms should return true for one of my user's Huawei device as it does have gms pre-installed.

## Additional notes
The user's device Play Store services version is 21.30.16

An investigation into this would be much appreciated!

ws8149:
## Description

Fixes #1300

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ If you want to use Install Referrer tracking, you will need to add this config t
 }
 ```
 
+If you are experiencing issues with hasGms() on your release apks, please add the following rule to your Proguard config
+```
+-keep class com.google.android.gms.common.** {*;}
+```
+
 ## AndroidX Support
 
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Thank you very much for this @ws8149 ! These debug/release build differences can be quite difficult to figure out for anyone that doesn't have deep Android experience, this should really help people. I'll leave the related issue open for deeper thinking in case there is something more automated we may be able to do for people

PR #1099, opened 2020-10-15
add iphone 12 info

stevelyc:
add iPhone 12 device info

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅  |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
* [x] I added the documentation in `README.md`
* [x] I mentioned this change in `CHANGELOG.md`
* [x] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [x] I added a sample use of the API (`example/App.js`)

PR #1307, opened 2021-09-15
fix(hasNotch): Add iPhone 13 lineup to list of devices with notch

rolfb:
## Description

iPhone 13 lineup is about a week out, so adding this to the device list for devices with notch.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

### DIFF
--- a/src/internal/devicesWithNotch.ts
+++ b/src/internal/devicesWithNotch.ts
@@ -3,6 +3,22 @@ import { NotchDevice } from './privateTypes';
 const devicesWithNotch: NotchDevice[] = [
   {
     brand: 'Apple',
+    model: 'iPhone 13 mini',
+  },
+  {
+    brand: 'Apple',
+    model: 'iPhone 13',
+  },
+  {
+    brand: 'Apple',
+    model: 'iPhone 13 Pro',
+  },
+  {
+    brand: 'Apple',
+    model: 'iPhone 13 Pro Max',
+  },
+  {
+    brand: 'Apple',
     model: 'iPhone 12 mini',
   },
   {


rolfb:
Just out of curiosity, what is following Conventional Commits useful for? I sense a learning opportunity.

schie:
> Just out of curiosity, what is following Conventional Commits useful for? I sense a learning opportunity.

@rolfb checkout [commitizen/cz-cli](https://github.com/commitizen/cz-cli)

mikehardy (approved):
Nice! Thanks

Conventional Commits, when followed strictly (a thing we enforce with a check here as you have found :-) ), allow you to infer the type of semantic version release (patch, feature, or breaking change) that is represented by a group of commits, which then allows you to automate releases completely when PRs are merged, via a github action that scans commit logs for conventional commits, infers the type of release and makes the correct version bump

You'll see it in action here - as soon as I merge this it'll do a patch release

rolfb:
@mikehardy Appreciate the explanation, and thank you for the quick merge and release :)

hiteshSavkare:
Hi @mikehardy
Just curious to know don't we require the IOS file changes for notch support? 
actually i am testing it on IOS 13 but always  returning me false 

I just found the reference here  how it was for IOS 12, there i can see IOS file changes 
https://github.com/react-native-device-info/react-native-device-info/pull/1099/files

rolfb:
Seems like I missed a file 🥺

hiteshSavkare:
> Hi @mikehardy && @rolfb
> Just curious to know don't we require the IOS file changes for notch support?
> actually i am testing it on IOS 13 but always returning me false
> 
> I just found the reference here how it was for IOS 12, there i can see IOS file changes
> https://github.com/react-native-device-info/react-native-device-info/pull/1099/files

seems like that, hope you will add that file soon, eagerly waiting for testing , thanks for your efforts

rolfb:
@hiteshSavkare can you have a look at #1309 and see if that fixes things for you?

hiteshSavkare:
> @hiteshSavkare can you have a look at #1309 and see if that fixes things for you?

@rolfb yes perfect, working as expected

rolfb:
> @rolfb yes perfect, working as expected

@hiteshSavkare great! if you have a chance to verify that all four iPhone 13 devices are correctly mapped now I'd appreciate it 😊 I couldn't find official docs for firmware identifiers but it's highly likely that they are following the same pattern.

PR #1333, opened 2021-10-28
fix(web): Illegal invocation from getBattery call

### LINKED ISSUE #1223: Illegal invocation from getBattery call
Not 100% sure if this is actually an issue within this project, but though I would ask if someone happens to have some information on this.
## Summary
Affected platform: web
|             |     |
| ----------- | --- |
| Version     | 6.2.1   |
| Affected OS | all   |
| OS Version  |  No pattern  |

## Current behavior
Some users apparently receive `TypeError: Failed to execute 'getBattery' on 'Navigator': Illegal invocation.`
This is coming from `react-native-device-info/lib/module/web/index.js in init`. 
I did some research and error would indicate that the call is missing context (for navigator object), but based on code I have no idea how this could be. Perhaps the call should be prefixed with `window.`?
So far we have had 156 users getting this error, so it seems pretty common although it does not happen to me. 
Screenshot 2021-04-21 at 9 24 22

## Expected behavior
If getBattery call is found from navigator, call should not give error and no call should happen if method is not supported.

antoinewg:
## Description

Fixes #1223 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [ ] ~I added the documentation in `README.md`~
* [ ] ~I updated the typings files (`privateTypes.ts`, `types.ts`)~
* [ ] ~I added a sample use of the API (`example/App.js`)~

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -133,7 +133,7 @@ export const isBatteryChargingSync = () => {
 };
 
 export const isCameraPresent = async () => {
-  if (navigator.getBattery) {
+  if (navigator.mediaDevices && navigator.mediaDevices.enumerateDevices) {
     return navigator.mediaDevices.enumerateDevices().then(devices => {
       return !!devices.find((d) => d.kind === 'videoinput');
     });
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -123,7 +123,7 @@ export const getUserAgent = async () => {
 
 export const isBatteryCharging = async () => {
   if (navigator.getBattery) {
-    return navigator.getBattery().then(battery => battery.level);
+    return navigator.getBattery().then(battery => battery.charging);
   }
   return false;
 };
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -55,10 +55,10 @@ export const getUsedMemorySync = () => {
   return -1;
 };
 
-const init = async () => {
-  if (typeof navigator !== 'undefined' && navigator.getBattery) {
-    const battery = await navigator.getBattery();
+const init = () => {
+  if (typeof navigator === 'undefined' || !navigator.getBattery) return;
 
+  navigator.getBattery().then((battery) => {
     batteryCharging = battery.charging;
 
     battery.addEventListener('chargingchange', () => {
@@ -81,7 +81,7 @@ const init = async () => {
         deviceInfoEmitter.emit('RNDeviceInfo_batteryLevelIsLow', level);
       }
     });
-  }
+  });
 };
 
 const getBaseOsSync = () => {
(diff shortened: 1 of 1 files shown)


mikehardy:
Well, no testers showed up :-)
I'd like to keep this API the same as the others though and I think the implication of not having async/await is that the API is subtly different. Is that correct? It may not be correct.

But going with the idea it might be correct - or maybe this is still useful anyway - it reminded me of this where you can manufacture a Promise and thus conform to async/await expectations for the API?

https://github.com/invertase/notifee/blob/4a720dc8ac52506b302f6d9252a99cefc1d8cb28/tests_react_native/specs/notification.spec.ts?#L63-L78

```typescript
      return new Promise(async (resolve, reject) => {
        const unsubscribe = notifee.onForegroundEvent((event: Event) => {
          try {
            if (event.type === EventType.DELIVERED) {
              expect(event.detail.notification).not.equal(undefined);
              if (event.detail.notification) {
                expect(event.detail.notification.title).equals('Hello');
                expect(event.detail.notification.body).equals('World');
              }
              unsubscribe();
              resolve();
            }
          } catch (e) {
            unsubscribe();
            reject(e);
          }
```

mikehardy:
I think this is still valid, and I'm working on react-native-web projects now as a primary thing, just not quite yet into the areas that I need device-info (or netinfo) but they're coming. Going to leave this one open as I think it's valuable for future harvesting

dpott197:
Hi, team! I hope all is well. I'd like to see this issue resolved as well. I tried to reproduce this issue on my local environment, but I had no luck. My app just reports these errors in production. How can we get more testers? Or does it make sense to merge this optimistically?

mikehardy:
> I think this is still valid, and I'm working on react-native-web projects now as a primary thing, just not quite yet into the areas that I need device-info (or netinfo) but they're coming. Going to leave this one open as I think it's valuable for future harvesting

Just staving off the bot - my previous comment still stands but I'm still a bit busy at the moment

mikehardy (approved):
I hate just leaving this sitting here but we've got a lot of preferences and no hard testing reproduction.

Here's what I think: 

- OP states it was tested before and after and works, let's assume that
- @schie states would prefer to get rid of async keyword in signatures
- @mikehardy I have a strong preference for identical APIs across all platforms, and I don't really see a downside to keeping the async

As the person typing this and since it's my preference, I of course lean towards just merging this as is: keeping async, assuming testing was sufficient, merging it and releasing.

That will be my default action in a couple days if I don't hear a good technical argument to the contrary.

I love a good technical argument though, so don't hesitate - just leaning towards "progress" vs "waiting even more"

Cheers

PR #1343, opened 2021-11-12
fix(android): gracefully handle SecurityException in getPhoneNumber

mikehardy:
## Description

For some reason on targetSdkVersion30+ in some cases, you can have the appropriate
permission but still take a SecurityException when you attempt to call getLine1Number

This allows the method to gracefully fallback to "unknown" instead of crashing

This tested directly in an app I'm working on, I can reproduce a crash every time even though
when I probe it, I appear to have READ_SMS permission, right before I get a SecurityException saying...I don't have READ_SMS permission? It's quite odd.

## Checklist


mikehardy:
If this passes CI on android I'm going to merge it for release - it defaults to our default, it is technically an error so we're not completely hiding it (it gets a log message, just in case, but no PII in the message), and it's fixing a crash bug

schie (approved):
LGTM

mikehardy:
If this passes CI on android I'm going to merge it for release - it defaults to our default, it is technically an error so we're not completely hiding it (it gets a log message, just in case, but no PII in the message), and it's fixing a crash bug

Both of the androids failed for a classic "CI flake" reason where the android emulator simply wouldn't install the app. I see that in other repos too. I restarted one of them and it went fine. Compiled fine every time

mikehardy:
Thanks for the double-check, I appreciate it. Launch!

PR #1344, opened 2021-11-13
fix(android, serial): Fix permission access to serial number

### LINKED ISSUE #1320: getSerial is not accessible even if permissions are available
## Summary

| | |
| --  | -- |
| Version     | v8.3.3  |
| Affected OS | Android  |
| OS Version  | API > 26   |

## Current behavior

Access to serial number on android doesn't work even if android.permission.READ_PRIVILEGED_PHONE_STATE is granted.
(no error displayed)
getSerial return "unknown"

## Expected behavior

Serial number should be retrieved

## Link to documentation

https://developer.android.com/reference/android/os/Build#getSerial()
`
If the calling app has been granted the READ_PRIVILEGED_PHONE_STATE permission; this is a privileged permission that can only be granted to apps preloaded on the device.`
That means it can be tested only with an app preloaded in android system image

freeboub:
## Description

Fixes #1320

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

Nothing this is a minor fix ...

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -712,9 +712,9 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   public String getSerialNumberSync() {
     try {
       if (Build.VERSION.SDK_INT >= 26) {
-        if (getReactApplicationContext().checkCallingOrSelfPermission(Manifest.permission.READ_PHONE_STATE) == PackageManager.PERMISSION_GRANTED) {
-          return Build.getSerial();
-        }
+        // There are a lot of conditions to access to getSerial api
+        // For details, see https://developer.android.com/reference/android/os/Build#getSerial()
+        return Build.getSerial();
       } else {
         return Build.SERIAL;
       }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -712,9 +712,9 @@ public String getDeviceNameSync() {
   public String getSerialNumberSync() {
     try {
       if (Build.VERSION.SDK_INT >= 26) {
-        if (getReactApplicationContext().checkCallingOrSelfPermission(Manifest.permission.READ_PHONE_STATE) == PackageManager.PERMISSION_GRANTED) {
-          return Build.getSerial();
-        }
+        // There are a lot of conditions to access to getSerial api
+        // For details, see https://developer.android.com/reference/android/os/Build#getSerial()
+        return Build.getSerial();
```
mikehardy, line 717: Just to make sure I checked to make sure SecurityException was truly a subclass of Exception so this will fall through and return "unknown", it is https://developer.android.com/reference/java/lang/SecurityException

mikehardy, line 717:
```suggestion
        // Rather than check each one, just try and rely on the catch below, for discussion on this approach, refer to
        // https://github.com/react-native-device-info/react-native-device-info/issues/1320
        return Build.getSerial();
```

mikehardy: Unfortunately I can't just commit this and then merge it, I guess you have to click the commit button for it? I just wanted to link it to the discussion

mikehardy (changes requested):
Just want to leave ourselves a breadcrumb for the future but this LGTM - if you could add that comment note (I think it's just a button on the github web UI?) I can merge this and release it.

Thanks!

freeboub:
I hope it should be ok now ?

mikehardy (approved):
yeah, was okay before really :-), I just like leaving future maintainers notes - the maintainer in the future might be you! Leave yourself the gift of understanding, philosophically. Thanks again

PR #1345, opened 2021-11-15
docs: add link to hooks and events

lisabaut:
## Description

This is a mini PR to help people find the docs about hooks and events.
The README is pretty long and if one does not scroll down below the long API section it is very likely that one never knows about the nice hooks of this lib. This happened to me.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ✅    |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Device Information for [React Native](https://github.com/facebook/react-native).
 - [Linking](#linking)
 - [Usage](#usage)
 - [API](#api)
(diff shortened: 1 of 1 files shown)


schie (approved):
You're the hero we don't deserve!

mikehardy (approved):
Nice! It's the little things that make a package nice to use, like this

Issue #1346, opened 2021-11-15
fix: useBatteryIsLow fires on initial render with current battery level


lisabaut:
## Summary

|             |     |
| ----------- | --- |
| Version     | 8.4.4   |
| Affected OS | iOS / Android   |
| OS Version  | ?   |

## Current behavior

When using the hook `useBatteryIsLow` it returns the current battery level on initial rendering- no matter if the battery level is currently low or not according to the defined percents (=> see [docs](https://github.com/react-native-device-info/react-native-device-info#usebatterylevelislow-or-rndeviceinfo_batterylevelislow)).

This is due to these lines here: https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L716-L719

**Please note:**
When running the `example` App on an emulator/simulator one can not detect this bug. 
You need to run the `example` App on a real device with an actual battery to see that `batteryIsLow` is rendered with the current battery level on initial rendering.

## Expected behavior

The hook `useBatteryIsLow` returns the battery level as a number in percent if the current battery level is currently low.
If the battery level is not low the hook returns `null`.

## Solution

I assume simply removing [the lines](https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L716-L719) would fix the issue. What do you think?

mikehardy:
Interesting - I'll admit I'm not so great at hooks, and further I didn't write that one :-), if we remove the lines completely, won't that miss the case where it should fire when the battery actually is low, vs waiting for a battery level change event to trigger evaluation vs the low battery thresholds ?

PR #1347, opened 2021-11-16
fix(useBatteryIsLow): return battery level only if low

lisabaut:
## Description

Fixes [#1346](https://github.com/react-native-device-info/react-native-device-info/issues/1346)

The hook `useBatteryIsLow` returns the current battery level on initial rendering - even if the current battery level is not considered as low (Android < 0.15, iOS < 0.20);

With removing the initial value from the useEffect hook on initial rendering, the returned value remains `null` until the listener `RNDeviceInfo_batteryLevelIsLow`  detects the first callback for a low battery status.

### Testing
On iOS simulator the battery level is always `-1`.
On Android emulator the battery level is always `1`.
At least for me.
I therefore tested it on an external physical device too. See the video in the linked ticket above. 

I did not find dedicated tests for the hooks. Let me know if I missed something.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅    |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -713,16 +713,9 @@ export function useBatteryLevelIsLow(): number | null {
   const [batteryLevelIsLow, setBatteryLevelIsLow] = useState<number | null>(null);
 
   useEffect(() => {
-    const setInitialValue = async () => {
-      const initialValue: number = await getBatteryLevel();
-      setBatteryLevelIsLow(initialValue);
-    };
-
     const onChange = (level: number) => {
       setBatteryLevelIsLow(level);
     };
-
-    setInitialValue();
 
     const subscription = deviceInfoEmitter.addListener('RNDeviceInfo_batteryLevelIsLow', onChange);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -618,6 +618,13 @@ export function hasSystemFeatureSync(feature: string) {
   return false;
 }
 
+export function isLowBatteryLevel(level: number): boolean {
+  if (Platform.OS === 'android') {
+    return level < 0.15;
+  }
+  return level < 0.2;
+}
+
 export const [
   getSystemAvailableFeatures,
   getSystemAvailableFeaturesSync,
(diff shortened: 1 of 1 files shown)


schie (commented):
I, for the most part, agree w/ this change. 

Would adding a check for the initial value behoove us? What I mean is, _'after getting the initial value, then check to see if it is below the threshold'_ ?

schie:
As as aside, this hook seems very weird. From just reading the name of it, I was assuming it would just spit-out a boolean rather than a number, but `¯\_(ツ)_/¯`

schie (approved):

mikehardy (changes requested):
@schie 
> I, for the most part, agree w/ this change.
> 
> Would adding a check for the initial value behoove us? What I mean is, _'after getting the initial value, then check to see if it is below the threshold'_ ?

That was my first take as well, From https://github.com/react-native-device-info/react-native-device-info/issues/1346#issuecomment-969169742

> if we remove the lines completely, won't that miss the case where it should fire when the battery actually is low, vs waiting for a battery level change event to trigger evaluation vs the low battery thresholds ?

So that's 2 immediate concerns registered about the same thing. I believe the API won't feel right unless there is a check on initial loading, .15 for android .20 for ios as noted - the API should trigger if the battery is already low during first use

Once that's in I can merge and release this, and thank you!

@schie I agree the API is a bit wonky but if you think of it from a javascript perspective, null vs the info you want is still false-y and truth-y, which is kind of javascript-y even if "yes, weird..."

lisabaut:
I added a helper function which does the check again if the current battery level is actually low:
- Android < 0.15
- iOS/web < 0.2

If `true` the initial value is set to the state and returned by the hook.
If `false` the default return value remains `null`. 

Please note that testing on iOS simulators is not really helpful as the battery level is always `-1`. 
I did not add an additional check for that (e.g. `return level > 0 && level < 0.2`).
On Android emulators the battery level is always `1`.

Bildschirmfoto 2021-11-17 um 18 50 14

mikehardy (approved):
Interesting, this bit:

>  If false the default return value remains null.
> Please note that testing on iOS simulators is not really helpful as the battery level is always -1.

Collides a little bit with the stated "default" values of the APIs from the README:

> Note that many APIs are platform-specific. If there is no implementation for a platform, then the "default" return values you will receive are "unknown" for string, -1 for number, and false for boolean. Arrays and Objects will be empty ([] and {} respectively).

However, this is the way it was so leaving it that way (null as default here) is a less-breaking change, and at least it maps along with -1 to "false-y" in javascript so it still evaluates the same on iOS simulator.

Kind of a mess but nothing that you did @lisabaut - this should have been a better API from the start, and I was here to review it so that's on me

I think this change handles the initial hook and subsequent events really well now and I appreciate it! Definitely an improvement over the previous behavior

PR #1359, opened 2022-01-02
docs: Add possible ipados return value for getSystemName

### LINKED ISSUE #1358: DeviceInfo.getSystemName() returns `ipados` instead of `ios`
## Summary

|             |     |
| ----------- | --- |
| Version     |  8.4.5  |
| Affected OS | iOS 15.2 (iPadOS 15.2)  |
| OS Version  |  iOS 15.2 (iPadOS 15.2)  |

## Current behavior

Looks like since some changes in Apple's ecosystem, currently iPads with latest OS do return `ipados` instead of `ios` when getting system name with `DeviceInfo.getSystemName()`. 

This might not be a bug from library side, but I'm raising this issue for consideration and for all the folks that are using this library that could run into issues because of this.

## Expected behavior

At least back in the day `DeviceInfo.getSystemName()` was returning `ios` on iPads.

lukebars:
## Description

Fixes #1358

Since iPadOS 15.0 (or iOS 15.0) iPads return "iPadOS" as the system name.
I've added a line in docs that states that.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -881,7 +881,7 @@ Gets the device OS name.
 
 ```js
 let systemName = DeviceInfo.getSystemName();
-// iOS: "iOS" on newer iOS devices "iPhone OS" on older devices, including older iPad's.
+// iOS: "iOS" on newer iOS devices "iPhone OS" on older devices (including older iPad models), "iPadOS" for iPads using iPadOS 15.0 or higher.
 // Android: "Android"
 // Windows: ?
 ```


schie (approved):
LGTM

mikehardy (approved):
this is great, thank you!

PR #1361, opened 2022-01-05
ci: Do not run release and stale flows on forks

friederbluemle:
`release.yml` runs on push to master, and `stale.yml` runs on schedule. Both workflows should **not be run on forked repositories**. The credentials and permissions are missing, resulting in tons of failed runs (including email notifications):

> [fork/react-native-device-info] Run failed: Mark stale issues and pull requests - master (808c1b5)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   release:
+    if: github.event.repository.fork == false
     name: Release
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -7,9 +7,8 @@ on:
 
 jobs:
   stale:
-
+    if: github.event.repository.fork == false
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/stale@v4
       with:


schie (commented):
Thanks!

Do you have a link to the documentation on this?

I'd love to read about what other goodies they have.

friederbluemle:
Not exactly sure what you're referring to.. All this PR does is update the two mentioned GitHub Actions workflows, so that they don't run in forked repositories anymore (and cause errors).

friederbluemle:
General info on the GitHub Actions API etc is available here:
https://docs.github.com/en/rest/reference/actions

schie (approved):
LGTM

schie:
> Not exactly sure what you're referring to.. All this PR does is update the two mentioned GitHub Actions workflows, so that they don't run in forked repositories anymore (and cause errors).

I was just curious where in the documentation you saw the jazz related to `github.event.repository.fork` so I could read up on it. To me, the documentation is difficult to find specific things like that. 

I found it though, so no worries

friederbluemle:
Yes, agree, the documentation is a bit lacking in that sense. I've successfully used this check in several repos. For example, you can see here that the `release` flow was skipped in my fork after I pushed the merged PR to my master branch: https://github.com/friederbluemle/react-native-device-info/runs/4709727033
If (obviously) failed for previous runs: https://github.com/friederbluemle/react-native-device-info/actions/workflows/release.yml

However, I did notice that the `stale` workflow (which has a "schedule" trigger) still ran, and failed. I'm not sure why, maybe it was still using an older version of the workflow file? I'll keep checking.

mikehardy (commented):
already merged but just commenting to say "nice one!", have added this check in a few repos but never did a comprehensive check of all repos I can push to so didn't realize this was happening here. Is frustrating for forks yes :)

friederbluemle:
As I mentioned above, it turned out that for workflows with a `schedule` trigger, this check does not work. The reason is the `github.event` object is not populated the same way as it is in builds started by other triggers. More details here:
https://github.community/t/schedule-workflow-missing-github-event-repository-context/212293
https://github.community/t/github-event-repository-fork-context-does-detect-forks-in-scheduled-jobs/185925

And a fix: #1363

PR #1363, opened 2022-01-09
Update stale workflow fork check

friederbluemle:
Use a different way to check if the `stale` workflow was started in a forked repo.

It turned out the check added in #1361 does not work as expected for workflows with a `schedule` trigger.

The solution is described in this thread:
https://github.community/t/github-event-repository-fork-context-does-detect-forks-in-scheduled-jobs/185925/3

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -7,10 +7,11 @@ on:
 
 jobs:
   stale:
-    if: github.event.repository.fork == false
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/stale@v4
+    - run: echo "IS_FORK=$(curl "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY" | jq -r .fork)" >> $GITHUB_ENV
+    - if: env.IS_FORK == 'false'
+      uses: actions/stale@v4
       with:
         operations-per-run: 100
         ascending: true
@@ -23,5 +24,3 @@ stake please comment and ping a maintainer to get this merged ASAP! Thanks for c
 utomatically'
         exempt-issue-labels: 'Keep Open'
         exempt-pr-labels: 'Keep Open'
-
-


.github/workflows/stale.yml
```diff
@@ -7,10 +7,11 @@ on:
 
 jobs:
   stale:
-    if: github.event.repository.fork == false
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/stale@v4
+    - run: echo "IS_FORK=$(curl "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY" | jq -r .fork)" >> $GITHUB_ENV
+    - if: env.IS_FORK == 'false'
```
mikehardy, line 13: That is so fancy, why not just take known knowledge we have - that is, we *know* the canonical main repo - and use it for a clean check like https://github.com/invertase/firestore-ios-sdk-frameworks/blob/a10077acdf3e9d4fbd006a729d4dac45711c657c/.github/workflows/build.yml#L11 ?

schie: I agree; let's just keep it simple for now.

friederbluemle: Sure, a check like that would work as well.
What's nice about the current (and proposed) workflow is that it is completely repo agnostic - Meaning there is nothing specific to react-native-device-info in it, it can be copied and reused without changes in any other repo.
If you prefer to hard-code the repo org and name in the file instead, I can also do that - Let me know and I'll amend the commit. :+1:

mikehardy: sorry this stalled, I like the idea of a snippet that's reusable across repos, that's a strong point and I'll probably end up copy/pasting this around myself.

mikehardy (approved):
this looks good to me, and is a general solution to the "don't run on forks" problem, so I think it's worth it. Thanks for the patience @friederbluemle

.github/workflows/stale.yml
friederbluemle: Thanks @mikehardy!

Issue #1364, opened 2022-01-11
getPhoneNumber stopped working in Android 12


colbygarland:
## Summary

|             |     |
| ----------- | --- |
| Version     | 8.4.8   |
| Affected OS | Android   |
| OS Version  | 12   |

## Current behavior

`getPhoneNumber` was pulling the device's phone number reliably for Android 10 and 11. Since the update to 12, none of my test devices (Pixel 3A and android emulator from Android Studio) have been pulling the phone number anymore.

## Expected behavior

`getPhoneNumber` should correctly pull the device's phone number from the device.

mikehardy:
Hi there, strange - I don't see anything in the privacy changes for 12 that would do that.
There was this for 11, you may need a different permission request https://developer.android.com/about/versions/11/privacy/permissions

This is corroborated, somewhat, by our out of date example which is still targeting 10 (API 29) and runs just fine with phone number showing on Android 12 emulator.

Please verify your permissions request I think?

Otherwise:
https://stackoverflow.com/help/how-to-ask
https://stackoverflow.com/help/minimal-reproducible-example

...or dig into the relevant java code in node modules and add some `System.err.println("Here I am etc etc");` to figure out what's happening in the control flow - the method's pretty basic, should be easy to catch.

colbygarland:
I did some looking, turns out I needed to add `<uses-permission android:name="android.permission.READ_SMS" />` permission to my manifest, as well as change the permissions request to `PermissionsAndroid.PERMISSIONS.READ_SMS` instead of `PermissionsAndroid.PERMISSIONS.READ_PHONE_STATE`. 

Interestingly enough, the `READ_PHONE_NUMBERS` permission didn't work, it was the `READ_SMS` that worked for me. Confirmed this works on the emulator, and a physical Pixel 3A, both running Android 12.

Closing this as this solves my problem, thanks for looking into this for me @mikehardy ! Appreciate it.

Issue #1105, opened 2020-10-20
getLine1Number() implementation for Android11


firdaussoberi:
## Description
My app crashed today when testing on Android11 device, simulator also shows same crashlog.
I believe this is related to changes in permission behavior starting from Android11 as pointed here:
https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers

![7e3180af-ee0b-4d22-ae38-e4058670b801](https://user-images.githubusercontent.com/8426299/96553009-20aebd00-12e7-11eb-927c-98741ece9d75.jpeg)

Currently, I only include this related permission in my AndroidManifest:
```
    ...
    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
``` 
and then proceed to call DeviceInfo.getPhoneNumber() after requesting PermissionsAndroid.PERMISSIONS.READ_PHONE_STATE from react-native.

I saw in RNDeviceModule.java that both ```READ_SMS``` and ```READ_PHONE_NUMBERS``` are already included in getPhoneNumberSync().
For android11 and higher, I should include also ```READ_PHONE_NUMBERS``` explicitly in my manifest right?
Would appreciate if anyone who tested successfully on Android11 can shed a light on this.

Am using v5.6.5 of the package, RN0.60.4, targetSDK 30.

firdaussoberi:
Updating manifest permission  to
```
    <uses-permission android:name="android.permission.READ_PHONE_STATE" android:maxSdkVersion="29" />
    <uses-permission android:name="android.permission.READ_PHONE_NUMBERS" />
```
solves the issue for targetSDK 30.

Closing issue.

mikehardy:
Interesting - could use a PR for the docs to solve future developer issues :pray:  :smile:

PR #1369, opened 2022-01-20
Add Android 11 permissions for getPhoneNumber

colbygarland:
## Description

This is just adding the new permission to the README for `getPhoneNumber`. Note that I only added the permission that worked for me, there is another permission (linked in #1105) that worked for somebody else. But this was an old ticket, and this solution didn't work for me, so I did not add this permission. If we want, I can definitely add that one too!

Relates to #1364, #1105

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -770,7 +770,8 @@ DeviceInfo.getPhoneNumber().then((phoneNumber) => {
 
 #### Android Permissions
 
-- [android.permission.READ_PHONE_STATE](https://developer.android.com/reference/android/Manifest.permission.html#READ_PHONE_STATE)
+- Android 10 and below: [android.permission.READ_PHONE_STATE](https://developer.android.com/reference/android/Manifest.permission.html#READ_PHONE_STATE)
+- Android 11 and up: [android.permission.READ_SMS](https://developer.android.com/reference/android/Manifest.permission.html#READ_SMS)
 
 #### Notes
 


schie (approved):
LTGM

mikehardy (commented):
I really appreciate this PR, as the old permission will not work, but I admit I am very confused why READ_SMS worked for you, but the read phone numbers one did not? That just seems contrary to the documentation. I don't have time to personally test though, and I believe your actual testing experience. I just worry because the documentation for READ_SMS lists it as "hard restricted"  https://developer.android.com/reference/android/Manifest.permission#READ_SMS

What would make me more comfortable is if the permissions discussion here was a little less prescriptive now and linked back to the actual documentation listing how it is supposed to work: https://developer.android.com/about/versions/11/privacy/permissions#phone-numbers - along with a note that READ_SMS may work as well, if the documented permissions aren't successful.

Basically, refer people to upstream docs, but also include your experience. What do you think?

colbygarland:
@mikehardy okay yeah I totally agree. I was hesitant to even create this PR, since my testing was so different than others.. _but_ if we just link to the upstream docs, and I add a disclaimer about what worked for me, that might be the best solution. With these "dangerous" permissions, we should probably put more information then less, since they are not only hard to get users to accept, but also get accepted by the App stores.

mikehardy (approved):
I think this is the best we (that is: you, thanks for the PR :-) ) can do. It's not great to just expose the mess to users, but there's not a better way I think, especially with permissions that are tricky like this (and require disclosure on the store listing and everything). LGTM

PR #1372, opened 2022-01-25
feat(ios): Add isMac() function

### LINKED ISSUE #1371: isTablet() returns false if running on M1 Macs
## Summary

As you may know, Macs with Apple Silicon chips can run any iOS app from the App Store. When running on a Mac, the app is shown in a tablet-sized window. However, the isTablet() function from this library returns false, so apps like mine look a bit weird.

|             |     |
| ----------- | --- |
| Version     | 8.4.7   |
| Affected OS | macOS   |
| OS Version  | 12.1   |

## Current behavior

isTablet() returns false

## Expected behavior

I'm honestly not sure. For me it makes sense for isTablet() to return true, as the view is the same size as an iPad. However, it isn't really a tablet. So it would be interesting to hear what you think. It is possible to check if it is an iOS app is running on a Mac using the NSProcessInfo class, so we could add a new function to this library to check for M1 Macs, and then people can implement the "isTablet or M1 Mac" logic themselves.

filiphorvei:
## Description

Fixes #1371

Added a new function to check if we are running an iOS app on an Apple Silicon Mac. I also implemented it on Android and Windows, where it always returns false. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ The example app in this repository shows an example usage of every single API, c
 | [isPinOrFingerprintSet()](#ispinorfingerprintset)                 | `Promise<boolean>`  |  ✅  |   ✅    |   ✅    | ❌  |
 | [isTablet()](#istablet)                                           | `boolean`           |  ✅  |   ✅    |   ✅    | ❌  |
 | [isTabletMode()](#istabletmode)                                   | `Promise<bool>`     |  ❌  |   ❌    |   ✅    | ❌  |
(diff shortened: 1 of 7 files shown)


filiphorvei:
Closing because this functionality is already present using getDeviceType() === 'Desktop'.

PR #1380, opened 2022-02-11
fixed in package.json

NagavikasPulluri:
ts issue fixed

## Description

Fixes #<issue-number>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "react-native-device-info",
   "version": "8.4.7",
   "description": "Get device information using react-native",
-  "react-native": "src/index.ts",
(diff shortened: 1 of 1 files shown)


mikehardy:
You have proposed a change to package.json with no discussion at all of why you would want the change, nor reference to who might be affected and why or who might be fixed and why

That's not how the process works, I have nothing to go on for review.

We can reopen if explain the PR and provide supporting documentation on why needed, how it will affect people

PR #1325, opened 2021-10-09
Example update WIP - needs windows testing, likely broken until fixed

mikehardy:
## Description

Fixes #<issue-number>

Added `yourNewMethodName()` that allows ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

* [ ] I have tested this on a device/simulator for each compatible OS
* [ ] I added the documentation in `README.md`
* [ ] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [ ] I added a sample use of the API (`example/App.js`)

PR #1389, opened 2022-03-04
feat(windows, supportedAbis): Add supportedAbis for windows

kpete2017:
## Description

Added the getSupportedAbis methods for windows which allows a user to grab a windows device's current supported CPU architectures including x86, x64, and arm.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
* [x] I added the documentation in `README.md`
* [x] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [x] I added a sample use of the API (`example/App.js`)

PR #1390, opened 2022-03-04
build(deps): example update, github workflow updates

mikehardy:
Supercedes #1325 

tested on windows ios and android after running refresh-example and checking it out for test, CI working as well


schie:
Whoa! 

This is a big boy — I'll review in a few

mikehardy:
@schie it looks huge but go commit by commit and it's actually really bite-size, each commit is it's own thing and the big ones are auto-generated from refresh-example (thus, skippable). Should be an easy review I hope :crossed_fingers:

refresh-example.sh
```diff
@@ -37,9 +37,11 @@ fi
 \rm -fr example
 
 # Make the new example
-npx react-native init example
+npm_config_yes=true npx react-native init example --skip-install
 pushd example
-npx react-native-windows-init --overwrite
+rm -f _ruby-version Gemfile*
```
mikehardy, line 44: if you don't do this, and you are not running the *exact* version of ruby that react-native template wants, it fails the install, it's really annoying and I'm going to work to change it upstream. I use ruby 3.1.0 locally not 2.7.4 and it should work (I know because I worked to fix all the incompatibilities!).
So just delete the ruby version pinning stuff.

refresh-example.sh
```diff
@@ -37,9 +37,11 @@ fi
 \rm -fr example
 
 # Make the new example
-npx react-native init example
+npm_config_yes=true npx react-native init example --skip-install
```
mikehardy, line 42: npm_config_yes=true makes npx run non-interactively after they added a  new prompt confirming you really want to download and execute internet code

example/patches/react-native-windows+0.67.3.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/node_modules/react-native-windows/Mso/dispatchQueue/dispatchQueue.h b/node_modules/react-native-windows/Mso/dispatchQueue/dispatchQueue.h
+index 0a7369e..e909b93 100644
+--- a/node_modules/react-native-windows/Mso/dispatchQueue/dispatchQueue.h
++++ b/node_modules/react-native-windows/Mso/dispatchQueue/dispatchQueue.h
```
mikehardy, line 4: This is a merged PR upstream but not released yet. patch-package will nicely give us a warning in the future while still working

package.json
```diff
@@ -31,14 +31,17 @@
...
     "lint": "npx eslint ./ --ignore-pattern example --ignore-pattern node_modules --fix --quiet",
     "shipit": "yarn clean && semantic-release",
     "example:install": "yarn && cd example && yarn && cd .. && yarn dev-sync",
+    "example:install-windows": "yarn && cd example && yarn && cd .. && yarn dev-sync-windows",
     "example:install:pods": "cd example/ios && pod install",
     "example:android": "cd example && yarn android",
     "example:android:build": "cd example/android && ./gradlew assembleDebug",
     "example:ios": "cd example && yarn ios",
     "example:ios:build": "cd example/ios && xcodebuild CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ -workspace example.xcworkspace -configuration Debug -scheme example -sdk iphonesimulator -arch x86_64 ONLY_ACTIVE_ARCH=YES | xcpretty",
+    "example:windows": "cd example && yarn windows",
```
mikehardy, line 44: This file has the only useful changes in this commit, but the other files were perturbed by the windows build and it seemed best to include them
The build did work successfully locally on windows! And it works in CI too (which is VS16 toolchain, older than my local one) which proves we work on a pretty wide windows toolchain skew right now

mikehardy:
@schie I left notes in the commit around specific parts that represent knowledge I earned with time, so you can have the knowledge in less time :nerd_face:

mikehardy:
The e2e ios failure is in no way the fault of the changes here even though I *did* change caching. Frustrating flaky, but I'm comfortable with it working since I've seen it working.

https://github.com/react-native-device-info/react-native-device-info/runs/5427679162?check_suite_focus=true#step:35:19
> Fri, 04 Mar 2022 21:04:04 GMT
Error: uploadChunk (start: 100663296, end: 120631203) failed: connect ETIMEDOUT 13.107.42.16:443

working previously:
https://github.com/react-native-device-info/react-native-device-info/runs/5427626919?check_suite_focus=true#step:35:19

and it works in a few other repos :shrug: 

I'm going to consider CI green on this one

mikehardy:
discovered a tiny problem - my `xcopy` switches were wrong in `package.json::scripts::dev-sync-windows` which I noticed while testing #1389 - just re-pushed with the correct ones.

#1389 was really useful as a driver / motivation to work through all this stuff, I must have run refresh-example and the windows dev sync stuff 10 times (which is a *lot* of wall clock time) today as well as the ios / android example scripts a couple times to make sure I didn't break them. It's as good as I can make it now I think

mikehardy:
Pushing my luck here, but the last e2e runs passed, so why not update the core dependencies as well? Last commit is the root package.json dependencies updated where possible. CI should still pass, nothing seemed difficult at all except the NativeEventEmitter mock and making sure the breaking changes in the packages were all just clipping off node < 16

mikehardy:
for what it's worth, I pulled this on a clean machine running windows and tested android + windows on it, everything works cleanly. In fact, it's worth noting that this is the first time really that windows works well with the example app as a 1st class citizen via package.json run scripts, previously you had to do all sorts of junk via manual commands for it to run. It's a nice improvement!

example/android/gradlew.bat
```diff
@@ -1,89 +1,89 @@
...
+@rem      https://www.apache.org/licenses/LICENSE-2.0
+@rem
+@rem Unless required by applicable law or agreed to in writing, software
+@rem distributed under the License is distributed on an "AS IS" BASIS,
+@rem WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+@rem See the License for the specific language governing permissions and
+@rem limitations under the License.
+@rem
+
+@if "%DEBUG%" == "" @echo off
```
schie, line 17: @mikehardy,
I see that the `gradlew.bat` file was updated — was this just a line-ending situation?

example/windows/example.sln
```diff
@@ -54,24 +54,24 @@ Global
...
-		{3D9C9FEA-EB1D-4808-9917-E1E90FDBDFDC}.Release|x86.Build.0 = Release|Win32
-		{3D9C9FEA-EB1D-4808-9917-E1E90FDBDFDC}.Release|x86.Deploy.0 = Release|Win32
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|ARM64.Build.0 = Debug|ARM64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|ARM64.Deploy.0 = Debug|ARM64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|x64.ActiveCfg = Debug|x64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|x64.Build.0 = Debug|x64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|x64.Deploy.0 = Debug|x64
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|x86.ActiveCfg = Debug|Win32
+		{9C5CD441-1749-47EC-90E6-3899B5731898}.Debug|x86.Build.0 = Debug|Win32
```
schie, line 66: all auto generated jazz from Visual Studio?

example/android/gradlew.bat
mikehardy: it appears that way - it was the result of a `./gradlew wrapper` call after bumping to gradle 7.4, and it works on windows (I checked) so... :shrug: - should be okay

example/windows/example.sln
mikehardy: This was actually the result of running the app on windows, it altered things just a touch in response to `yarn example:windows` and then the files were stable on repeated re-runs so it seemed the right thing to do to commit them, so folks trying the example didn't immediately see dirty git tree after running things

schie (approved):

mikehardy:
Nice! Thanks for the review @schie the module is back up to date after a long languish

PR #1394, opened 2022-03-18
fix(ios): Missing device name for iPhone SE 3rd gen

caiohsramos:
Adds missing device name by code for iPhone SE 3rd gen

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |


mikehardy (approved):
Isn't it kind of irritating that these can't be differentiated by name? I guess that's how they do it though, I just checked. Also looks like there were no other new devices so this should do it, thanks!

PR #1398, opened 2022-03-27
feat(ios): add getBrightness support for iOS

luoxuhai:
## Description

Added `getBrightness()` support for iOS to get the current brightness level of the device's main screen.

**Device test screenshot:**

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -819,6 +819,22 @@ RCT_EXPORT_METHOD(getInstallerPackageName:(RCTPromiseResolveBlock)resolve reject
     resolve([EnvironmentValues objectAtIndex:[EnvironmentUtil currentAppEnvironment]]);
 }
 
+- (NSNumber *) getBrightness {
+#if !TARGET_OS_TV
+    return @([UIScreen mainScreen].brightness);
+#else
+    return @(-1);
+#endif
+}
+
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getBrightnessSync) {
+    return self.getBrightness;
+}
+
+RCT_EXPORT_METHOD(getBrightness:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
+    resolve(self.getBrightness);
+}
+
 #pragma mark - dealloc -
 
 - (void)dealloc
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ The example app in this repository shows an example usage of every single API, c
 | [getUsedMemory()](#getusedmemory)                                 | `Promise<number>`   |  ✅  |   ✅    |   ✅    | ✅  |
 | [getUserAgent()](#getuseragent)                                   | `Promise<string>`   |  ✅  |   ✅    |   ❌    | ✅  |
 | [getVersion()](#getversion)                                       | `string`            |  ✅  |   ✅    |   ✅    | ❌  |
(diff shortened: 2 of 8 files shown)


schie (approved):
Very cool!

Is there a way for us to watch for brightness changes?

luoxuhai:
> Very cool!
> 
> Is there a way for us to watch for brightness changes?

`UIScreenBrightnessDidChangeNotification` can watch for brightness changes on ios, maybe the next PR :clap:

mikehardy:
I'm seeing a github error at the moment but will review as soon as github lets me ;-)

>  Sorry, this diff is temporarily unavailable due to heavy server load.

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -819,6 +819,14 @@ - (NSDictionary *) getAvailableLocationProviders {
...
 }
 
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getBrightnessSync) {
+    return @([UIScreen mainScreen].brightness);
+}
+
+RCT_EXPORT_METHOD(getBrightness:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
+    resolve(@([UIScreen mainScreen].brightness));
+}
+
```
mikehardy, line 829: I just checked and this symbol does not seem to be available on tvOS so this will break some people's compilation: https://developer.apple.com/documentation/uikit/uiscreen/1617830-brightness

Also, the method implementation is duplicated when the async one can just refer to the sync one, so we can do the tvOS ifdef once:

```suggestion
- (float) getBrightness {
#if !TARGET_OS_TV
    return @([UIScreen mainScreen].brightness);
#else
    return [@1 floatValue];
#endif
}

RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getBrightnessSync) {
    return @(self.getBrightness);
}

RCT_EXPORT_METHOD(getBrightness:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
    resolve(@(self.getBrightness));
}

```

what do you think of this refactor / tvOS compat suggestion? Does it still work for you like this? (I just coded it in the web ui here so I'm unsure...)

mikehardy (changes requested):
Nice! Needs a tvOS compat enhancement though I think - commented on affected line with suggestion

ios/RNDeviceInfo/RNDeviceInfo.m
luoxuhai: Yes, I agree with this suggestion

luoxuhai:
> Nice! Needs a tvOS compat enhancement though I think - commented on affected line with suggestion

Thank you for your suggestion, I have modified as suggested。

mikehardy (approved):
Before I merge - just confirming - did you actually run the changed code? I can test it tomorrow if not, but if you've already checked it that works. Just want to make sure because I never ever trust my "I coded it in the github web UI" code ;-)

luoxuhai:
> Before I merge - just confirming - did you actually run the changed code? I can test it tomorrow if not, but if you've already checked it that works. Just want to make sure because I never ever trust my "I coded it in the github web UI" code ;-)

Yes, I've tested code :ok_hand:

ios/RNDeviceInfo/RNDeviceInfo.m
mikehardy: :-) - my new years resolution is "support all the platforms", and though I'm not there yet - it's a process - it means I'm keenly aware right now of all the little common failures in the build matrix...

PR #1408, opened 2022-04-28
fix(windows): addListener/removeListeners native stubs to quiet warnings

### LINKED ISSUE #1288: Getting warning with react native 0.65
I'm getting warning with react native new version 0.65
Screenshot_1629435286

namrog84:
Doing windows equivalent for
https://github.com/react-native-device-info/react-native-device-info/pull/1289
https://github.com/react-native-device-info/react-native-device-info/issues/1288
https://github.com/react-native-device-info/react-native-device-info/commit/c518c32202c7af548ffd4bd797662d9493fed3bf

## Description

Fixes #1288

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/windows/code/RNDeviceInfoCPP.h
+++ b/windows/code/RNDeviceInfoCPP.h
@@ -822,6 +822,18 @@ namespace winrt::RNDeviceInfoCPP
         });
     }
 
+    REACT_METHOD(addListener);
+    void addListener(std::string) noexcept
+    {
+        // Keep: Required for RN built in Event Emitter Calls.
+    }
+
+    REACT_METHOD(removeListeners);
+    void removeListeners(int64_t) noexcept
+    {
+        // Keep: Required for RN built in Event Emitter Calls.
+    }
+
   };
 
 }


mikehardy (approved):
Hey @namrog84 :wave: - consider this tenative approval and "sure, let's launch it" - but - there is a nagging concern that this may cause problems. I believe we backed something similar out in other module (netinfo) and my worry is that there is a need to call super or similar if you implement the stub.

Have you investigated that? and/or have you definitely tested listener-related behavior on windows if you implement these?

Would hate to cause a regression by adding these when they do look innocent but we were bitten on netinfo --> https://github.com/react-native-netinfo/react-native-netinfo/pulls?q=is%3Apr+stubs

namrog84:
I've probably applied a dozen of these as patch-packages and into misc projects over the last few months without issue on Windows and Android.   In some cases we handle the code in constructor and destructor and other places just noop it.  Tested it many times in many places without ever a single hiccup. 
Short answer is these basically should just at like noop.  You call the typescript register/unregister but the callback event will never be called, since nothing is implemented. 


For the longer answer and for what you probably already know most of. But for those that might not know, ideally it'd be where we register or unregister the appropriate add and remove listeners for native layer. As of today, for react-native-device-info that it appears to even call into NativeEventEmitter and that's for useOnEvent. The only place I see using useOnEvent is for useIsHeadphonesConnected().

As of today it doesn't have any impact or usage on windows, so nothing will change.
At some point one might want to add windows support and in that case we'd want to add some code to these.
Or if some other event we want to start watching we'd need to add appropriate windows specific code. 




However for specifically the isHeadphoneConnected scenario on windows, it's far more nuanced as it doesn't differentiate the same ways as mobile does about headsets as opposed to just any speaker/attached device, with the exception of a headset connected via a game controller. 
https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input.rawgamecontroller.headsetconnected?view=winrt-22000
But in that case I feel like that'd better be served under a different yet to be implemented API.


Bonus Android isHeadphoneConnected comment
I didn't look too closely, but it appears like in RNDeviceModule.java, the initialize() calls into initializeHeadphoneConnectionReceiver where it registers for a callback that ultimately triggers the RNDeviceInfo_headphoneConnectionDidChange.  
Ideally that should live inside the AddListener and RemoveListener, but right now it happens once at startup automatically and then never unregisters. 
I personally am a fan and would prefer to keep the current design and I think it is better then you don't contend with adding duplicate listeners for the same thing, and you only care about adding/removing at typescript layer, not at the native layer. Perhaps if it was expensive to register for an event handler in native layer it'd make sense too for some too frequently fast firing events. But in every case I've seen it's better to just always initialize the event listener at module initialize and don't specifically unregister it (until app/module shuts down). And thus basically always ignore the Native implementation of AddListener and removeListener stubs.

namrog84:
I haven't investigated trying to call a Super yet but I will investigate that here soon.

mikehardy (approved):
works for me - and I appreciate the details, thank you
w.r.t. the call-super idea, if there is nothing there (maybe not even inheritance, but related machinery for listeners) then obviously it won't matter. I think that is just what bit netinfo on ios, that there was some related machinery and by implementing the methods we were essentially cutting out some default implementation.

That said, I did not have a lot of time for netinfo in the time so we just reverted without deeply inspecting + testing things

For this: let's go!

PR #1424, opened 2022-06-04
fix(example, windows): Define an explicit 10.0 in example project to use latest windows 10 sdk

namrog84:
example project didn't have a defined windows 10 sdk, so for some users it might use a cached older version.
Change it to be explicitly using 10.0 (use latest installed version) to minimize any potential confusion or issue.

https://i.imgur.com/Ldfn81S.png

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Windows |    ✅     |

## Checklist

Tested on a relatively fresh clone and install of things. I only have VS2022 and windows sdk versions 10.0.19041.0 and 10.0.20348.0 installed.

https://i.imgur.com/JEtmkgg.png


mikehardy (approved):
looks reasonable - thanks - I have another little time window to give the windows build another shot locally to see. It's still working in CI at least, that's something :-)

PR #1428, opened 2022-06-23
fix(android, uniqueId)!: remove get uniqueId from startup constants

### LINKED ISSUE #1427: "uniqueId" not compliant with Huawei App market Policies
image
image

fengqingyun2008:
## Description

accessing ANDROID_ID (the underlying system information behind getUniqueId) prior to getting consent from users violates policy in some app stores, this removes it from the startup constants so that users may be informed prior to accessing the information

Fixes #1427 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -189,7 +189,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
     final Map<String, Object> constants = new HashMap<>();
 
-    constants.put("uniqueId", getUniqueIdSync());
+//     constants.put("uniqueId", getUniqueIdSync());
     constants.put("deviceId", Build.BOARD);
     constants.put("bundleId", getReactApplicationContext().getPackageName());
     constants.put("systemName", "Android");


mikehardy (changes requested):
I think we need the other platforms to remove it as well so that the platforms behave the same

https://github.com/react-native-device-info/react-native-device-info/blob/0341ab5b5909a410d40d168832f415b93bd557be/ios/RNDeviceInfo/RNDeviceInfo.m#L59

https://github.com/react-native-device-info/react-native-device-info/blob/0341ab5b5909a410d40d168832f415b93bd557be/windows/code/RNDeviceInfoCPP.h#L30

Also the native constants in typescript layer:
https://github.com/react-native-device-info/react-native-device-info/blob/0341ab5b5909a410d40d168832f415b93bd557be/src/internal/privateTypes.ts#L23

Also the getter must be adjusted to stop looking for it in constants
https://github.com/react-native-device-info/react-native-device-info/blob/0341ab5b5909a410d40d168832f415b93bd557be/src/index.ts#L24

Also the example app should not have it in constants

https://github.com/react-native-device-info/react-native-device-info/blob/0341ab5b5909a410d40d168832f415b93bd557be/example/App.js#L110

I checked the documentation, it's fine.

This is largely incomplete without the changes mentioned in this comment

stawecki:
@fengqingyun2008 do you have time this week to make the suggested changes or do you need some help here?

PR #1432, opened 2022-06-30
fix(android): remove uniqueId from constants

### LINKED ISSUE #1427: "uniqueId" not compliant with Huawei App market Policies

stawecki:
## Description

Obtaining uniqueId on Android has to be explicit in line with newly enforced Huawei and Google Play guidelines.
As an alternative, native method getUniqueIdSync has been implemented on all applicable platforms and is used in the current getUniqueId API method instead of constants.

Fixes #1427

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1001,8 +1001,6 @@ DeviceInfo.getTotalMemory().then((totalMemory) => {
 
 ### getUniqueId()
 
-This is a constant and may be referenced directly
-
 Gets the device unique ID.
 On Android it is currently identical to `getAndroidId()` in this module.
 On iOS it uses the `DeviceUID` uid identifier.
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -189,7 +189,6 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
     final Map<String, Object> constants = new HashMap<>();
 
-    constants.put("uniqueId", getUniqueIdSync());
     constants.put("deviceId", Build.BOARD);
     constants.put("bundleId", getReactApplicationContext().getPackageName());
     constants.put("systemName", "Android");
--- a/example/App.js
+++ b/example/App.js
@@ -22,6 +22,7 @@ import {
   getManufacturer,
   getManufacturerSync,
   syncUniqueId,
+  getUniqueId,
   useBatteryLevel,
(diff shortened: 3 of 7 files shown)


stawecki:
If someone can try this on Windows that would be helpful. I’m only going to be able to do that sometime next week.

stawecki:
I've reverted changes made to `windows/code/RNDeviceInfoCPP.h` after testing on Windows. I think on platforms that cannot call synchronous methods, `uniqueId` has to come from constants. On other platforms, `uniqueId` can be removed and API's `getUniqueId()` will call `getUniqueIdSync()` instead. This will make `uniqueId` an optional value in NativeConstants, instead of being removed completely, but `getUniqueId()` in the API can remain synchronous.

Considering `uniqueId` is no longer in constants on iOS/Android, but was never exposed in the types, this PR has a breaking change, only because the README mentions:
> This is a constant and may be referenced directly

mikehardy:
I thought windows did the sync thing like so though?

>     REACT_SYNC_METHOD(getSupportedAbisSync);

Am I misunderstanding? I am far from an expert on react-native-windows but that macro seems pretty explicit

stawecki:
@mikehardy this is what I've tried [initially](https://github.com/stawecki/react-native-device-info/commit/29cd398ea2b1a00651052348e877614b405bd0f7), but noticed a problem during runtime. Got this for other methods like `isKeyboardConnectedSync()` too. Not sure if this is normal on react-native-windows.

syncmodules_chrome

stawecki:
@mikehardy If you know someone who has experience with Windows, that would be best. Otherwise my suggestion is to settle on the “optional” constant for now and plan to remove from all platforms as an enhancement.

mikehardy:
@namrog84 is my "resident" windows expert - he may know more about the ability of the windows code to really do sync stuff if he has minute :pray: ?

stawecki:
Alright, success! The issue was that I was trying to run it with a fresh react-native-windows template, while [in the example project](https://github.com/stawecki/react-native-device-info/blob/remove-uniqueid-constant-winsync/example/windows/example/App.cpp#L29) `UseWebDebugger` is set to `false`. Potentially something to document a little better for Windows package users.

I created a separate PR #1433 which excludes the unnecessary changes made earlier, so closing this one.

Screenshot 2022-07-06 at 11 55 21

namrog84:
Excellent to see you had success! 👍

PR #1433, opened 2022-07-06
fix!: remove uniqueId from constants, move to sync/async methods

### LINKED ISSUE #1427: "uniqueId" not compliant with Huawei App market Policies

stawecki:
## Description

Obtaining uniqueId on Android has to be explicit in line with newly enforced Huawei and Google Play guidelines.
As an alternative, native method getUniqueIdSync has been implemented on all applicable platforms and is used in the current getUniqueId API method instead of constants.

Fixes #1427

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist


README.md
```diff
@@ -1001,8 +1001,6 @@ DeviceInfo.getTotalMemory().then((totalMemory) => {
 
 ### getUniqueId()
 
-This is a constant and may be referenced directly
-
```
mikehardy, line 1005: Maybe "This is considered sensitive information in some app stores (for example Huawei). For those stores you must obtain user permission prior to accessing it" ?

example/App.js
```diff
@@ -127,6 +127,7 @@ export default class App extends Component {
   getSyncDeviceInfo() {
     let deviceJSON = {};
 
+    deviceJSON.uniqueId = getUniqueId();
```
mikehardy, line 130: This is the synchronous API exercise, so should be the sync version? May need the import above

```suggestion
    deviceJSON.uniqueId = getUniqueIdSync();
```

...then I would expect the getUniqueId() call to be down below, somewhere in the block that starts near line 200 ?

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -364,7 +363,7 @@ - (NSString *) getBuildId {
     return self.getBuildId;
 }
 
-- (NSString *) getUniqueId {
```
mikehardy, line 367: I would expect both flavors here and an internal method ? An internal method that does the work, One that is the getUniqueIdSync and is RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD calling the internal method directly, and one that is a regular RCT_EXPORT_METHOD that takes theh RCTPromiseResolveBlock and resolves by calling the internal method. isTablet or isEmulator could serve as a good reference

src/index.ts
```diff
@@ -918,6 +918,7 @@ const deviceInfoModule: DeviceInfoModule = {
   getType,
   getTypeSync,
   getUniqueId,
+  getUniqueIdSync: getUniqueId,
```
mikehardy, line 921: this is unexpected, I would expect both method signatures to exist, and not be mapped to each other?

src/internal/privateTypes.ts
```diff
@@ -114,6 +113,7 @@ interface ExposedNativeMethods {
   getTotalMemorySync: () => number;
   getType: () => Promise<string>;
   getTypeSync: () => string;
+  getUniqueIdSync: () => string;
```
mikehardy, line 116: same here - i expect both methods?

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -189,7 +189,6 @@ public Map<String, Object> getConstants() {
 
     final Map<String, Object> constants = new HashMap<>();
 
-    constants.put("uniqueId", getUniqueIdSync());
```
mikehardy, line 192: with the constant gone, I expect to see a getUniqueIdSync method that is blocking/sync in the java implementation?

mikehardy (changes requested):
I think this is close - and I really appreciate the effort working through this to get it done
I think all of my comments are related to a conceptual gap where my expectation is that now that uniqueId is no longer in constants, I expect to see the "typical" (non-constant) pair of getUniqueIdSync/getUniqueId in all of the areas and then used alternately in the sync / non-sync version of the example app, but it looks like you've only got one flavor of the native APIs and have "collapsed" the two together ? There are examples of the sync/non-sync API pairs all over the native code (and used in the example app) - if it looked like those I think this would be ready to go

the interaction with the debugger is exactly why there is a non-sync flavor by the way :-), most people want their stuff sync (or they sure said they did in v3!) but then they don't want a slow startup time (hello, constants - you are slow on startup) but then we all learned sync methods don't work in the debugger (so hello async methods). Basically you get to pick your tradeoffs depending on how important the info is to you at startup

stawecki:
> I think this is close - and I really appreciate the effort working through this to get it done

You're welcome!

> I expect to see the "typical" (non-constant) pair of getUniqueIdSync/getUniqueId in all of the areas and then used alternately in the sync / non-sync version of the example app, but it looks like you've only got one flavor of the native APIs and have "collapsed" the two together ? 

Pretty much. I probably should've included some explanations, but basically the problem with changing the API to fit this pattern is that the current getUniqueId method which is also exported in index.ts is actually sync:

```
export const getUniqueId = () =>
  getSupportedPlatformInfoSync({
    defaultValue: 'unknown',
    memoKey: 'uniqueId',
    supportedPlatforms: ['android', 'ios', 'windows'],
    getter: () => RNDeviceInfo.uniqueId,
  });
```
https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L19-L25

In the projects that I'm working on, we actually use `getUniqueId()` in synchronous manner and I wasn't aware it relied on constants as I haven't seen it in the type definitions. So my initial thought was that changing this behaviour now, would be a serious breaking change. For people who already used the synchronous `getUniqueId()` and want to start complying with Play/Huawei policies, the simplest solution seems to fix the constants and retain the current API contract.

I see your point, but I wonder, if this is something that should happen separately in a major release? In some cases people may not be type-checking whether `getUniqueId()` returns a `string` or `Promise<string>` so this change would have to be well-communicated.

I think this applies to any constant getter too. Right now I don't see a reason to refactor `getModel` or `getBrand`, but currently they are synchronous as well, so separating these calls into sync/async methods would mean a significant API contract change.

Thanks for checking this!

mikehardy:
It's definitely going to be a breaking change with appropriate release note in big capital letters
I'm only considering getUniqueId now, for this one - so one small-ish code change, one big release note + major version with instructions on what exactly maps to what (I'll handle that part during release), and we're set?

mikehardy:
Oh - I think I wasn't clear, my preference is to do it all now, not two version bumps one that is a halfway. Pull uniqueId from constants and make it a "full-fledged" Sync/Async API just like all other non-constants, issue the breaking change and be done. For me anyway (traveling now, wanting to optimize time driving the release machinery here / triple-checking everything) it's best to batch it and be done

stawecki:
Oh I see. That makes sense. I'll adjust to match Sync/Async API pattern.

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
stawecki: `getUniqueIdSync()` was already implemented on line 804.  `getUniqueId(Promise p)` added.

ios/RNDeviceInfo/RNDeviceInfo.m
stawecki: See new methods. Internal method getUniqueId renamed to uniqueId. React methods getUniqueIdSync/getUniqueId added.

src/internal/privateTypes.ts
stawecki: Added

example/App.js
stawecki: I've updated examples for sync and async device info.

stawecki:
I think that's a wrap.

Note `RNDeviceInfoCPP.h` already has getUniqueIdSync() and getUniqueId() [implemented](https://github.com/react-native-device-info/react-native-device-info/blob/master/windows/code/RNDeviceInfoCPP.h#L654-L670), so now it's just finally mapped in index.ts

I've completed a run test on all platforms. Since it's sync+async, example is in both sections:

uniqueid_runtest

mikehardy (approved):
this looks fantastic, thank you!

mikehardy:
Oh, no good deed unpunished. Looks like the jest tests need a tweak 

```
 ● string getters › getUniqueId* › supports windows
    expect(received).toEqual(expected) // deep equality
    Expected: "unknown-test"
    Received: {"_U": 0, "_V": 2, "_W": [TypeError: nativeInterface_1.default.getUniqueId is not a function], "_X": null}
      174 |     it.each(['ios', 'android', 'windows'])('supports %s', (os) => {
      175 |       Platform.OS = os as any;
    > 176 |       expect(fn()).toEqual('unknown-test');
          |                    ^
      177 |     });
      178 |   });
      179 |
      at __tests__/getters.test.ts:176:20
```

stawecki:
@mikehardy Can you try now? It's passing for me.

PR #1437, opened 2022-07-11
fix: add getDeviceType and getDeviceTypeSync mocks

maximilize:
## Description

The mocks `getDeviceType` and `getDeviceTypeSync` were missing after updating to v10.


schie (approved):
LGTM — Thanks!

jest/react-native-device-info-mock.js
```diff
@@ -37,6 +37,8 @@ const diMock = {
   getDeviceName: stringFnAsync(),
   getDeviceNameSync: stringFnSync(),
   getDeviceToken: stringFnAsync(),
+  getDeviceType: stringFnAsync(),
```
mikehardy, line 40: This duplicates the existing one, but the existing `getDeviceType` needs to change to `stringFnAsync()`

mikehardy (changes requested):
Thanks for posting a PR to sync up the mocks, much appreciated - looks like it needs a small tweak?

jest/react-native-device-info-mock.js
schie: gah! this is what I get for reviewing before the tests complete!

jest/react-native-device-info-mock.js
```diff
@@ -122,6 +122,7 @@ const diMock = {
   getBundleId: stringFnSync(),
   getDeviceId: stringFnSync(),
   getDeviceType: stringFnSync(),
```
mikehardy, line 124: This one needs to be stringFnAsync now I think?
```suggestion
  getDeviceType: stringFnASync(),
```

mikehardy (changes requested):
new signature is in and looks good, but old needs to be async now I think?

no good deed unpunished! :sweat_smile:

jest/react-native-device-info-mock.js
maximilize: The original function is also not async, so I guess it's fine like this.

See https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L565

mikehardy: Sorry for the delay! You are correct

mikehardy (approved):

PR #1441, opened 2022-07-18
fix: add missing getUniqueIdSync in index.js.flow

grit96:
`getUniqueIdSync` was added in #1433 but `index.js.flow` wasn't updated


mikehardy:
Sorry! I just don't use flow and not many do so these slip through the cracks. I wonder if flow-typed might be a better home for the typings now that we're typescript internally. That said I'll do my best to merge any PRs related of course

mikehardy (approved):
spot on

PR #1460, opened 2022-09-13
fix(hasNotch): add iPhone 14 Pro / Pro Max to devices with notch

manuhook:
## Description

Add iPhone 14 Pro / iPhone 14 Pro Max to devices with notch

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |

## Checklist

* [ X] I have tested this on a device/simulator for each compatible OS

### DIFF
--- a/src/internal/devicesWithNotch.ts
+++ b/src/internal/devicesWithNotch.ts
@@ -11,6 +11,14 @@ const devicesWithNotch: NotchDevice[] = [
   },
   {
     brand: 'Apple',
+    model: 'iPhone 14 Pro',
+  },
+  {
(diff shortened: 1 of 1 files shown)


manuhook:
Yes sorry I should have explained more, I thought first that the omission was voluntary, but as for Android phones we consider that the "holes / punctures" for the camera in the screen are "notch" devices, it should be the same for iPhone 14 Pro / Pro max.

mikehardy:
I still think https://github.com/th3rdwave/react-native-safe-area-context but if people are still using this older style, happy to have it maintained, yes :-)

PR #1464, opened 2022-09-17
feat(ios, firstInstallTime): implement first install time for iOS

### LINKED ISSUE #1440: add getFirstInstallTime() for iOS
## Description

I want to know when the app was installed. I think `getFirstInstallTime()` will do that, but it is only available for Android.

Maybe this [article](https://medium.com/quick-code/how-to-get-app-install-date-in-react-native-ba4519150478) is helpful.

## Use Case

Add/complement iOS part for `getFirstInstallTime()`

mikehardy:
## Description

Fixes #1440 by implementing getFirstInstallTime for iOS

Also includes a "please be quiet" fix for Xcode builds of the example, and updates the example just a little bit - those were just ergonomic things I did while I was in there since I was testing/re-testing/re-testing etc while adding this method

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

(android, windows already implemented)

## Checklist

types and sample did not need an update as it was just extending an API to a new platform vs having it return default (-1) there


schie (approved):
LGTM

mikehardy:
Always nice to virtually "see" you @schie - thanks for the extra set of :eyes:

mikehardy:
proof is better then simply asserting you tested it, and I did struggle with the number formatting since NSDate in obj-c wants to kick out a double and that left 4 decimal places in the first attempt, so here's what it looks like in the end, nice and pretty

image

PR #1471, opened 2022-10-05
Fixes getDeviceName on Android when targeting API level 32 or newer

### LINKED ISSUE #1463: DeviceInfo.getDeviceName() returns 'unknown' on Android
## Summary

We just upgraded from DeviceInfo v8.3.2 to v10.1.2 and noticed that `DeviceInfo.getDeviceName()` is always returning 'unknown' on Android.

|             |     |
| ----------- | --- |
| Version     | 10.1.2   |
| Affected OS | Android   |
| OS Version  | 12, 13   |

## Current behavior

`DeviceInfo.getDeviceName()` always returns 'unknown' as the device name.

## Expected behavior

`DeviceInfo.getDeviceName()` should return the actual device name.

jcarioti:
## Description

Fixes #1463

This fixes devices targeting API 32 or 33 from always returning 'unknown' due to an exception when fetching the bluetooth name.

## Checklist

* [N/A] I added the documentation in `README.md`
* [N/A] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [N/A] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -687,9 +687,11 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   @ReactMethod(isBlockingSynchronousMethod = true)
   public String getDeviceNameSync() {
     try {
-      String bluetoothName = Settings.Secure.getString(getReactApplicationContext().getContentResolver(), "bluetooth_name");
-      if (bluetoothName != null) {
-        return bluetoothName;
+      if (Build.VERSION.SDK_INT <= 31) {
+        String bluetoothName = Settings.Secure.getString(getReactApplicationContext().getContentResolver(), "bluetooth_name");
+        if (bluetoothName != null) {
+          return bluetoothName;
+        }
       }
 
       if (Build.VERSION.SDK_INT >= 25) {


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -687,9 +687,11 @@ public double getLastUpdateTimeSync() {
...
-      if (bluetoothName != null) {
-        return bluetoothName;
+      if (Build.VERSION.SDK_INT <= 31) {
+        String bluetoothName = Settings.Secure.getString(getReactApplicationContext().getContentResolver(), "bluetooth_name");
+        if (bluetoothName != null) {
+          return bluetoothName;
+        }
       }
 
       if (Build.VERSION.SDK_INT >= 25) {
```
mikehardy, line 697: Hey - first of all, thanks for posting this - it does look functional
However, after looking back at git blame, I think it's more confusing to read than it needs to be, and that is mostly because of the way I re-structured it going on 3 years ago.

I think a clearer way to express the idea is to realize that there are really only 2 cases "You are API >= 25" and "you are API <25". The new thing where API>31 throws seems like it adds to the conditional but I don't think it really does?

So if you instead moved the `if (Build.VERSION.SDK_INT >= 25) {` conditional chunk entirely above the other one, and kept the other one as the "default case / no conditional fallback" I think the code will work the same as what you proposed, but without the new SDK_INT <= 31 check?

That should make it easier to read - especially if there was a comment saying "this was the old way to get the device name before API 26" on the old implementation (that now throws)

What do you think?

mikehardy (changes requested):

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
jcarioti: By reversing that order, it would be changing the expected behavior for anyone still on APIs 25 - 31. Instead of getting the bluetooth name as it was before (even if it's permitted), they'd be getting the custom device name now instead.

That was how I originally implemented my patch, but for a PR I figured you'd want consistency with previous versions of the library.

Besides the consistency thing, I don't have a major opinion one way or the other. It seems like the device name is more appropriate than the bluetooth name, if we want to be technical about it.

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -687,9 +687,11 @@ public double getLastUpdateTimeSync() {
...
   public String getDeviceNameSync() {
     try {
-      String bluetoothName = Settings.Secure.getString(getReactApplicationContext().getContentResolver(), "bluetooth_name");
-      if (bluetoothName != null) {
-        return bluetoothName;
+      if (Build.VERSION.SDK_INT <= 31) {
+        String bluetoothName = Settings.Secure.getString(getReactApplicationContext().getContentResolver(), "bluetooth_name");
+        if (bluetoothName != null) {
+          return bluetoothName;
+        }
```
schie, line 694: should return value be set here if the condition `bluetoothName != null`  is not met?

jcarioti: It's handled a few lines lower outside of the try/catch block. If neither of these blocks short circuit and return early, the last thing that happens is it returns 'unknown'.

schie: ah, I see

android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
mikehardy: Ah, I see what you mean, and it appears that some devices do return something in that 25-31 API range, for the "old" implementation (https://stackoverflow.com/a/56024910/9910298). I was thinking (incorrectly) that you would always want the "new" implementation for API >= 25, and while that may be true it is not the way it has been done, so...

So yes, this looks like the most backwards-compatible implementation, works for me!

mikehardy (approved):

PR #1473, opened 2022-10-12
fix(devicetype): resolve getMetrics is deprecated when targetApi 31

thuongtv-vn:
## Description

Get device screen width and height via recommended API WindowMetrics.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [ x] I have tested this on a device/simulator for each compatible Android

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceTypeResolver.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceTypeResolver.java
@@ -6,6 +6,7 @@ import android.content.res.Configuration;
 import android.os.Build;
 import android.util.DisplayMetrics;
 import android.view.WindowManager;
+import android.view.WindowMetrics;
 
 import com.learnium.RNDeviceInfo.DeviceType;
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceTypeResolver.java
```diff
@@ -66,19 +67,29 @@ private DeviceType getDeviceTypeFromPhysicalSize() {
...
+
+      widthInches = windowMetrics.getBounds().width() / (double) metrics.xdpi;
+      heightInches = windowMetrics.getBounds().height() / (double) metrics.ydpi;
     }
 
-    // Calculate physical size.
-    double widthInches = metrics.widthPixels / (double) metrics.xdpi;
-    double heightInches = metrics.heightPixels / (double) metrics.ydpi;
-    double diagonalSizeInches = Math.sqrt(Math.pow(widthInches, 2) + Math.pow(heightInches, 2));
+    double diagonalSizeInches = Math.floor(Math.sqrt(Math.pow(widthInches, 2) + Math.pow(heightInches, 2)));
```
mikehardy, line 92: Hi there!
I like the change to remove the deprecation by using new APIs, however, you've altered the implementation here with the addition of Math.floor, and indeed some of the breakpoint tests below are in non-integer ranges, so it's possible this will return a different result now if the handset has an actual diagonal width of 6.91 inches - now that will be a handset instead of a tablet as you'll round 6.91 down to 6.0 and it will fall into a different bracket of conditionals

```suggestion
    double diagonalSizeInches = Math.sqrt(Math.pow(widthInches, 2) + Math.pow(heightInches, 2));
```

mikehardy (changes requested):

PR #1475, opened 2022-10-21
feat(android, instanceId): implement as UUIDv4 in shared prefs, w/fallback

mikehardy:
## Description

firebase IID and GMS IID are both long long deprecated and problematic now

We need to drop them but if anyone happens to be relying on them they will need to maintain
the current value

So start in this version by saving current values to shared preferences for stability

If Firebase IID and GMS IID are not available, start generating a UUIDv4 instance ID and saving it

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
@@ -1,9 +1,14 @@
 package com.learnium.RNDeviceInfo.resolver;
 
 import android.content.Context;
+import android.content.SharedPreferences;
+import android.os.Build;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.UUID;
+
+import com.learnium.RNDeviceInfo.RNDeviceModule;
 
 /**
  * Instance Id resolver's single purpose is to get the device's Instance Id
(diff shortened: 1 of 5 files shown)


mikehardy:
@schie would love a review here if you have a minute. The 3rd commit is the only one that's interesting.

- We really need to drop the IID packages (firebase and GMS)
- We need to not depend on Firebase Installations (it would pull in the Task library and Firebase more deeply - which makes this non-free-software-compliant...)

So I'm trying to make a release + major version where we persist whatever current values people have, if they have them, then next major version I'll purge out the IID stuff with a warning for people that were using the IID dependencies that they need to make sure they've had a release with wide adoption on the old major version first to keep instanceId stability

android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
```diff
@@ -1,46 +1,80 @@
...
+import com.learnium.RNDeviceInfo.RNDeviceModule;
 
 /**
  * Instance Id resolver's single purpose is to get the device's Instance Id
  * author: Andres Aguilar
  */
 public class DeviceIdResolver {
 
-  private final Context context;
+  private final ReactApplicationContext context;
```
schie, line 21: Why was the type changed from `Context` to `ReactApplicationContext` ? I may just be out of the loop

schie (commented):
Overall, it looks good to me, but I am curious about the context type change.

I'll approve once I understand that bit.

android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
mikehardy: What I really needed is the SharedPreferences object, which, if *I* understand correctly is normally fetched via

```java
SharedPreferences sharedPref = getActivity().getPreferences(Context.MODE_PRIVATE);
```

What I don't have normally is the `Activity`

- you can access it from the module via an instance method but we are in a separate object here
- you have to be careful because frequently the Activity is null when you least expect it! Classic react-native trap - the bridge may dispose of the Activity as app goes to background, but code is executing async and you get a null

What I want is the shared prefs though, and it turns out the `ReactApplicationContext` has that as a direct method, and that is what we get during module construction.

So I plumb that through to this class during instantiation, and now I can get the shared prefs.

Okay so then I did abstract the actual shared prefs get to a static on the main module, but that's so they were all in one spot without duplicating the constant and access style everywhere. But it still requires that ReactApplicationContext to get them

It's a little janky, but it works and does not depend on the Activity (which I don't always have)

mikehardy: It looks like this is actually available from *any* Context, so you're right, I don't really need to change the type here, I can keep it more generic

https://developer.android.com/reference/android/content/Context#getSharedPreferences(java.lang.String,%20int)

schie (approved):
:shipit:

android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
mikehardy: @schie pushed 3rd commit with use of `Context` vs `ReactApplicationContext`

schie:
@mikehardy, what change happened in the force push?

mikehardy:
> @mikehardy, what change happened in the force push?

Just the change from ReactApplicationContext to Context - tested locally, works fine with the generic type, you were right to pick on that

mikehardy:
@schie here was the diff I stashed before rebasing to edit it and pushing it out:

```diff

mike@osxvm:~/work/react-random/react-native-device-info (@mikehardy/android-instance-id-refactor *) % git diff
diff --git a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
index ccdbb18..b01e9d2 100644
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -164,7 +164,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     return NAME;
   }
 
-  public static SharedPreferences getRNDISharedPreferences(ReactApplicationContext context) {
+  public static SharedPreferences getRNDISharedPreferences(Context context) {
     return context.getSharedPreferences("react-native-device-info", Context.MODE_PRIVATE);
   }
 
diff --git a/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java b/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
index 7f2beb5..3e90e13 100644
--- a/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/resolver/DeviceIdResolver.java
@@ -8,8 +8,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.UUID;
 
-import com.facebook.react.bridge.ReactApplicationContext;
-
 import com.learnium.RNDeviceInfo.RNDeviceModule;
 
 /**
@@ -18,9 +16,9 @@ import com.learnium.RNDeviceInfo.RNDeviceModule;
  */
 public class DeviceIdResolver {
 
-  private final ReactApplicationContext context;
+  private final Context context;
 
-  public DeviceIdResolver(ReactApplicationContext context) {
+  public DeviceIdResolver(Context context) {
     this.context = context;
   }
 
```

schie (approved):

mikehardy:
Thanks for the review @schie - appreciated as ever. And you motivated me to tag up issues here, and in a few other repos with hacktoberfest-accepted :smile: let's plant some trees

PR #1476, opened 2022-10-21
refactor(index): renaming variable name of default export

### LINKED ISSUE #1461: Main module is declared as  `deviceInfoModule` instead of `DeviceInfo`
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

In the documentation, the main module is imported as `DeviceInfo`.

However, in the code the exported variable is named `deviceInfoModule`. This means the generated TypeScript declaration file (`/lib/typescript/index.d.ts`) defines it as:

```ts
declare const deviceInfoModule: DeviceInfoModule;
export default deviceInfoModule;
```

The auto import feature VS Code will thus treat (and import) the module as if it is named `deviceInfoModule`:

image
image

IMO, it would be better to match the documentation and declare it as:

```ts
declare const DeviceInfo: DeviceInfoModule;
export default DeviceInfo;
```

Very minor, but it makes the experience a bit nicer. I'd be happy to provide a PR - for now I'm just fixing it locally with `patch-package`.

schie:
## Description

Fixes #1461

renames constant name to `DeviceInfo` to improvde dev-experience of those that use IDEs that
auto-complete imports

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |
| Windows |    ✅❌     |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -841,7 +841,7 @@ export function useBrightness(): number | null {
 
 export type { AsyncHookResult, DeviceType, LocationProviderInfo, PowerState };
 
-const deviceInfoModule: DeviceInfoModule = {
+const DeviceInfo: DeviceInfoModule = {
   getAndroidId,
   getAndroidIdSync,
   getApiLevel,
@@ -991,4 +991,4 @@ const deviceInfoModule: DeviceInfoModule = {
   useBrightness,
 };
 
-export default deviceInfoModule;
+export default DeviceInfo;


mikehardy (approved):
Does seem reasonable, per discussion before.
Hopefully this will avoid the curse of "no good deed unpunished" !

mikehardy:
Android CI failure was unrelated (just flaky emulator), I just kicked it off again and marked pending merge
If you see it went green before I did please feel free to whack that merge button

schie:
> Does seem reasonable, per discussion before.
> Hopefully this will avoid the curse of "no good deed unpunished" !

I ran it locally, so it should be fine, but 🤞

schie:
interesting... a 'refactor' does not trigger a release

PR #1495, opened 2023-01-03
feat: add DisplayZoomed ios

### LINKED ISSUE #1485: Support for Display Zoom
## Description

Would like to see a way to determine if the device's display is set to Zoom vs Standard

## Use Case

When users zoom in on our app, the styling gets messed up. Having a way to determine if the device is zoomed or not would be incredibly helpful

eddyriv3ra:
add feat to check if the app has the zoomed feature enabled only for ios

## Description

Fixes #1485

Added `isDisplayZoomed()` that allows to check if the user has enabled the `zoomed` feature from IOS

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist


mikehardy (approved):
Seems reasonable - sorry for the delay on review!

mikehardy:
Just closing / reopening to see if that kicks off the CI checks, hopefully it does...

PR #1502, opened 2023-02-02
fix(ios, getCarrier): deal with carrierName deprecation

taka1068:
## Description

This PR deals with `carrierName` [deprecation since iOS 16](https://developer.apple.com/documentation/coretelephony/ctcarrier/1620313-carriername).

According to the header file, `carrierName` returns "--" in the future.
If so, carrierName returns 'unknown' with this PR.

Screenshot 2023-02-02 at 22 56 07

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,10 @@ DeviceInfo.getCarrier().then((carrier) => {
 });
 ```
 
+#### Notes
+
+> - According to [apple.developer.com](https://developer.apple.com/documentation/coretelephony/ctcarrier/1620313-carriername), getting the carrier name in iOS 16 is deprecated and returns "unknown" at some point in the future in iOS.
+
 ---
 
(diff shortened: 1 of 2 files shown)


mikehardy (changes requested):
Sorry for the delay in review - I like this change in the code, thank you! - seems like the future behavior should get a mention in the docs for the related API though? Otherwise it may be a little unexpected for future developers and result in issues

taka1068:
@mikehardy 
Sorry for the late reply.
> seems like the future behavior should get a mention in the docs for the related API though? Otherwise it may be a little unexpected for future developers and result in issues

Thanks a lot for your review! I totally agree and have added some comments to avoid some confusion.

PR #1506, opened 2023-02-14
getUserAgentSync: fix memoKey

kesha-antonov:
## Description

Fixes `getUserAgent` returning `unknown` when used with the `getUserAgentSync`

For example,

```javascript
  let userAgent = '' + Platform.select({
    android: DeviceInfo.getUserAgentSync(), // sync not supported by iOS - supportedPlatforms: ['android', 'web']
    default: await DeviceInfo.getUserAgent(),
  })
```

On `ios` it will return `unknown`

**The reason:**

`getUserAgent` and `getUserAgent` have same `memoKey`

https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L228
https://github.com/react-native-device-info/react-native-device-info/blob/master/src/index.ts#L236

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -233,7 +233,7 @@ export const getUserAgent = () =>
 
 export const getUserAgentSync = () =>
   getSupportedPlatformInfoSync({
-    memoKey: 'userAgent',
+    memoKey: 'userAgentSync',
     defaultValue: 'unknown',
     supportedPlatforms: ['android', 'web'],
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
interesting! This API (getUserAgent) is a corner case in so many ways, the change here seems like it is in tune with how it works though, thank you

PR #1509, opened 2023-03-03
feat: add new arch

WoLewicki:
PR adding new arch support to the library.


ajanuar:
Hello, any plan to merge this soon?

zabojad:
It doesn't compile in XCode.

zabojad:
> It doesn't compile in XCode. I get two compile errors: `Undefined symbol: _OBJC_CLASS_$_DCDevice` and `Undefined symbol: _OBJC_CLASS_$_LAContext`...
> 
> Screenshot 2023-06-14 at 15 51 23
> My RN version is 0.71.8 and I'm compiling with XCode 14.1

Upgrading XCode to 14.3.1 solved the problem...

RNDeviceInfo.podspec
```diff
@@ -15,7 +17,28 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '10.0'
 
   s.source       = { :git => "https://github.com/react-native-device-info/react-native-device-info.git", :tag => "v#{s.version}" }
-  s.source_files  = "ios/**/*.{h,m}"
+  s.source_files  = "ios/**/*.{h,m,mm}"
+
+  if fabric_enabled
+    folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
```
vonovak, line 23: there's an easier way btw: https://github.com/react-native-datetimepicker/datetimepicker/blob/4bd45866508738dd7511264f8ea12cbe64e07dda/RNDateTimePicker.podspec#L19

WoLewicki: We'd have to do it in all the libs we migrated then.

vonovak: @WoLewicki, @j-piasecki  actually, you don't need to wrap `install_modules_dependencies` into `if fabric_enabled` - it can be used without the condition: https://reactnative.dev/docs/next/the-new-architecture/backward-compatibility-turbomodules#ios

I missed that in the example I gave

j-piasecki: @vonovak That's true, but IIRC it's been added in RN 0.71 so that would mean dropping support for versions below that.

zabojad:
Gosh, that error popped up again on my 0.71.12 project (XCode 14.3.1):

Screenshot 2023-07-13 at 17 27 10

@WoLewicki on which version of RN have you been testing your PR?

Do you confirm that installing you PR with either:

`yarn add https://github.com/WoLewicki/react-native-device-info.git#@wolewicki/add-fabric`
or 
`yarn add react-native-device-info/react-native-device-info#1509/head`

should be fine, right? Or are they some extra steps?

zabojad:
mmm, I also tried cloning the PR, running `yarn build` and installing the module from that and I'm still having this... Cannot build the app for my device...

zabojad:
@WoLewicki could you add that to your PR?

WoLewicki:
@zabojad why should this be part of this PR? Wasn't it needed before also?

zabojad:
@WoLewicki well at least your PR doesn't work with latest XCode versions on a new RN project... I would include it in your PR as it anyway makes sense that the podspec references the frameworks the module uses... Nope?

WoLewicki:
Yeah, for sure, but first I'd like to know if it is the problem on old arch too. cc @vonovak do you know maybe why it can be the case?

vonovak:
> Yeah, for sure, but first I'd like to know if it is the problem on old arch too. cc @vonovak do you know maybe why it can be the case?

unfortunately, no, didn't deal with such case before..

WoLewicki:
I'd like to have it merged and released for sure @vonovak  😅

vonovak:
Hey, me too for sure, but I'm not a maintainer here 😄

WoLewicki:
Yeah, but the errors mentioned above don't seem to be caused by the migration itself. Please ping us when you will have some time to check it and then I can update it to the latest main 🚀

mikehardy:
For what it's worth, New Architecture is now compatible with `use_frameworks! :linkage => :static` which is my only react-native use case, so at least as your friendly but doesn't-have-much-time maintainer here this is now a lot easier for me to test and work with...

WoLewicki:
Nice, so ping me when you will have time to check it and I'll resolve the conflicts then 🚀

mikehardy:
Hey @WoLewicki - I'm still buried right now with a mix of personal stuff and paying work and this module deserves better but doesn't have any funding attached 🙈 . I'm not saying I have zero time, but not the kind of time to work through a new architecture conversion when I haven't personally done one before.

I'm happy to share out admin rights on the repo to you if you are interested? No firm commitment required other than "do your best, as you have time" but it would unblock you here as you could just get it done. I had nearly zero experience when I first started, so you'll be doing way better than that just from the start! 😆 

The repo has all the auto-publishing / semantic release stuff set up so it's really easy to work with, the hard part is just getting PRs right (as it should be - the real work...) and all the other stuff about releasing is handled via github actions

Let me know what you think

WoLewicki:
Hey @mikehardy, I might have some time to look at the issues and PRs since I already spend much time in the OSS libraries for React Native. So yeah, I'm interested in helping with maintaining it 🚀

mikehardy:
Wow - this last week has not been kind to me 😆 - took me this long to simply add you to the project and that's easy!

At any rate, I obviously need the help and will really appreciate it @WoLewicki so I have gone bling 🪄  and now you have access. Let me know if you are curious about anything or need anything and I'll do my best to help

j-piasecki:
@mikehardy @WoLewicki has there been any progress here?

Issue #1544, opened 2023-06-17
Follow-up: Illegal invocation/promise rejection from `navigator.getBattery()` call


bradjones1:
## Summary

|             |     |
| ----------- | --- |
| Version     | `9.0.2` |
| Affected OS | Chrome for Windows |
| OS Version  | `86.0.4240` |

## Current behavior

`navigator.getBattery()` rejected with `TypeError: Failed to execute 'getBattery' on 'Navigator': Illegal invocation`

Backtrace: https://not-vanilla-llc.sentry.io/share/issue/058cd8858c464b5f974f36a113994d16/

Basically the same issue as https://github.com/react-native-device-info/react-native-device-info/issues/1223 which was addressed at least in part by [this change](https://github.com/react-native-device-info/react-native-device-info/commit/77dcf2cab4db3523f4cf4da56a80e962a05cc593#diff-a2eea7d51d5408108433b3aa81ed471741b8f15ae58ca9dd358da064f819c3b5) in https://github.com/react-native-device-info/react-native-device-info/pull/1333 but there are cases in which the promise might still reject even if the method is truthy.

See also https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery#exceptions for (other?) reasons for exceptions, e.g. if the permissions are insufficient to introspect the battery.

## Expected behavior

I am using RNDI for font-based values, however the battery status event subscriptions/values are set up during the `init()` on the package's load, so any issue with the `getBattery()` call throws an exception that is not easily caught by the calling code... because this is not part of the method that the caller actually intends to call, it's run during `index.js` being loaded.

PR #1545, opened 2023-06-17
fix(web): Catch battery status promise rejections

bradjones1:
## Description

Fixes https://github.com/react-native-device-info/react-native-device-info/issues/1544

## Compatibility

Web-only fix.

## Checklist

* `[n/a]` I added the documentation in `README.md`
* `[n/a]` I updated the typings files (`privateTypes.ts`, `types.ts`)
* `[n/a]` I added a sample use of the API (`example/App.js`)

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -150,7 +150,7 @@ export const isCameraPresentSync = () => {
 
 export const getBatteryLevel = async () => {
   if (navigator.getBattery) {
-    return navigator.getBattery().then(battery => battery.level).catch(() => {});
+    return navigator.getBattery().then(battery => battery.level).catch(() => -1);
   }
   return -1;
 };
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -81,7 +81,7 @@ const init = () => {
         deviceInfoEmitter.emit('RNDeviceInfo_batteryLevelIsLow', level);
       }
     });
-  });
+  }).catch(() => {});
 };
 
 const getBaseOsSync = () => {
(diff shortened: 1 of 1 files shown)


machour:
LGTM, but should this be considered as a BC breaking change? 
Some users may be try/catching and handling errors differently than what you're doing here..

bradjones1:
>  Some users may be try/catching and handling errors differently than what you're doing here..

This is a fair point - how about a more narrow change that only affects the `init()` call, since that basically can't be caught right now (which is why I worked on this to begin with)?

bradjones1:
Frustratingly, it doesn't appear that the `catch()` at https://github.com/react-native-device-info/react-native-device-info/pull/1545/files#diff-a2eea7d51d5408108433b3aa81ed471741b8f15ae58ca9dd358da064f819c3b5R84 is consistently working on Windows. I have seen this resurface recently, although with much less frequency. Chrome 86 and 100. I can't recreate it with a similar browser setup on LambdaTest... invoking `navigator.getBattery()` in the JS console returns the `BatteryManager`.

mikehardy:
Thanks for your patience - sorry my time here is so intermittent, but I can't promise more in general 😬 - at any rate, I like the idea of the catches, but if it is not consistent it doesn't help much

(though as an aside: what !? how can a catch not consistently catch...how is that even a thing? So difficult trying to produce reliable code for all platforms...)

As a policy note on breaking changes (assuming consistent behavior can be created...) I am okay with breaking changes in general, my stance is "version numbers are free". My requirement is more about developer empathy in that breaking changes *do* get a semver major, *and* they are accompanied with a really good developer note explaining exactly who is affected, and exactly what changed, and a reasonable attempt to specify the likely change for adoption. That's all - and it's not difficult + and I usually handle that myself during the commit so that the auto-release infrastructure picks it up.

TL;DR: no big deal to do a breaking change normally

mikehardy:
Marking as "on hold" until behavior consistency is achieved so whatever path forward is the final path forward is reliable vs flaky

PR #1546, opened 2023-06-19
feat(android): add support for React Native 0.73

MasonLe2497:
# Summary
Starting from React Native v0.73 , all libraries will need to be updated with these two one-liners due to Android Gradle Plugin upgrade
[discussions-671](https://github.com/react-native-community/discussions-and-proposals/issues/671)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,8 +21,33 @@ def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
+def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
+def shouldUseNameSpace = agpVersion >= 7
+def PACKAGE_PROP = "package=\"com.learnium.RNDeviceInfo\""
+def manifestOutFile = file("${projectDir}/src/main/AndroidManifest.xml")
+def manifestContent = manifestOutFile.getText()
+if(shouldUseNameSpace){
+      manifestContent = manifestContent.replaceAll(
+        PACKAGE_PROP,
+        ''
+    )  
+} else {
+    if(!manifestContent.contains("$PACKAGE_PROP")){
+        manifestContent = manifestContent.replace(
+            '<manifest',
+            "<manifest $PACKAGE_PROP "
+        )
+    }
+}
+manifestContent.replaceAll("  ", " ")
+manifestOutFile.write(manifestContent)
+
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 30)
+    
+    if(shouldUseNameSpace){
+        namespace = "com.learnium.RNDeviceInfo"
+    }
 
     defaultConfig {
         minSdkVersion safeExtGet('minSdkVersion', 16)
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,32 +21,13 @@ def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
-def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
-def shouldUseNameSpace = agpVersion >= 7
-def PACKAGE_PROP = "package=\"com.learnium.RNDeviceInfo\""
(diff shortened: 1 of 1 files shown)


schie (commented):
Can you provide any documentation on if this change will affect applications using production react native (0.71)?

I may have overlooked or missed something

MasonLe2497:
I tested on RN 0.71.8. It work.

MasonLe2497:
This [comment](https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1597031450) confirmed rn 0.67 work

MasonLe2497:
@schie hi, I updated my PR, now it can support RN <0.71

felipemillhouse:
we can just keep the namespace on AndroidManisfest, it will not fail for RN 0.71+ just a warning message https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1597386954

MasonLe2497:
Yes, but i just remove the warning.

mikehardy (commented):
Interesting - this is a big chunk of compatibility code, but as near as I understand it, the goal is to either have namespace in AndroidManifest.xml *or* to have it in build.gradle

This is keyed here off the gradle version, which is technically correct as far as android gradle plugin goes, but isn't the react-native CLI part of the equation as well?

I wonder if it might be most compatible at the moment to simply leave AndroidManifest alone, and include the build.gradle namespace on AGP >=7 or not, such that there is always something there (AndroidManifest namespace) in case of old react-native CLI?

This has the downside of maybe leaving a warning for people because of the namespace in AndroidManifest, but I'm really leery of accidentally having a breaking change, or of having a complicated / subtle compatibility matrix of what version of this module works with what version of react-native + AGP

mikehardy:
I'm afraid this will fail in some environment as well, there are some environments that have read-only areas during the build and this requires the manifest file to be writeable

felipemillhouse:
@mikehardy 
Community are implementing it in multiple ways as I can see, for example:

Sentry: https://github.com/getsentry/sentry-react-native/blob/main/android/build.gradle#L18
react-native-gesture-handler: https://github.com/software-mansion/react-native-gesture-handler/blob/main/android/build.gradle#L171

or simply adding the namespace and accepting a warning message on build. A common sense is don't touch on AndroidManifest.xml keep it as it is.

mikehardy:
I went with this as the final version on netinfo module, I'm comfortable with the tradeoffs in it (it is very compatible, very simple, but will generate a warning)

https://github.com/react-native-netinfo/react-native-netinfo/blob/f3a0c2d70140aebd2337353cf298ebd0983fa2a1/android/build.gradle#L34-L38

```groovy
  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
  // Check AGP version for backward compatibility reasons
  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
    namespace = "com.reactnativecommunity.netinfo"
  }
```

What do you think? (obviously different namespace tho)

felipemillhouse:
Yeah, looks like people are going more with that solution

android/build.gradle
```diff
@@ -21,8 +21,33 @@ def safeExtGet(prop, fallback) {
...
+    if(!manifestContent.contains("$PACKAGE_PROP")){
+        manifestContent = manifestContent.replace(
+            '<manifest',
+            "<manifest $PACKAGE_PROP "
+        )
+    }
+}
+manifestContent.replaceAll("  ", " ")
+manifestOutFile.write(manifestContent)
+
```
mikehardy, line 44:
```suggestion
```

android/build.gradle
```diff
@@ -21,8 +21,33 @@ def safeExtGet(prop, fallback) {
...
 android {
     compileSdkVersion safeExtGet('compileSdkVersion', 30)
+    
+    if(shouldUseNameSpace){
+        namespace = "com.learnium.RNDeviceInfo"
+    }
```
mikehardy, line 50:
```suggestion
    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
    // Check AGP version for backward compatibility reasons
    if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
      namespace = "com.learnium.RNDeviceInfo"
    }
```

PR #1548, opened 2023-06-29
build(deps): refresh example to react-native 0.72

mikehardy:
## Description

Update the example app so that it is based on react-native 0.72

This should allow it to run on Windows even if you have the latest Visual Studio installed, which then transitively should allow me to test #1537 to make sure it works

Have tested android and iOS, working on windows etc

## Checklist


mikehardy:
Can't merge this until I figure out

https://github.com/react-native-device-info/react-native-device-info/actions/runs/5416532711/jobs/9847408770#step:10:5768

mikehardy:
Tried removing the setup-msbuild action to mirror more what the react-native-windows team does in their CI, just relying on the pre-installed Visual Studio toolchains on the windows-latest runner, but it did not help.

🤔 

Maybe there is something that needs fixing in the solution files, but I'm not sure

mikehardy:
With sincere apologies in case of irritation, I'm desperation-tagging @jonthysell after a mention in this comment https://github.com/react-native-device-info/react-native-device-info/pull/1537#issuecomment-1613930897 from @TatianaKapos where she mentioned you may have some institutional knowledge that could quickly fix the CI failure about missing packages here?

I am traveling so it sat for a while but I have time to collaborate + solve this now if that's possible.

My working hypothesis as I thought about it in the background over last few days is that the [example construction process](https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh) is perhaps carrying-forward something from previous react-native-windows versions that is the problem, and carrying in fresh project files from rn 0.72 might clear them out - but a close look at the script steps / related comments doesn't reveal anything to me, and it works for me locally, just not in CI. 

So I wonder if it could be as easy as some specific windows CI yaml command to just nuget a package (or however you do it) could fix this in 2 seconds, but I don't have the appropriate wisdom+skills currently

mikehardy:
@efstathiosntonas here is the current status of upgrading the example. There is some snag on getting it to run on windows for some reason. This branch is still at react-native 0.72 so it is behind, but it represents the state of the art. A windows machine (with functioning visual studio development environment) is strictly required in order to play with the example app unfortunately as react-native-device-info is used pretty heavily in the react-native-windows ecosystem and breaking it would be a bad thing...

mikehardy:
Still need to solve this bit:


> Build failed with message D:\a\react-native-device-info\react-native-device-info\example\node_modules\react-native-device-info\windows\RNDeviceInfoCPP\RNDeviceInfoCPP.vcxproj(152,5): error : This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is D:\a\react-native-device-info\react-native-device-info\example\windows\\packages\Microsoft.Windows.CppWinRT.2.0.210312.4\build\native\Microsoft.Windows.CppWinRT.props.. Check your build configuration.


...although it's possible if I bump things to current it will start working as this is still react-native 0.72.

So next steps - try bumping to current, then perhaps try to run the script it mentions in CI if that's possible, then call for help again

mikehardy:
going to switch to react-native-test-app, done it in a couple other repos now and it worked well

PR #1571, opened 2023-09-15
feat(android): add new isLowRamDevice API

### LINKED ISSUE #1568: Android Go
Hi, can we add Android Go check if it's a device?

https://stackoverflow.com/questions/58340558/how-to-detect-android-go

exzos28:
Added `isLowRamDevice()` that allows to check if the device is "isLowRam", this can be useful to determine if it is an Android GO device

Fixes #1568

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


mikehardy:
@exzos28 yep - the android e2e logs look like what happens when the emulator simply doesn't start at all. iOS passed fine. I imagine this will pass fine once the emulator behaves, I have restarted the failed job

exzos28:
@mikehardy emulator didn't start, what we can do?

mikehardy:
"what can we do" --> Have patience 😄 - restarting it again. It'll go through at some point I'm sure. I circle back to look at this every couple hours

mikehardy:
The tests themselves passed fine but another symptom that I take to mean "github is running kind of slowly in general" is that after the tests simply doing the gradle caching failed because disk access was slow and it couldn't hash files quickly enough. The important parts passed though, the PR is 💯

Issue #1035, opened 2020-06-25
Add events to know when location was enabled/disabled. Same for bluetooth/wifi


heroic:
# Feature request

Currently the package provides a way to know if location/bluetooth/wifi is available or not, but there is no event to subscribe to for when these are changed from outside the app.

## Why it is needed

If we need this functionality we need to use a different package for this, which makes the presence of isLocationEnabled, et all, functions useless from this package. Subscriptions to listen to changes to these hardware settings will make this package complete in the value provided.

## Possible implementation

Add subscriptions for hardware on/off events, much like how we have battery value update events.

mikehardy:
PRs happily accepted :-)

mikehardy:
if anyone wants to try this, the react-native-community/netinfo object shape for the event, event listener etc are a great model

heroic:
@mikehardy I'd like to help with this, but I'm afraid I'm not great with iOS

mikehardy:
Usually people are good with one or the other, I'm pretty sucky at Objective-C too. Propose an Android implementation and we can start from there, I won't let perfect get in the way of good

PR #1580, opened 2023-10-24
feat(android/ios): Add event to monitor location services are turned on/off

snazarkoo:
## Description

Partially covered the feature request #1035

Added `useIsLocationEnabled()` hook to monitor whether the device has location services are turned on/off 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -73,6 +73,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   private final DeviceIdResolver deviceIdResolver;
   private BroadcastReceiver receiver;
   private BroadcastReceiver headphoneConnectionReceiver;
+  private BroadcastReceiver locationServicesEnabledReceiver;
   private RNInstallReferrerClient installReferrerClient;
 
   private double mLastBatteryLevel = -1;
@@ -132,6 +133,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     };
 
     getReactApplicationContext().registerReceiver(receiver, filter);
+    initializeLocationServicesEnabledReceiver();
     initializeHeadphoneConnectionReceiver();
   }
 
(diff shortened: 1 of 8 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -151,6 +153,21 @@ public void onReceive(Context context, Intent intent) {
...
+        sendEvent(getReactApplicationContext(), "RNDeviceInfo_locationEnabledDidChange", isEnabled);
+      }
+    };
+
+    getReactApplicationContext().registerReceiver(locationServicesEnabledReceiver, filter);
+  }
+
 
   @Override
   public void onCatalystInstanceDestroy() {
```
mikehardy, line 173: should it stop listening on catalyst destroy? should probably also de-register if there are no listeners

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -76,6 +76,10 @@ - (id)init
     if ((self = [super init])) {
 #if !TARGET_OS_TV
         _lowBatteryThreshold = 0.20;
+
+        self.locationManager = [[CLLocationManager alloc] init];
+        self.locationManager.delegate = self;
```
mikehardy, line 81: I'm not sure we can do this in the `constantsToExport` area - that code is executed prior to the javascript portion of react-native even starting, meaning there is no ability to do this *optionally* via javascript code - it will run every time. Based on stackoverflow it looks like the locationmanager needs to set up on the  main thread but I think there are other ways to do that, if that is why it's in constantsToExport...

I don't believe app stores and users like it when apps start up location services as a mandatory item 🤔 - I may be misunderstanding but if I am understanding correctly then I think this should be initialized lazily only if ever subscribed, and should be removed both when all subscriptions are closed and on catalyst destroy

src/index.ts
```diff
@@ -1,7 +1,7 @@
 import { useCallback, useEffect, useState } from 'react';
 import { Dimensions, NativeEventEmitter, NativeModules, Platform } from 'react-native';
 import { useOnEvent, useOnMount } from './internal/asyncHookWrappers';
-import devicesWithDynamicIsland from "./internal/devicesWithDynamicIsland";
+import devicesWithDynamicIsland from './internal/devicesWithDynamicIsland';
```
mikehardy, line 4: do not include unrelated changes in a PR, it is very hard on maintainers - please revert *all* unrelated changes from this PR

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -881,6 +893,13 @@ - (long long) getFirstInstallTime {
     return [@(floor([installDate timeIntervalSince1970] * 1000)) longLongValue];
 }
 
+#pragma mark - CLLocationManagerDelegate
+
+- (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatus:(CLAuthorizationStatus)status
```
mikehardy, line 898: did a quick scan of stackoverflow and this one looks interesting! What was the behavior of the callback like in real life? Did it get a callback immediately? Was it correct (thus indicating threading is correct...) value on that first call? Subsequent calls? Curious what your results were

https://search.brave.com/search?q=didChangeAuthorizationStatus+CLAuthorizationStatus

mikehardy (commented):
Sorry for the delay! Thanks for this PR, looks useful to have this as a hook, but I had a few concerns I left in comments

Issue #1587, opened 2023-11-22
README documentation inconsistency around `getPowerState` and `usePowerState` / `RNDeviceInfo_powerStateDidChange`


simon-assistiq:
In the README `getPowerState()` returns an object with `batteryState`, `batteryLevel` and `lowPowerMode`.

But for `usePowerState` or `RNDeviceInfo_powerStateDidChange` it only returns whether the device is plugged in or not, which is called "batteryState" in `getPowerState`. And there is no listener listed called `useBatteryState` (and for that matter there isn't one for low power mode).

Is there a reason for this apparent inconsistency?

I suppose renaming at this point would be an annoying breaking change though.

mikehardy:
Inconsistent perhaps, but naming is hard (no one wants to code or help troubleshoot but everyone wants to suggest a name... see "bikeshedding" https://en.wiktionary.org/wiki/bikeshedding) naming is usually not the actual important thing over time as long as the docs explain exactly what everything does. Which they do I think.

There is useBatteryLevel which seems fine.

Is there a reason? Not really 🤷  people that propose PRs usually get to name things and if they seem reasonable at all I'm inclined to merge them because thank god sometimes people actually propose PRs and move open source stuff forward. I try not to get bogged down too much in PR review back and forth

simon-assistiq:
Seems actually `usePowerState` does in fact return the full object, but `RNDeviceInfo_powerStateDidChange` just returns the battery state (charging/not charging)

The comment `// 'charging'` next to the `usePowerState` code sample in the README implies otherwise.

mikehardy:
Ah! If that's the case a PR to the README to indicate what it's actually doing would be fantastic

simon-assistiq:
Worth noting my project is still on only 8.7.1 but I skimmed through the releases and didn't see anything obvious that would have affected this. Still, would need to test the latest release first.

simon-assistiq:
ok, seems like both give the power state object with the three properties on the latest version.
README adjustment still worthwhile, but at least it's relatively simple.

PR #1588, opened 2023-11-22
docs: Update README documentation for `usePowerState`

simon-assistiq:
On latest version (10.11.0) `usePowerState` and the listener both return a power state object, as opposed to just the battery state (e.g. 'charging')

* Fixes react-native-device-info/react-native-device-info#1587


simon-assistiq:
> Fired when the battery state changes or device enters in the power saving mode, for example when the device enters charging mode or is unplugged.

I wonder if this fires also when the battery level changes? I can test that out too.

PR #1595, opened 2023-12-16
feat: add isTabletMode for android

zhangyongz:
## Description

Added isTabletMode for Android that allows devices with foldable screen dynamically obtained mode when the screen sizes chagne.

## Compatibility

| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

* [√ ] I have tested this on a device/simulator for each compatible OS
* [√ ] I added the documentation in `README.md`
* [√ ] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [√] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ The example app in this repository shows an example usage of every single API, c
 | [isTablet()](#istablet)                                           | `boolean`           |  ✅  |   ✅    |   ✅    | ❌  |
 | [isLowRamDevice()](#istablet)                                     | `boolean`           |  ❌  |   ✅    |   ❌    | ❌  |
 | [isDisplayZoomed()](#isdisplayzoomed)                             | `boolean`           |  ✅  |   ❌    |   ❌    | ❌  |
-| [isTabletMode()](#istabletmode)                                   | `Promise<bool>`     |  ❌  |   ❌    |   ✅    | ❌  |
+| [isTabletMode()](#istabletmode)                                   | `Promise<bool>`     |  ❌  |   ✅    |   ✅    | ❌  |
 | [supported32BitAbis()](#supported32BitAbis)                       | `Promise<string[]>` |  ❌  |   ✅    |   ❌    | ❌  |
 | [supported64BitAbis()](#supported64BitAbis)                       | `Promise<string[]>` |  ❌  |   ✅    |   ❌    | ❌  |
 | [supportedAbis()](#supportedAbis)                                 | `Promise<string[]>` |  ✅  |   ✅    |   ✅    | ❌  |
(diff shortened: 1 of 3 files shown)


mikehardy:
I discourage the idea of "tablet" mode in general as it is not specific enough for a well-done responsive app.

What you really want is to dynamically determine the exact size of the screen then determine whether it is portrait or landscape for the long-axis, and exactly how big that width is so that you can decide between at least three layouts, small / medium / large which should each have pretty different default layouts.

That's responsive

How do you do that? `@react-native-community/hooks` has a `useDimensions` hook, you use that and make a layout hook perhaps, like so

```typescript
import {ScaledSize} from 'react-native';
import {useDimensions} from '@react-native-community/hooks';
import {useEffect, useState} from 'react';

export const enum LayoutSize {
  'small' = 'small',
  'medium' = 'medium',
  'large' = 'large',
}

export const enum LayoutOrientation {
  'landscape' = 'landscape',
  'portrait' = 'portrait',
}

export type LayoutInfo = {
  window: ScaledSize;
  screen: ScaledSize;
  orientation: LayoutOrientation;
  layoutSize: LayoutSize;
  hasBeenSmall: boolean; // FIXME required until drawer navigator re-renders correctly on mode switch from slide to permanent
};

export function useLayoutInfo(): LayoutInfo {
  const {screen, window} = useDimensions();
  let orientation = LayoutOrientation.portrait;
  let layoutSize = LayoutSize.medium;

  // if we are landscape, swap width and height
  if (window.width > window.height) {
    orientation = LayoutOrientation.landscape;
  }

  // get more specific with our layout breakpoints in case medium doesn't fit
  if (window.width <= 640) {
    layoutSize = LayoutSize.small;
  } else if (window.width >= 1008) {
    layoutSize = LayoutSize.large;
  }

  const [hasBeenSmall, setHasBeenSmall] = useState(
    layoutSize !== LayoutSize.large,
  );

  useEffect(() => {
    if (!hasBeenSmall && layoutSize !== LayoutSize.large) {
      setHasBeenSmall(true);
    }
  }, [layoutSize, hasBeenSmall]);

  const layoutInfo = {
    screen,
    window,
    orientation,
    layoutSize,
    hasBeenSmall,
  };

  return layoutInfo;
}

```

Then you have your components / screens lay themselves out based on LayoutInfo returned from the hook

You don't need this library to do that

zhangyongz:
Thanks for explaining. I got your point. While the concept you declared is reasonable, from another aspect, if we don't add a method to dynamically get the tablet mode, the [isTablet()] method has a problem in that its return value depends on the launch mode. When the app is launched in fold mode, the value will always be false; otherwise, it will always be true. I think we need an API for getting the accurate mode. @mikehardy

mikehardy:
I think the entire method should be removed, that is, go the other direction. The idea of "is it a tablet" is ill defined and not even useful if it were correct, now that is a dynamic concept the actual thing to do is have people get real dimensions with a hook and layout responsively 

Complicating this library further for what's really a design error in apps just doesn't seem like the right path

PR #1599, opened 2023-12-28
fix: wrap return value of useOnEvent in useMemo

itsramiel:
## Description
Since `useOnEvent` returns an object which gets recreated on every render, this makes the value returned unstable and cause bug by retriggering useEffects when the value did not really change.

Assume the following:
```
import {useEffect, useState} from 'react';
import {useIsHeadphonesConnected} from 'react-native-device-info';
function App() {
  const isHeadphonesConnected = useIsHeadphonesConnected();
  const [_, setCounter] = useState(0);

  useEffect(() => {
    const intervalId = setInterval(() => {
      setCounter(prev => prev + 1);
    }, 1000);

    return () => {
      clearInterval(intervalId);
    };
  }, []);

  useEffect(() => {
    // some intensive work
    console.log('isHeadphonesConnected', isHeadphonesConnected);
  }, [isHeadphonesConnected]);
  return null;
}

export default App;
```

Since `useIsHeadphonesConnected` uses `useOnEvent` this will make the useffect with the `isHeadphonesConnected` to retrigger every time a rerender happens which cause unexpected behavior

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/src/internal/asyncHookWrappers.ts
+++ b/src/internal/asyncHookWrappers.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useMemo } from 'react';
 import { NativeEventEmitter, NativeModules } from 'react-native';
 import type { AsyncHookResult } from './types';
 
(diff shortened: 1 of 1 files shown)


schie (approved):
LGTM

schie:
hm... I'm not sure why the tests haven't reported back yet. Nonetheless, I'll let it in.

schie:
@mikehardy, good to know. I'll try that next time.

PR #1603, opened 2024-01-26
feat(android): Added MEmu emulator check in isEmulator function

### LINKED ISSUE #1602: isEmulator() is not detecting the MEMU emulator
## Description

I would like the library to detect the MEmu Play emulator in the `isEmulator` function. This emulator can pass the validations that exist today.  Through the installed keyboard package I can validate it. I saw that other people have already needed to detect it at other times, I hope I'm helping.

## Use Case

Several emulators are being detected using the `isEmulator` function, but MEmu Play is not detected.

devdaniellima:
## Description

Fixes #1602

This change allows the MEmu Play emulator to be detected in the `isEmulator()` function.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -25,6 +25,8 @@ import android.os.BatteryManager;
 import android.os.Debug;
 import android.os.Process;
 import android.provider.Settings;
+import android.view.inputmethod.InputMethodManager;
+import android.view.inputmethod.InputMethodInfo;
 import android.webkit.WebSettings;
 import android.telephony.TelephonyManager;
 import android.text.TextUtils;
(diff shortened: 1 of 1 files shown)


fabiotakaki:
@devdaniellima I have been testing your branch and was really good solution! Thanks a lot for your support with this library. I tested on nox player because we have this check too but is return false on this function isEmulator. I found this PR being able to make additional checks on emulator cases #721 . Do you think is possible to enhance your solution expanding to another kinds of emulators?

devdaniellima:
@fabiotakaki I believe it can be improved, I thought about adding other validations by looking at some apps that are pre-installed on the emulator, but I ended up not doing it because of false positives. I saw your PR with some new validations for vbox and nox and found it very interesting.

mikehardy (approved):
Nice! Looks good to me - I rebased it to current master so it has all the CI work necessary to correctly check it in CI, and reworded the commit to be extremely
specific about what it does

mikehardy:
@fabiotakaki I'm going to merge this as is, if your project needs something else / something different, open a PR after this one goes in and if it's reasonable I can get it merged. Cheers

mikehardy:
passed android CI and it's an android-only change so merging

PR #1614, opened 2024-02-23
feat: add visionos platform support

gupta-ji6:
## Description

Added `visionOS` platform support for the library.

<details>
<summary>1. Constant</summary>

image

</details>

<details>
<summary>2. Sync</summary>

image

</details>

<details>
<summary>3. Async</summary>

image

</details>

<details>
<summary>4. Hooks</summary>

image

</details>

## Help

- [ ] `getFontScale`: [preferredContentSizeCategory](https://developer.apple.com/documentation/uikit/uitraitcollection/1771748-traitcollectionwithpreferredcont?language=objc) is supported in visionOS, but I need help with finding what can we use besides UIScreen. This [section of blog](https://www.oskarkwasniewski.dev/blog/bringing-react-native-libraries-to-apple-vision-pro#usage-of-uiscreen) might help.

## References
- [Bringing React Native libraries to Apple Vision Pro](https://www.oskarkwasniewski.dev/blog/bringing-react-native-libraries-to-apple-vision-pro)
- https://callstack.github.io/react-native-visionos-docs/

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS
* [x] I added the documentation in `README.md`
* [ ] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [x] I added a sample use of the API (`example/App.js`)

PR #1615, opened 2024-02-24
fix(getDeviceType): add TARGET_OS_VISION guard for UIUserInterfaceIdiomVision

gupta-ji6:
## Description

@UNIDY2002 raised a build failing issue [here](https://github.com/react-native-device-info/react-native-device-info/pull/1614#issuecomment-1962225526) due to visionos changes introduced by #1614.
I could also see same error in our github action ios build logs [here](https://github.com/react-native-device-info/react-native-device-info/actions/runs/8021357309/job/21913661609#step:12:299).

After referring to https://github.com/facebook/react-native/pull/42243/files#diff-ec1473a9d036148f566ce45864dfa02def7a9edc156862c9ea1ff8f374b59087 , I've added similar guardrail for our lib.

[`UIUserInterfaceIdiomVision`](https://developer.apple.com/documentation/uikit/uiuserinterfaceidiom/uiuserinterfaceidiomvision) is available only for iOS 17.0+.

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -134,7 +134,9 @@ RCT_EXPORT_MODULE();
             return DeviceTypeTablet;
         case UIUserInterfaceIdiomTV: return DeviceTypeTv;
         case UIUserInterfaceIdiomMac: return DeviceTypeDesktop;
+    #if TARGET_OS_VISION
         case UIUserInterfaceIdiomVision: return DeviceTypeHeadset;
+    #endif
         default: return DeviceTypeUnknown;
     }
 }


gupta-ji6:
the errors aren't coming now in our action logs as well, we should review & get this merged to fix the issue upstream.

https://github.com/react-native-device-info/react-native-device-info/actions/runs/8028925329/job/21934648504?pr=1615

mikehardy (approved):
Oh this is obvious in hindsight but I did not think of it during review. 🤦 
This is fix works - I think an `@available` would likely also work but *certainly* if we are targeting vision os we will have the minimum vision os symbols, so this works fine.

Thanks for listening in on the comments on the original PR and proposing this fix, that follow-up work is a big help

PR #1617, opened 2024-02-27
fix(android)!: removed deprecated getPhoneNumber - use react-native-sim-cards-manager if needed

### LINKED ISSUE #1612: Problem with submitting application to Google Play because of the usage of getLine1Number inside the library
## Summary
The issue lies in the utilization of TelephonyManager.getLine1Number() within the library's code ([accessible via this link](https://github.com/react-native-device-info/react-native-device-info/blob/e46a5d4dba4e7d1f5e1da8b17c82829600f84d77/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L919)).

The problem stems from receiving notifications from Google for several applications we are developing, prompting the implementation of Prominent Disclosure. This requirement arises due to the inclusion of this function within the codebase. We speculate that Google flags this through static code analysis. The issue surfaced during Google Play's application verification process, where Google's automated systems detected the function's presence. Despite not actively collecting phone numbers, the function's existence necessitates a Prominent Disclosure and a revision of our data collection policies. Attempting to address this, we forked the code and successfully published a version devoid of this function.

Consequently, we propose offering a version of the library without this function. Notably, the getLine1Number function is deprecated, and alternative methods for retrieving phone numbers from SIM cards are recommended. However, access to phone number retrieval typically requires specific "carrier" permissions or designation as the default "SMS application." Thus, we believe retaining this function may hinder the seamless publication of many applications with react-native-device-info library inside on Google Play.

| | |
|-----|---|
| Version | All versions |
| Affected OS | Android |

KatyaSmishlyaeva:
## Description

Fixes #1612 

The getPhoneNumber() method has been removed. Because this method uses deprecated getLine1Number function. It is recommended to use alternative methods to get phone numbers from SIM cards.  You can use https://github.com/odemolliens/react-native-sim-cards-manager instead.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,6 @@ The example app in this repository shows an example usage of every single API, c
 | [getManufacturer()](#getmanufacturer)                               | `Promise<string>`   |  ✅  |   ✅    |   ✅     | ❌   |   ✅     |
 | [getMaxMemory()](#getmaxmemory)                                     | `Promise<number>`   |  ❌  |   ✅    |   ✅     | ✅   |   ❌     |
 | [getModel()](#getmodel)                                             | `string`            |  ✅  |   ✅    |   ✅     | ❌   |   ✅     |
-| [getPhoneNumber()](#getphonenumber)                                 | `Promise<string>`   |  ❌  |   ✅    |   ❌     | ❌   |   ❌     |
 | [getPowerState()](#getpowerstate)                                   | `Promise<object>`   |  ✅  |   ✅    |   ✅     | ✅   |   ✅     |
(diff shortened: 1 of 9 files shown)


KatyaSmishlyaeva:
I didn't remove all TelephonyManager usage because only the getLine1Number method is deprecated.

mikehardy (approved):
Looks like exactly as discussed on the related issue, thank you!

Issue #1541, opened 2023-05-29
Missing Apple devices and their matching product names


florinbardosi:
For some Apple devices the fallback model name is returned instead of device model name

## Summary

In `ios/RNDeviceInfo /RNDeviceInfo.m`, the device model mapping should be updated.

Here is an updated list of Apple devices and their matching product names:
https://gist.github.com/adamawolf/3048717

## Current behavior

`iPad` is retuned instead of iPad (9th generation)

## Expected behavior

Device model name should be returned

PR #1619, opened 2024-03-04
fix: added missing device models

Catalin-Iagar:
## Description

Fixes #<https://github.com/react-native-device-info/react-native-device-info/issues/1541>

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist


mikehardy (commented):
sorry this went stale - I would be open to merging this if it did what it says "added missing device models"

However, have you looked at the diff? It's almost a diff of the full list. Why? I have no idea but as a reviewer that means you expect me to go line by line and see if any were accidentally dropped etc. The comments on the lines you altered seem less specific now, which is a -1

If a PR came in with a diff *purely added models* we could merge it

PR #1621, opened 2024-03-05
feat(android)!: targetSdk 34 requires receiver export specification during registration

andresesfm:
## Description

Targeting Android 14 requires the specification of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED when registering a receiver. Docs: https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported

using CompatConext allow us to avoid the android version check as it handles it internally

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

* [n/a (Only android) ] I have tested this on a device/simulator for each compatible OS
* [ n/a] I added the documentation in `README.md`
* [ n/a] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [ n/a] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,7 +22,7 @@ def safeExtGet(prop, fallback) {
 }
 
 android {
-    compileSdkVersion safeExtGet('compileSdkVersion', 30)
+    compileSdkVersion safeExtGet('compileSdkVersion', 34)
     
     def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
     // Check AGP version for backward compatibility reasons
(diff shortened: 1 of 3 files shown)


mikehardy:
This is important and relates to the PR just merged (which added some new runtime receivers) #1623

mikehardy (approved):
I modified this with an internal implementation of the ContextCompat register receiver 

it is a slightly better tradeoff to have the breaking change be a new requirement for a compile SDK of 33+ then to make sure people have the androidx.core transitive dependency (which they may not have had before...and which I couldn't get to work in the example app anyway)

This worked well in the example app once I bumped compile SDK to 33

Rebased on top of CI work so that should go green, pending merge

PR #1623, opened 2024-03-16
feat: Add isWiredHeadphonesConnected and IsBluetoothHeadphonesConnected

hrastnik:
## Description

This PR adds two new methods `isWiredHeadphonesConnected` and `isBluetoothHeadphonesConnected` and their sync versions. I've come across the need to check if I have a BT headset connected, but since the existing `isHeadphonesConnected` method checks for either BT or wired headset I've decided to extract the specific methods.

## Compatibility

Same as `isHeadphonesConnected`  - supports Android, iOS and VisionOS.

## Checklist


hrastnik:
Any change you can approve this? 
@mikehardy @machour @GantMan @Johan-dutoit @schie

mikehardy (approved):
sorry - haven't been paying attention to notifications here but you got me with the direct ping ;-)
does look well done, there is an opportunity to eliminate some code duplication by having the general headphone check delegate to the two new specific headphones checks but I consider that an optimization,  I can merge this now
would be cool to have a follow-up that de-duplicated that code by doing the internal method delegation though...

thanks!

mikehardy:
I am closing and reopening only because this typically kicks off a fresh attempt at CI

mikehardy:
CI was a mess. Cleaned it up with #1638 then rebased this on top of the merged result of that one. CI should go green now, if it does I can merge...

PR #1636, opened 2024-04-25
fix react-native 0.74 Use invalidate method instead onCatalystInstanceDestroy

a289459798:
fix react-native 0.74 Use invalidate method instead onCatalystInstanceDestroy

## Description

Fixes 0

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -155,7 +155,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
 
   @Override
-  public void onCatalystInstanceDestroy() {
+  public void invalidate() {
     getReactApplicationContext().unregisterReceiver(receiver);
     getReactApplicationContext().unregisterReceiver(headphoneConnectionReceiver);
   }


schie:
Hi @a289459798 ,

Do you have any documentation on this change?

a289459798:
onCatalystInstanceDestroy is remove in 0.74
so move onCatalystInstanceDestroy to invalidate

mikehardy (changes requested):
I saw this come up in another repo from you @a289459798 

I have the same request I had in comment https://github.com/react-native-netinfo/react-native-netinfo/pull/723#pullrequestreview-2051265099

I think they are easy to do, just need to pick one, alter the PR to go with the choice, and we can get this in

----

Hi there! This looks like a fix, yes, but what version of react-native added invalidate? 

It appears to be first added+released in react-native 0.65.0, and our README currently mentions support back to react-native 0.60 etc, so this is a breaking change

https://github.com/facebook/react-native/commit/18c8417290823e67e211bde241ae9dde27b72f17

Two choices to get this in:

1) avoid the breaking change by pushing an edit here where you keep the rename of the method, but you also keep an onCatalystInstanceDestroy override with an implementation that just delegates to the invalidate method with `this.invalidate()`

2) go ahead and do the breaking change, but include edits to the README that say the minimum supported version of react-native is now 0.65+, and any chunks of the README that have information related to versions less than 0.65 are removed

I actually prefer option 2. It's not a complicated breaking change to describe to people and those versions are 3 years old, if it affects anyone they really needed to upgrade anyway, and they can always just defer upgrading netinfo here

If you do that - make those changes to the README - we can get this released

mikehardy:
I went the backward-compatible route but I'm unable to push to this PR so I carried the changes into #1660

PR #1642, opened 2024-06-29
fix(ios): Make unique id, unique per device and not per ios account(k…

### LINKED ISSUE #1526: Invalid ID in getUniqueId in iOS devices after migration
## Generating the UID for iOS devices does not work properly. When a user migrate between devices the device UID remains the same, though it's a new device. 

|             |     |
| ----------- | --- |
| Version     | 10.5.1  |
| Affected OS | iOS |
| OS Version  | 6+ |

## Current behavior

When calling the `getUniqueId` method the UID calculation in iOS devices looks for the UIDKey in the keychain and uses it if exists. However, when storing the UIDKey, it is saved with the `kSecAttrAccessibleAfterFirstUnlock` attribute instead of the `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.
When using [kSecAttrAccessibleAfterFirstUnlock](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlock?language=objc) the value will be present after migration, hence making it not uniqued per device. 
If it would be saved using [kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessibleafterfirstunlockthisdeviceonly?language=objc) then it would not be passed upon migration. In Apple's words:
_Thus, after restoring from a backup of a different device, these items will not be present._

## Expected behavior

Upon migration, the `getUniqueId` method should generate a new ID(according to your code the IDFV) and not keep the one from the old device. 

## Suggested fix
Change the `CFStringRef` to `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` instead of `kSecAttrAccessibleAfterFirstUnlock`

humaidk2:
…eychain)

## Description

Fixes #1526

The unique id is persisted in the keychain using kSecAttrAccessibleAfterFirstUnlock.
This causes unexpected behaviors sometimes when using the same ios account in different devices.
The keychain syncs these variables and the device id is no longer unique.
To mitigate this, the breaking change is required to store the unique id per device only using the config variable
kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly instead.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1028,7 +1028,7 @@ DeviceInfo.getUniqueId().then((uniqueId) => {
 
 #### Notes
 
-> - iOS: This is [`IDFV`](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) or a random string if IDFV is unavaliable. Once UID is generated it is stored in iOS Keychain and NSUserDefaults. So it would stay the same even if you delete the app or reset IDFV. You can _carefully_ consider it a persistent, cross-install unique ID. It can be changed only in case someone manually override values in Keychain/NSUserDefaults or if Apple would change Keychain and NSUserDefaults implementations.
+> - iOS: This is [`IDFV`](https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor) or a random string if IDFV is unavaliable. Once UID is generated it is stored in iOS Keychain and NSUserDefaults. So it would stay the same even if you delete the app or reset IDFV. You can _carefully_ consider it a persistent unique ID. It can be changed only in case someone manually override values in Keychain/NSUserDefaults or if Apple would change Keychain and NSUserDefaults implementations.
 >   Beware: The IDFV is calculated using your bundle identifier and thus will be different in app extensions.
 > - android: Prior to Oreo, this id ([ANDROID_ID](https://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID)) will always be the same once you set up your phone.
 > - android: [Google Play policy](https://support.google.com/googleplay/android-developer/answer/10144311), see "persistent device identifiers". [Huawei - AppGallery Review Guidelines](https://developer.huawei.com/consumer/en/doc/30202) see "permanent device identifier" and "obtaining user consent".
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -274,7 +274,7 @@ PODS:
     - React-jsi (= 0.67.3)
     - React-logger (= 0.67.3)
     - React-perflogger (= 0.67.3)
-  - RNDeviceInfo (10.13.2):
+  - RNDeviceInfo (11.1.0):
     - React-Core
   - Yoga (1.14.0)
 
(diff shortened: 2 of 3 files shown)


humaidk2:
I thought about adding support for cross-install device id, just using an isShared variable
but that kinda complicates the docs and not sure you'd want that.

Also while developing,
I found few changes that could be implemented, let me know if you'd like prs for those
- Convert all objc modules to swift for easier maintenance
- Support new node js for example app
- Convert example app from class based to functional based

humaidk2:
@mikehardy your thoughts?

humaidk2:
related #1646

mikehardy (approved):
Tough one, definitely a breaking change, but it makes the behavior the same on android and ios (per device id, not per user)

Additionally with https://github.com/oblador/react-native-keychain it is possible for people to get a per-user id if they like, so I think people have a forward path if they relied on the per-user behavior

mikehardy:
@humaidk2 hi there 👋 

you had some other questions:

> Convert all objc modules to swift for easier maintenance

This would be great!

> Support new node js for example app
> Convert example app from class based to functional based

Spending time on the example app is difficult for me, I typically consider them both really important and also a waste of time (since the upgrade cost between react-native versions is something paid over and over in lots of repos)

So what I want to see for the example is either a PR with 2 commits, one of which is nothing but changes to https://github.com/react-native-device-info/react-native-device-info/blob/master/refresh-example.sh and the second one which is just the commit of the result of running the changes

I cannot accept a PR that just directly modifies the example app, because that commits us to a future course where we have to continue hand-modifying it

...or a switch to https://github.com/microsoft/react-native-test-app so that we no longer have to maintain anything.

I prefer using react-native-test-app these days but I'm not sure if it is feasible in this repo with all the different platforms supported

Aure77:
I don't think this issue is fixed on iOS. Due to the fallback on `[DeviceUID uid]`, the value in the keychain is not restored, but `NSUserDefaults` is still backed up. As a result, a new device will get the UID of the previous device from `NSUserDefaults`: https://github.com/humaidk2/react-native-device-info/blob/209b0a2fa3da223ca2dd328427e0818461a46956/ios/RNDeviceInfo/DeviceUID.m#L51
Therefore, the device UID is not unique across multiple devices.

mikehardy:
@Aure77 all reasonable PRs accepted, with bonus points for either backwards compatibility or really really really good commentary to include in the CHANGELOG for all us overworked devs that just need to know if it affects them, and what to do if so.

This isn't a use case of mine so with apologies, I have no time to donate for it. But I do merge PRs and release...

flogy:
Is it correct that after migrating to this library version and pushing out a new app version to my users, their current device IDs will be inevitably reset / regenerated after updating the app?

I did some tests and it looks like this is the case. For apps where the device ID is used for user authentication and similar, this would mean users will lose their data after the app update.

I think it might make sense to offer a flag that could be set to just keep the old mechanism without having to use [react-native-keychain](https://github.com/oblador/react-native-keychain). What do you guys think?

mikehardy:
> What do you guys think?

I'm the only current maintainer that does any work here at all and I have no interest in implementing it.

My previous comment stands


> all reasonable PRs accepted, with bonus points for either backwards compatibility or really really really good commentary to include in the CHANGELOG for all us overworked devs that just need to know if it affects them, and what to do if so.
> 
> This isn't a use case of mine so with apologies, I have no time to donate for it. But I do merge PRs and release...

PR #1644, opened 2024-07-02
feat: add getStartupTime

RikSchefferAmsterdam:
## Description

Added `getStartupTime()` that allows getting the startup time of the app in milliseconds since the unix epoch. 

This can for example be used to measure the time since app launch via `new Date().getTime() - getStartupTimeSync()`.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -971,6 +971,8 @@ RCT_EXPORT_METHOD(getStartupTime:(RCTPromiseResolveBlock)resolve rejecter:(RCTPr
     resolve(@(self.getStartupTime));
 }
  
+// Reads the process startup time and returns it in milliseconds since 1970
+// Reading the process startup time from the system is more accurate than comparing a date which is initiallized at launch with the current time
 - (long long) getStartupTime {
     size_t len = 4;
     int mib[len];
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -963,6 +963,29 @@ RCT_EXPORT_METHOD(getFirstInstallTime:(RCTPromiseResolveBlock)resolve rejecter:(
     return [@(floor([installDate timeIntervalSince1970] * 1000)) longLongValue];
 }
 
+RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD(getStartupTimeSync) {
+    return @(self.getStartupTime);
+}
+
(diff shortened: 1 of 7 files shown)


RikSchefferAmsterdam:
Is there anyone who can review this?

ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -963,6 +963,29 @@ - (long long) getFirstInstallTime {
...
+    struct kinfo_proc kp;
+
+    sysctlnametomib("kern.proc.pid", mib, &len);
+    mib[3] = getpid();
+    len = sizeof(kp);
+    sysctl(mib, 4, &kp, &len, NULL, 0);
+
+    struct timeval startTime = kp.kp_proc.p_un.__p_starttime;
+    double startTimeMilliSeconds = startTime.tv_sec * 1e3 + startTime.tv_usec / 1e3;
+    return [@(floor(startTimeMilliSeconds)) longLongValue];
```
schie, line 986: wouldn't the following be simpler?

```obj-c
NSDate *processStartDate = [[NSProcessInfo processInfo] systemUptime];
NSTimeInterval processStartTime = [[NSDate date] timeIntervalSince1970] - [processStartDate timeIntervalSince1970];
```

schie (dismissed):

ios/RNDeviceInfo/RNDeviceInfo.m
mikehardy: Agreed - @RikSchefferAmsterdam if you altered this to let the platform handle the mathematical gymnastics, this would be a lot easier to approve + merge

Alternatively, if you had a very good reason to implement it as you did, then leaving a comment specifically mentioning why you did not use NSProcessInfo and timIntervalSince1970 methods would be useful

mikehardy (changes requested):
Has a lot of potential - could be a useful method, but has a review comment to address

ios/RNDeviceInfo/RNDeviceInfo.m
RikSchefferAmsterdam: Sorry it took so much time to investigate this from my side.

My implementation retreives the starttime of the app process, which is more accurate than comparing the current time with the system startuptime. Specifically when you want to track the app launch time, you would want the starttime of the process. 

I have not found a way to retreive the app process startup time with NSProcessInfo.

I will add some comments in the code to explain this.

mikehardy: that is a reasonable defense of the style :-), works for me, thank you

mikehardy (approved):

PR #1649, opened 2024-08-01
fix(ios): isDisplayZoomed on iPhone 12/13 mini

### LINKED ISSUE #1581: isDisplayZoomed always returns true
## Summary

|             |     |
| ----------- | --- |
| Version     | 10.4.0 - 10.11.0 |
| Affected OS | iOS |
| OS Version  |16.6|

## Current behavior

Logging the result value from the function `isDisplayZoomed` on an iOS device always results in the value `true`.
Logging the result value on a simulator always results in false. This is because there is no display zoom support on the simulator.

I believe the value should change if you adjust the `Display Zoom` found in the settings `Display & Brightness` -> `Display Zoom` and switch between `Default` and `Larger Text`

## Expected behavior

`Display Zoom` with value `Default`:
Result value from `isDisplayZoomed` on an iOS device should be `false`.

`Display Zoom` with value `Larger Text`:
Result value from `isDisplayZoomed` on an iOS device should be `true`

## Needed for
This feature is useful to detect if the user is zooming their screen. If the function works correctly, we can optimize the UI for the user.

ffraaz:
## Description

Fixes #1581.

On iPhone 12 mini and iPhone 13 mini, `isDisplayZoomed()` always returns true, independent of whether `Display & Brightness` -> `Display Zoom` is enabled.

We fix this issue by changing the comparison operator (see [here](https://stackoverflow.com/a/65844985)).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -179,7 +179,7 @@ RCT_EXPORT_METHOD(getDeviceName:(RCTPromiseResolveBlock)resolve rejecter:(RCTPro
 
 - (BOOL) isDisplayZoomed {
     #if !TARGET_OS_VISION
-        return [UIScreen mainScreen].scale != [UIScreen mainScreen].nativeScale;
+        return [UIScreen mainScreen].scale < [UIScreen mainScreen].nativeScale;
     #else
         return NO;
     #endif


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -179,7 +179,7 @@ - (NSString *) getDeviceName {
 
 - (BOOL) isDisplayZoomed {
     #if !TARGET_OS_VISION
-        return [UIScreen mainScreen].scale != [UIScreen mainScreen].nativeScale;
+        return [UIScreen mainScreen].scale < [UIScreen mainScreen].nativeScale;
```
schie, line 182: @ffraaz, this is an interesting change. 

This implies that `.scale` is _can_ be greater than `.nativeScale` — am I understanding that correctly?

I haven't used this feature, so I don't know.

ffraaz: Yes, exactly. On iPhone 12 mini and iPhone 13 mini `scale` is `3.0` and `nativeScale` is `2.88` with default settings.
This leads to the false positives for `isDisplayZoomed` when using `!=` as the comparison operator.
From John Gruber's [iPhone 12 mini review](https://daringfireball.net/2020/11/the_iphone_12_mini_and_iphone_12_pro_max):
> The 12 Mini is also doing something else to pack more content onto the screen at a smaller size: its standard display resolution is ever so slightly scaled down...on the 12 Mini, the display is 2340 x 1080, but standard scale screenshots are 2436 x 1125 — a scaling factor of 96 percent. Software running on the 12 Mini “sees” the display as 2436 x 1125, but everything is downsampled on-the-fly to 2340 x 1080.

mikehardy: That is...kind of crazy. Wow! +1 for posting the docs 😆 makes it easy to understand at least

mikehardy (approved):
I closed/reopened the PR just to kick off CI
Change looks okay based on documentation

PR #1655, opened 2024-09-10
fix(android)!: powerStateDidChange returns full powerState object not batteryState value

### LINKED ISSUE #1651: Android RNDeviceInfo_powerStateDidChange does not return powerState object
## Summary

|             |     |
| ----------- | --- |
| Version     |  v11.1.0  |
| Affected OS | Android   |
| OS Version  | All   |

## Current behavior

The `RNDeviceInfo_powerStateDidChange` callback on Android [only sends](https://github.com/react-native-device-info/react-native-device-info/blob/98de4817c8911349023af5725f0a789dabb8641c/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L125) the `batteryState` instead of the entire `powerState` as described in the documentation. iOS [correctly sends](https://github.com/react-native-device-info/react-native-device-info/blob/98de4817c8911349023af5725f0a789dabb8641c/ios/RNDeviceInfo/RNDeviceInfo.m#L679) the entire object.

## Expected behavior

Android and iOS should provide the same object.

sumnii:
## Description

Fixes #1651

I expected the `usePowerState` or `RNDeviceInfo_powerStateDidChange` to return a `powerState` :object, but it was only returning an object at first and then only the `batteryState` value afterwards.
I have modified it to always return an object.

According to the documentation, it is supposed to return a `powerState` object, so I made functional changes without modifying the documentation.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -122,7 +122,12 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
         Boolean powerSaveState = powerState.getBoolean(LOW_POWER_MODE);
 
         if(!mLastBatteryState.equalsIgnoreCase(batteryState) || mLastPowerSaveState != powerSaveState) {
-          sendEvent(getReactApplicationContext(), "RNDeviceInfo_powerStateDidChange", batteryState);
+          WritableMap updatedPowerState = Arguments.createMap();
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
This is unfortunately a breaking change as people may be relying on the existing behavior, but it is a correct one, thank you for the fix!

PR #1657, opened 2024-09-12
feat(ios): add iPhone 16 device names / add missing iPad device names

### LINKED ISSUE #1541: Missing Apple devices and their matching product names

efstathiosntonas:
## Description

Added iPhone 16 support for notch and dynamic island. 

Devices codes picked from: https://gist.github.com/adamawolf/3048717

Fixes #1541
Supercedes #1619

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -279,6 +279,10 @@ RCT_EXPORT_METHOD(getDeviceName:(RCTPromiseResolveBlock)resolve rejecter:(RCTPro
         @"iPhone15,5": @"iPhone 15 Plus",
         @"iPhone16,1": @"iPhone 15 Pro",
         @"iPhone16,2": @"iPhone 15 Pro Max",
+        @"iPhone17,1": @"iPhone 16 Pro",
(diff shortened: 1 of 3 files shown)


efstathiosntonas:
mikehardy Hi, according to the gist's comments, this list is being picked from Xcode 16RC, so the weird numbering of the device codes should be correct 🤷

mikehardy (approved):
We have collaborated well in the past so I took the risk that you would not mind me piggy-backing on your PR to add the missing iPads and reorder the whole list to match the gist (so hopefully maintenance will be easier in future...) :-)

efstathiosntonas:
@mikehardy no worries!! Thanks, totally forgot about the iPads :face_palm:

PR #1664, opened 2024-09-29
feat(ios): implement all free disk storage types for ios 11+

### LINKED ISSUE #1148: getFreeDiskStorage invalide value

a-ghorbani:
## Description

Fixes #1148

This issue has been closed, but no action has been taken afaik. 
Please let me know if you want me to open a new issue. 

Apple's suggestion on how to get storage capacity: https://developer.apple.com/documentation/foundation/urlresourcekey/checking_volume_storage_capacity

## Checklist

* [ X] I have tested this on a device/simulator for each compatible OS (note: I have only tested on iOS version 11+)
* [ NA] I added the documentation in `README.md` 
* [ NA] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [NA] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -599,7 +599,6 @@ RCT_EXPORT_METHOD(getTotalDiskCapacity:(RCTPromiseResolveBlock)resolve rejecter:
 }
 
 - (double)getFreeDiskStorage {
-    uint64_t freeSpace = 0;
     NSError *error = nil;
 
     // iOS 11 and above: Use NSURLVolumeAvailableCapacityForImportantUsageKey
@@ -613,16 +612,12 @@ RCT_EXPORT_METHOD(getTotalDiskCapacity:(RCTPromiseResolveBlock)resolve rejecter:
             return 0;
         }
 
-        if (storageValues) {
-            NSNumber *availableCapacityForImportantUsage = storageValues[NSURLVolumeAvailableCapacityForImportantUsageKey];
-            if (availableCapacityForImportantUsage) {
-                freeSpace = [availableCapacityForImportantUsage unsignedLongLongValue];
-            }
+        NSNumber *availableCapacityForImportantUsage = [storageValues objectForKey:NSURLVolumeAvailableCapacityForImportantUsageKey];
+        if (availableCapacityForImportantUsage) {
+            return (double)[availableCapacityForImportantUsage unsignedLongLongValue];
         }
-    }
-
-    // Fallback for older iOS versions: Use NSFileSystemFreeSize
-    if (freeSpace == 0 && !error) {
+    } else {
+        // Fallback for older iOS versions: Use NSFileSystemFreeSize
         NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
         NSDictionary *dictionary = [[NSFileManager defaultManager] attributesOfFileSystemForPath:[paths lastObject] error:&error];
 
@@ -631,13 +626,13 @@ RCT_EXPORT_METHOD(getTotalDiskCapacity:(RCTPromiseResolveBlock)resolve rejecter:
             return 0;
         }
 
-        if (dictionary) {
-            NSNumber *freeFileSystemSizeInBytes = [dictionary objectForKey:NSFileSystemFreeSize];
-            freeSpace = [freeFileSystemSizeInBytes unsignedLongLongValue];
+        NSNumber *freeFileSystemSizeInBytes = [dictionary objectForKey:NSFileSystemFreeSize];
+        if (freeFileSystemSizeInBytes) {
+            return (double)[freeFileSystemSizeInBytes unsignedLongLongValue];
         }
     }
 
-    return (double)freeSpace;
+    return 0;
 }
 
 
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -598,23 +598,24 @@ RCT_EXPORT_METHOD(getTotalDiskCapacity:(RCTPromiseResolveBlock)resolve rejecter:
     resolve(@(self.getTotalDiskCapacity));
 }
 
-- (double)getFreeDiskStorage {
+- (double)getFreeDiskStorage:(NSString *)storageType {
     NSError *error = nil;
 
     // iOS 11 and above: Use NSURLVolumeAvailableCapacityForImportantUsageKey
     // https://developer.apple.com/documentation/foundation/urlresourcekey/checking_volume_storage_capacity
     if (@available(iOS 11.0, *)) {
         NSURL *fileURL = [NSURL fileURLWithPath:NSHomeDirectory()];
-        NSDictionary *storageValues = [fileURL resourceValuesForKeys:@[NSURLVolumeAvailableCapacityForImportantUsageKey] error:&error];
+        NSString *capacityKey = [self keyForStorageType:storageType];
+        NSDictionary *storageValues = [fileURL resourceValuesForKeys:@[capacityKey] error:&error];
 
         if (error) {
             NSLog(@"Error retrieving storage information: %@", error);
             return 0;
         }
 
-        NSNumber *availableCapacityForImportantUsage = [storageValues objectForKey:NSURLVolumeAvailableCapacityForImportantUsageKey];
-        if (availableCapacityForImportantUsage) {
-            return (double)[availableCapacityForImportantUsage unsignedLongLongValue];
+        NSNumber *availableCapacity = [storageValues objectForKey:capacityKey];
+        if (availableCapacity) {
+            return (double)[availableCapacity unsignedLongLongValue];
         }
     } else {
         // Fallback for older iOS versions: Use NSFileSystemFreeSize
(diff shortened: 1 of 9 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -598,17 +598,49 @@ - (double) getTotalDiskCapacity {
...
+            NSNumber *availableCapacityForImportantUsage = storageValues[NSURLVolumeAvailableCapacityForImportantUsageKey];
+            if (availableCapacityForImportantUsage) {
+                freeSpace = [availableCapacityForImportantUsage unsignedLongLongValue];
+            }
+        }
     }
-    return (double) freeSpace;
+
+    // Fallback for older iOS versions: Use NSFileSystemFreeSize
+    if (freeSpace == 0 && !error) {
```
mikehardy, line 625: This control flow is confusing to me.
I understand the desire to avoid if/else and do quick-return short-circuit style conditional constructions, but this one avoids the if/else but still doesn't short-circuit, which gives me a high mental load when reading the code

I'd prefer it to be a traditional if/else where the `if(@available...` is the same above, this is now an else that isn't keyed on error existing or not or freespace already being non-zero or not

then each branch of the conditional could simply return the value it sees directly or the error directly instead of storing it in method scope

mikehardy (changes requested):
Hi there 👋 
Interesting! I had not paid attention to how the APIs had moved in iOS, this appears to be a fine pull request

I had some comments on the specific key used and on the structure but in general no real issue with it though they need to be handled pre-merge

As an opinion question though: these are sensitive APIs per Apple rules and they will require declaration in the privacy manifest. That's a breaking change - which is fine, I have no problem releasing a breaking change as long as it is documented well. The opinion question is - as a fellow Apple developer - would you prefer the library itself has a Privacy manifest? Or should we leave that up to the app developer at the app level? I'm not sure how devs prefer to do it with cocoapods-integrated source libraries, or how other react-native libraries are doing it

I think the react-native modules just avoid it and things are declared at the app level though, and we can just point to the documentation for the APIs we use

a-ghorbani:
As for 'sensitive APIs,' that's a good question, personally, I feel that leaving it up to the app developer seems like a flexible approach. but, I am not a professional coder, so unsure how most RN module developers handle that. I asked Sonnet 3.5 and ChatGPT 4o, and they don’t seem to agree on what 'most module developers' do,which I expected them to agree, since the training data would mostly be the same (e.g., GitHub, etc.) :)

**ChatGPT 4o**
Screenshot 2024-10-01 at 08 38 25

**Claude Sonnet 3.5**
Screenshot 2024-10-01 at 08 37 57


If you agree to leave it up to the developers, happy to put a note in the documentation.

mikehardy:
We are leaving it up to the developer up to this point and I'm not sure we could ever explain correctly what the API is used for since we're just wrapping+exposing it, so leaving it to the developer is the most flexible way. Not having a strong opinion about it is fine, I guess I was mostly asking just in case you did you have a strong opinion and good reasoning, then I'd learn something and behave accordingly ;-)

A note in the documentation on iOS integration mentioning that "we use sensitive APIs (for example NSURLVolumeAvailableCapacityForImportantUsageKey) and at the app level you will need to include + explain the usage in your Privacy manifest" should do it

That reminds me - I thought I asked but can't see it now - why did you choose NSURLVolumeAvailableCapacityForImportantUsageKey vs NSURLVolumeAvailableCapacityKey ? It seems NSURLVolumeAvailableCapacityKey should be the total amount whereas "for important usage" would be some subset potentially and not the whole amount?

ios/RNDeviceInfo/RNDeviceInfo.m
mikehardy: really easy to understand now, I appreciate it

a-ghorbani:
a couple of points:

1) this [post](https://developer.apple.com/documentation/foundation/urlresourcekey/checking_volume_storage_capacity) for *Checking Volume Storage Capacity* from Apple doesn’t mention `NSURLVolumeAvailableCapacityKey` in the body of the post (only on see also). It says:  
> "If you’re storing data based on a **user request** or resources the app requires to function properly (for example, a video the user is about to watch or resources that are needed for the next level in a game), query against [volumeAvailableCapacityForImportantUsageKey](https://developer.apple.com/documentation/foundation/urlresourcekey/2887126-volumeavailablecapacityforimport)."

2) it seems that `NSURLVolumeAvailableCapacityKey` is equivalent to the old API ( https://forums.developer.apple.com/forums/thread/748619).

3) also`NSURLVolumeAvailableCapacityForImportantUsageKey` appears to be the closest match to what is shown in the system settings  ([here](https://stackoverflow.com/a/49214836/2911976 ) and [here](https://github.com/itinance/react-native-fs/issues/1128#issuecomment-1929865526) ).

There are 3 options that we can go about this:
1) If system settings use `NSURLVolumeAvailableCapacityForImportantUsageKey`, then we should use it as well.
2) Provide an argument to let the users choose.
3) Expose both APIs, like `getFreeDiskStorage` and `getFreeDiskStorageForImportantUsage`.

In my opinion, we should keep it simple and go with option 1, wdyt?

(We could also consider adding `volumeAvailableCapacityForOpportunisticUsageKey` into the mix :) )

mikehardy:
Wow. This situation is bonkers. These APIs are objectively stupid and user unfriendly because they provide information to apps that apps can't really use to help users perform their tasks.

This thread you link on the forums in particular:

> Note It’s likely that none of these will yield exactly the same results as you get in Settings. There is no API to get exactly the values you see in user-level locations like Settings, Xcode, and Finder. Each of these calculates usage in their own unique way.

So...there's no API which will match what the users see in Settings, so they'll be confused to start as you can only provide contradictory information, and in the end - there was no recommendation on the thread helping apps make decisions on what do propose to users when it looks like there are 110GB free (for example) but they can't save a 10GB video.

Ridiculous.

But that's not your fault!

I prefer this module just expose what the system provides so we don't have to fight battles over opinions. We trust app developers to use the information well (privacy permitting!) and we just want to convey device information as purely as possible.

Option 3 would mean a proliferation of APIs so I don't like that so much.

> Provide an argument to let the users choose.

I like this option. Especially if the option is optional and defaults to the previous behavior - that is the "total" key that maps to old behavior.

Then define some constants like "important", "opportunistic" and let devs send those in, switch on them and use the other URL keys if provided?

I don't mean to over-engineer this, and I'm sorry already on your behalf that this has taken as much back and forth as it has. It's a "disk available" API it should be easy right? But nope, everything is subtle apparently, and mobile disk availability is too now it seems.

a-ghorbani:
hey, sorry for the silence for a few days, but I finally got the chance to work on this.

The `getFreeDiskStorage` function now accepts an option ('storageType') with three values: 'total,' 'important,' and 'opportunistic.'
On non-iOS devices, the param will be ignored. 

If I understood the code logic correctly, I had to change `type Getter<T> = () => T` (and related parts like `getSupportedPlatformInfoFunctions`) to support getter functions that can accept params like `getFreeDiskStorage`.

mikehardy (approved):
That's pretty amazing. Seems to work perfectly when I check it locally, code looks great
Nice!

mikehardy:
Ah - there is some subtle thing happening now with the getters and whether the module is gating native module access correctly

When I look through my local packager log from testing I do see something:

>  LOG  Trouble getting device info  [TypeError: _nativeInterface.default.getHostNames is not a function. (In '_nativeInterface.default.getHostNames()', '_nativeInterface.default.getHostNames' is undefined)]

...indicating that some getters are going native when they shouldn't, and the jest failure is related - it is probably the easiest reproduction

mikehardy:
I got it working - it was a subtle change to the async getter method - basically it had to be a little less fancy then you made it, and now it appears to work well - jest tests pass and on Simulator you get the same results every time which concerned me but then I brought it up on a real device and sure enough I get different total amounts

👍

schie (approved):
:shipit:

mikehardy:
lint caught me on an unused args thing and very very close inspection on a real device of sync vs async showed I still wasn't setting up the args correctly into the async function (which is why they were unused, of course)

Very carefully re-did that part, and now it passes inspection on a real device, sync and async, and lint, and jest

good to go!

added a tiny follow-on with a `yarn lint` run, that had decayed a little bit

schie:
I'll wait for the linter this time :)

schie:
Android e2e failed - gonna run it again JIC it was a fluke

schie:
The android emulator is just timing out. I'm fine with merging

mikehardy:
@schie ubuntu-latest just moved from ubuntu-22 to ubuntu-24 and it messes up the way the android emulator works, I need to fix that - related fun: https://github.com/ReactiveCircus/android-emulator-runner/issues/400

``` 
ERROR   | HOME is defined but there is no file TestingAVD.ini in $HOME/.android/avd
ERROR   | (Note: Directories are searched in the order $ANDROID_AVD_HOME, $ANDROID_SDK_HOME/avd and $HOME/.android/avd)
```

that runner is what I use most of the time now, so as soon as a new release of that action is published I'd purge the whole section of android start/retry/retry and just use that action

PR #1666, opened 2024-10-04
docs(README): typo fix in sample code

paulo-hortelan:
Just a missing parenthesis that could mislead someone who is distracted


mikehardy (approved):
ha! you are right. nice, thank you

Issue #1671, opened 2024-10-24
RNDeviceInfo is adding `AD_ID` permission through `play-services-iid` ?


dprevost-LMI:
## Summary

|             |     |
| ----------- | --- |
| Version     |  10.14.0   |
| Affected OS | Android   |
| OS Version  | 14 |

## Current behavior
Hi, we upgraded to SDK 34 and then uploaded a build manually to the Google Store. We got a consent form request for using `com.google.android.gms.permission.AD_ID` permission.

After using `./gradlew app:dependencies > dependencies.txt` and looking at `manifest-merger-blame-gotoconnect-debug-report.txt`, I was able to find the lib `play-services-iid-17.0.0` pulling this permission `AD_ID`. It comes from `react-native-device-info` pulling it from the [build.gradle](https://github.com/react-native-device-info/react-native-device-info/blob/34a80b6e1516f04f251a50273c41b1f3fb39e1f7/android/build.gradle#L67-L70)

I commented the deps to be sure and then ended up seeing `firebase-messaging-23.2.1` pulling the same permission, but that is for another repository 😉 

## Expected behavior

`react-native-device-info` should not add the `com.google.android.gms.permission.AD_ID` permission even through it's dependencies.

## More Comments

From the `build.grade`, I can see that we either pull the Google Play service or the Firebase one. I'm not sure where and how I can configure that to try to stop pulling `implementation "com.google.android.gms:play-services-iid:$iidVersion"` and fix my issue

## Logs
relevant content of `/manifest-merger-blame-gotoconnect-debug-report.txt`
```
109-->[com.google.android.gms:play-services-iid:17.0.0] [obfucated]/.gradle/caches/transforms-3/eabf05445f53c5e97813df2a1d4cf2cb/transformed/jetified-play-services-iid-17.0.0/AndroidManifest.xml:23:5-82
109-->[com.google.android.gms:play-services-iid:17.0.0] [obfucated].gradle/caches/transforms-3/eabf05445f53c5e97813df2a1d4cf2cb/transformed/jetified-play-services-iid-17.0.0/AndroidManifest.xml:23:22-79
110    <uses-permission android:name="com.google.android.gms.permission.AD_ID" />
```

gradle relevant dependency tree `./gradlew app:dependencies > dependencies.txt`:

```
+--- project :react-native-device-info
|    +--- com.facebook.react:react-native:+ -> com.facebook.react:react-android:0.73.9 (*)
[...]

PR #1673, opened 2024-10-25
fix: RNDeviceInfo is adding AD_ID permission through play-services-iid

mrpmohiburrahman:
## Description

Fixes https://github.com/react-native-device-info/react-native-device-info/issues/1671

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -64,9 +64,6 @@ dependencies {
       implementation "com.google.firebase:firebase-iid"
   } else if(firebaseIidVersion){
       implementation "com.google.firebase:firebase-iid:${firebaseIidVersion}"
-  }else{
-      def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
-      implementation "com.google.android.gms:play-services-iid:$iidVersion"
   }
 
   testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -59,11 +59,15 @@ dependencies {
   implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1.2')}"
   def firebaseBomVersion = safeExtGet("firebaseBomVersion", null)
   def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
+  def googlePlayServicesIidVersion = safeExtGet('googlePlayServicesIidVersion', null)
+
   if (firebaseBomVersion) {
       implementation platform("com.google.firebase:firebase-bom:${firebaseBomVersion}")
       implementation "com.google.firebase:firebase-iid"
   } else if(firebaseIidVersion){
       implementation "com.google.firebase:firebase-iid:${firebaseIidVersion}"
+  } else if(googlePlayServicesIidVersion){
+      implementation "com.google.android.gms:play-services-iid:$googlePlayServicesIidVersion"
   }
 
   testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.0'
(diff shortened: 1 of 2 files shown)


PaperbagWriter (approved):

schie:
@mrpmohiburrahman 

Thanks for the PR!

I'll review in a few.

android/build.gradle
```diff
@@ -64,9 +64,6 @@ dependencies {
       implementation "com.google.firebase:firebase-iid"
   } else if(firebaseIidVersion){
       implementation "com.google.firebase:firebase-iid:${firebaseIidVersion}"
-  }else{
-      def iidVersion = safeExtGet('googlePlayServicesIidVersion', safeExtGet('googlePlayServicesVersion', '17.0.0'))
-      implementation "com.google.android.gms:play-services-iid:$iidVersion"
```
mikehardy, line 69: You have made this change so that it is not even possible to make this backwards compatible for people by removing this logic completely

The more developer-friendly way to do this is not to just delete everything and YOLO a release because it does not affect yout

The more developer-friendly way is to at least make it so developers can maintain the old functionality if they need it

Move the safeExtGet of `googlePlayServicesIidVersion` out of the if/else if/else area, put it up with the other two, and have it also default to null if it does not exist

Then change the final `else` to an else if (googlePlayServicesIidVersion) and pull in the dependency if it was non-null

Leave the code below that allows fetching GMS version if people did set a googlePlayServicesIidVersion

Now in the changelog we can issue a breaking change *but* we can say:

"Np longer add GMS Instance ID fetch as default fallback. If you relied on this functionality, update your gradle build file to define `googlePlayServicesIidVersion` if you have not already, previously it defaulted to `17.0.0`"

Which is much better than "If you relied on this functionality, you can't update this module ever"

mikehardy (changes requested):
thanks for taking the time to get this started - I do strongly prefer a backwards-compatible-possible change - left specific direction on how to achieve that..

android/build.gradle
mrpmohiburrahman: Hi, @mikehardy ,
Thanks for the detailed feedback!
I've made all the suggested changes and updated the PR accordingly.

mikehardy (approved):
this looks great now - hopefully won't trigger any rejections or bad notices from play stores, but still gives any developers with this use case an option to handle it

Thanks

mikehardy:
Proposed commit message (while waiting for CI):

----

fix(android)!: remove AD_ID permission requirement by removing auto-addition of play-services-iid

BREAKING CHANGE: removed default fallback getDeviceId via play-services-iid dependency

This dependency has the unfortunate side effect of including AD_ID permission, which is
not permissible for many types of applications.

If you relied on this functionality, you will need to define `googlePlayServicesIidVersion` in your
android gradle build files ext version block. We previously used version 17.0.0 but there may be
newer versions you could use if they work for

mrpmohiburrahman:
> Proposed commit message (while waiting for CI):


Just to clarify, Should I amend the last commit message using 
```git commit --amend```

**Comment by mikehardy** [comment]
No - not necessary, I do rebase-merge for some repos I run and that would be appreciated, but this repo we're fine squashing so I get a chance to splat a commit message in there exactly as I like :)

**Comment by mikehardy** [comment]
passed android CI already ✅  so merged it in, will kick the release off if it doesn't go automatically

File: android/build.gradle [hunk 2]
```diff
@@ -59,14 +59,15 @@ dependencies {
   implementation "com.android.installreferrer:installreferrer:${safeExtGet('installReferrerVersion', '1.1.2')}"
   def firebaseBomVersion = safeExtGet("firebaseBomVersion", null)
   def firebaseIidVersion = safeExtGet('firebaseIidVersion', null)
+  def googlePlayServicesIidVersion = safeExtGet('googlePlayServicesIidVersion', null)
+
   if (firebaseBomVersion) {
       implementation platform("com.google.firebase:firebase-bom:${firebaseBomVersion}")
       implementation "com.google.firebase:firebase-iid"
```
efstathiosntonas, line 66: @mikehardy hi, after setting `firebaseBomVersion` on build.gradle I get:

```
Could not determine the dependencies of task ':react-native-device-info:compileDebugJavaWithJavac'.
> Could not resolve all dependencies for configuration ':react-native-device-info:debugCompileClasspath'.
   > Could not find com.google.firebase:firebase-iid:.
     Required by:
         project :react-native-device-info
```

android/build.gradle:

```
buildscript {
    ext {
        googlePlayServicesIidVersion = "17.0.0"
        firebaseBomVersion = "33.14.0"
    }
```

i believe that if we set `firebaseBomVersion` we should also set the `FirebaseIidVersion` too, right?

PR #1684, opened 2024-12-10
fix(tvos): Fix compilation issues due to missing apis in keyForStorageType for tvos (#1670)

### LINKED ISSUE #1670: Version 13.1.0 getFreeDiskStorage breaks tvOS support
## Summary

Since Version 13.1.0 the library does not support tvOS anymore since the getFreeDiskStorage implementation uses unsupported classes on tvOS.
The previous version 13.0.0 works fine.

|             |     |
| ----------- | --- |
| Version     | 13.1.0   |
| Affected OS | tvOS  |
| OS Version  | 18.0   |

## Current behavior

Compile fails on tvOS, as unsupported classes are used.
image

## Expected behavior

Compile works again on tvOS.

adrianrdzv:
## Description

- Fixes #1670. Apis are [not available for tvos](https://developer.apple.com/documentation/foundation/urlresourcekey/checking_volume_storage_capacity).
- Fixes compilation issue: `sysctl` needs to be imported for tvos as well.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ✅     |

## Checklist

### DIFF
--- a/ios/RNDeviceInfo/RNDeviceInfo.m
+++ b/ios/RNDeviceInfo/RNDeviceInfo.m
@@ -37,9 +37,7 @@ typedef NS_ENUM(NSInteger, DeviceType) {
 @import CoreTelephony;
 #endif
 
-#if !TARGET_OS_TV
 @import Darwin.sys.sysctl;
-#endif
 
 @implementation RNDeviceInfo
 {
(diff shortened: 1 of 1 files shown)


ios/RNDeviceInfo/RNDeviceInfo.m
```diff
@@ -37,9 +37,7 @@ typedef NS_ENUM(NSInteger, DeviceType) {
 @import CoreTelephony;
 #endif
 
-#if !TARGET_OS_TV
```
schie, line 40: Why was this removed?

schie (commented):
Overall, I think this is fine. Just have the one question to be answered (I'm not up-to-date on tvOS)

ios/RNDeviceInfo/RNDeviceInfo.m
adrianrdzv: Due to the introduction of [getStartupTime](https://github.com/react-native-device-info/react-native-device-info/blob/master/ios/RNDeviceInfo/RNDeviceInfo.m#L1062) and the usage of `kinfo_proc`. Now tvos also requires this import. I updated the PR description.

mikehardy (approved):
LGTM - thank you!

Issue #1647, opened 2024-07-27
Implement Rooted Device Check 


5cientist:
## Description

Implement a feature to check if a device is rooted or jailbroken 

let isRooted = DeviceInfo.isRooted();
// true

## Use Case

When users launch the app, the feature will run a quick check to detect any rooted or jailbroken devices. If a rooted device is identified, appropriate steps will be taken to ensure the security and integrity of the app environment.

zaferatli:
Is it on the plan, i can create this feature @mikehardy

mikehardy:
Nothing is on the plan, it's open source
Propose a PR

Issue #593, opened 2019-03-14
Can I detect a jailbreak?


envyTang:
## Can I detect a jailbreak?

react-native-cli: 2.0.1
react-native: 0.57.1
npm: 5.5.1

machour:
No

Issue #553, opened 2018-12-25
how to increase the judge whether root or mobile phone jailbreak function


suwu150:
how to increase the judge whether root or mobile phone jailbreak function

machour:
Please use the provided issue template

PR #1690, opened 2025-01-30
Jailbroken/Rooted Device Detection Feature

zaferatli:
## Description

Added `isRootedDevice()` that catches rooted/jailbroken devices. #1647 #593 #553 
Added `rootedReason()` that returns rooted/jailbroken reason.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


mikehardy:
🤔 hmm - my thinking on this has evolved somewhat as I've been more involved with firebase and app-check, which also attempts to verify if you can trust a device.

My current thinking has a few parts:
- this is a very hard problem and all you have signals indicating more or less trust. You cannot say "isRooted", or perhaps you can say that but you definitely cannot say "is *not* rooted". It is very difficult to assert a device may be trusted and I could not in good faith publish an API that gives people false confidence
- there are problems with app stores rejecting apps for simply including code. This happens a lot in China but also happens with Google and Apple - I have a serious worry that even including this code in a major module like react-native-device-info will cause false positives and app rejections for a lot of developers

So with regret I've got two reasons I don't want to include this PR though it seems to be well-written

However, I understand the need (I do maintain `@react-native-firebase/app-check` after all...) - luckily there *is* already a maintained package here - and I strongly encourage you to join forces with the maintainer - https://github.com/GantMan/jail-monkey

I like the idea of a specific package focused on this much more than including it here, so that only people who really seek it out and include it are at risk of app store rejections, and it should be easy to collaborate on a smaller package

anta40:
Looks like the commits are not merged yet?

I just installed this lib (still using RN 0.71.0):
```
import DeviceInfo from 'react-native-device-info';

let isRooted = await DeviceInfo.isRootedDevice();

```

The error is:
`Property 'isRootedDevice' does not exist on type 'DeviceInfoModule'`

PR #1691, opened 2025-02-03
fix(android): restore compatibility for Android <= 23

mikehardy:
## Description

the minSdk for supported versions of react-native will be 24 once react-native 0.78 ships, but there is no reason not to maintain compatibility with 23 right now

however, the API the process uptime feature depends on doesn't exist before then so just return our "undefined" numeric value of -1

Fixes unlogged issue but noticed it when I went to upgrade react-native-firebase, which depends on this module but has minSdk 23 *and* a NewAPI lint check (which apparently we don't have here so we didn't catch it...)

## Checklist

* [x] I have tested this on a device/simulator for each compatible OS - yes, checked API23 sim, crash reproduced, crash fixed

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -772,15 +772,19 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
 
   @ReactMethod(isBlockingSynchronousMethod = true)
   public double getStartupTimeSync() {
-    // Get time in milliseconds since unix epoch
-    long currentTime = System.currentTimeMillis();
-    // Get the time when the process started in milliseconds since system boot
-    long processStartTime = Process.getStartUptimeMillis();
-    // Get the milliseconds since system boot time
-    long currentUptime = SystemClock.uptimeMillis();
-    // Calculate the process startup time in milliseconds since unix epoch
-    long startupTime = currentTime - currentUptime + processStartTime;
-    return BigInteger.valueOf(startupTime).doubleValue();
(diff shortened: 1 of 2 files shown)


mikehardy:
Note for anyone that does not like the -1, the appropriate way to fix this is to make a ContentProvider (since these are started very early in Android Process startup) and add that to the AndroidManifest for the module. In that ContentProvider capture the system time when the ContentProvider was started, and then in our module, pull that to use as our startup time.

This isn't a use case of mine so I was not going to volunteer the time to do it for such an old platform compatibility reason, but if you are interested I would take a PR.

In the meantime, getting a -1 which we document as "undefined for numbers" is something anyone can handle, vs a crash.

schie (approved):

mikehardy:
I always appreciate the extra set of 👀  - thanks @schie 

iOS has some unrelated failure with boost pod checksum (but...I did not touch iOS)

Hopefully a Podfile lock bump will fix it

mikehardy:
Nope, it's a persistent failure. Works locally though. I don't have time to troubleshoot at the moment as my main goal is to get react-native-firebase updated, so I'll have to leave it. But certainly strange why boost would fail checksum in CI this way

```
Installing boost (1.76.0)
[!] Error installing boost
Verification checksum was incorrect, expected f0397ba6e982c4450f27bf32a2a83292aba035b827a5623a14636ea583318c41, got 79e6d3f986444e5a80afbeccdaf2d1c1cf964baa8d766d20859d6[53](https://github.com/react-native-device-info/react-native-device-info/actions/runs/13121827949/job/36609380705?pr=1691#step:10:54)a16c39848
error Command failed with exit code 1.
```

I'm going to merge since I changed android code and that works in CI plus on API23 and API24 emulator (I checked!)

PR #1692, opened 2025-02-03
fix(android): quiet a build warning for users

### LINKED ISSUE #1689: onCatalystInstanceDestroy is deprecated (RNDeviceModule.java)
In React Native 0.77, the onCatalystInstanceDestroy method in NativeModule is deprecated and marked for removal. This causes a deprecation warning when building with react-native-device-info.

**Temporary Workaround:**
I suppressed the warning with:

```java
@Override
@SuppressWarnings("deprecation")
    invalidate();
}
```

**Expected Fix:**
The library should replace onCatalystInstanceDestroy with the recommended lifecycle method, such as ReactLifecycleEventListener, to ensure compatibility.

mikehardy:
## Description

we already handle the onCatalystInstanceDestroy method deprecation pretty well, but we get build warnings for even keeping the old API, though we do that on purpose for backwards-compatibility

So suppress those build warnings since they look worrisome but are not actionable

There was one poster that mentioned it, and I noticed it as well in my own usage of the module:

- Fixes #1689

## Tests

I checked compiling this against a really old react-native (the example here being so out of date actually helped! CI will show it as well) and I tested it against a modern react-native where the warning was generated. Both still worked, so I think this is backwards- and forwards-compatible

## Checklist

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -197,6 +197,7 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
   // the upstream method was removed in react-native 0.74
   // this stub remains for backwards compatibility so that react-native < 0.74
   // (which will still call onCatalystInstanceDestroy) will continue to function
+  @SuppressWarnings({"deprecation", "removal"})
   public void onCatalystInstanceDestroy() {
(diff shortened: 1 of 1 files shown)


mikehardy:
I have checked modern react-native and CI just went green here for old react-native android compile (which I also checked, but this is objective evidence...)

So I'm going to merge this - it's certainly small and non-functional, just needs to not break anyone...

PR #1693, opened 2025-02-04
Replace deprecated onCatalystInstanceDestroy with lifecycle methods

### LINKED ISSUE #1687: Windows Issue: Build errors for Windows New Arch
I've tried creating a fresh new React Native project, initialising it for windows with the new cpp-app template and installing react-native-device-info. The project then fails to build with:

` Building Solution: XamlLibMetadataProvider.g.cpp...
 ✖ Build failed with message C:\Users\joshh\coding_bootcamp\projects\zonal\deviceInfo\node_modules\react-native\ReactCommon\react\runtime\ReactInstance.cpp(205,21): error C2511: 'void facebook::react::ReactInstance::loadScript(std::unique_ptr<const facebook::react::JSBigString,std::default_delete<const facebook::react::JSBigString>>,const std::string &)': overloaded member function not found in 'facebook::react::ReactInstance' [C:\Users\joshh\coding_bootcamp\projects\zonal\deviceInfo\node_modules\react-native-windows\Microsoft.ReactNative\Microsoft.ReactNative.vcxproj]. Check your build configuration.`

Is the new arch on Windows something that is supported? Or will be supported in the future?

### LINKED ISSUE #1689: onCatalystInstanceDestroy is deprecated (RNDeviceModule.java)

seco35:
## Description

Fixes #1689

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    Android-Only   |
| Android |    ✅     |
| Windows |    Android-Only    |

## Checklist

* [Y ] I have tested this on a device/simulator for each compatible OS
* [ N] I added the documentation in `README.md`
* [ N] I updated the typings files (`privateTypes.ts`, `types.ts`)
* [ N] I added a sample use of the API (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -38,6 +38,7 @@ import android.hardware.camera2.CameraManager;
 import androidx.annotation.Nullable;
 
 import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
@@ -72,7 +73,7 @@ import static android.os.BatteryManager.BATTERY_STATUS_FULL;
 import static android.provider.Settings.Secure.getString;
 
 @ReactModule(name = RNDeviceModule.NAME)
-public class RNDeviceModule extends ReactContextBaseJavaModule {
+public class RNDeviceModule extends ReactContextBaseJavaModule implements LifecycleEventListener 
   public static final String NAME = "RNDeviceInfo";
   private final DeviceTypeResolver deviceTypeResolver;
   private final DeviceIdResolver deviceIdResolver;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
```diff
@@ -194,12 +195,21 @@ public void onReceive(Context context, Intent intent) {
...
+  public void onHostResume() {
+    // Do something?
+  }
+
+  @Override
+  public void onHostPause() {
+    // Do something?
+  }
+  // Will fix the onCataystInstanceDestroy warning / build and work for react-native >= 0.74
+  // Tested on react-native 0.77.0
```
mikehardy, line 208: 🤔 so you have added something that will work on react-native >= 0.74, but how will it go for people on older versions? Are these symbols for the lifecycle listener and it's required methods available there? More specifically: what react-native version *exactly* were they first shipped in?

seco35: I was referring to RN 0.74.0+ since that’s the last officially supported version. What’s the target range for react-native-device-info? I always assumed libraries should aim for the last officially supported version and are good to go, right? 🤔

Looking back, I was quite sure this was supported in 0.66 since that’s the version I upgraded from when moving to 0.71. So don’t mind the comment in the code —I just meant to reference the latest officially supported version was enough.

After some deep digging, I found that LifecycleEventListener, along with the functions I use, was actually introduced all the way back in RN v0.18.0 (yep, nearly a decade ago!). Safe to say, anyone using react-native-device-info will probably be on a slightly newer version. 😄

So I think the PR can be accepted to fix the issue. 🚀

LifecycleEventListener.java (RN 0.18.0):
[](https://github.com/facebook/react-native/blob/v0.18.0/ReactAndroid/src/main/java/com/facebook/react/bridge/LifecycleEventListener.java)

seco35:
@mikehardy Will you have time to merge this PR soon? It was closed by a bot without a proper fix addressing the issue (unless there’s another PR I haven’t seen).

Issue #1722, opened 2025-08-28
`inputMethodInfo.getServiceName()` can be null, resulting in a `NullPointerException`


andy-j:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-native-device-info@14.0.4` for the project I'm working on.

As it turns out, `inputMethodInfo.getServiceName()` (called in the `hasKeyboard` function) is not guaranteed to be non-null. In the case that it _is_ `null`, we get a `NullPointerException`.

Since we're calling this to determine whether it contains a substring, we can simply fall back to `""` and the comparison will be `false` for this service name / id.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java b/node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
index 23e8fb4..768a713 100644
--- a/node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
+++ b/node_modules/react-native-device-info/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java
@@ -1113,8 +1113,8 @@ public class RNDeviceModule extends ReactContextBaseJavaModule {
     List<InputMethodInfo> inputMethodList = this.inputMethodManager.getEnabledInputMethodList();
     if (inputMethodList != null && !inputMethodList.isEmpty()) {
       for (InputMethodInfo inputMethodInfo : inputMethodList) {
-        String serviceName = inputMethodInfo.getServiceName().toLowerCase();
-        String id = inputMethodInfo.getId().toLowerCase();
+        String serviceName = inputMethodInfo.getServiceName() != null ? inputMethodInfo.getServiceName().toLowerCase() : "";
+        String id = inputMethodInfo.getId() != null ? inputMethodInfo.getId().toLowerCase() : "";
         if (serviceName.contains(name.toLowerCase()) || id.contains(name.toLowerCase())) {
           return true;
         }
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #1723, opened 2025-08-28
fix(android): Resolve NullPointerException in hasKeyboard function

andy-j:
## Description

Change `hasKeyboard` to use empty strings for comparison when `getServiceName()` or `getId()` are null. In the case that `inputMethodInfo` has no service name or id, `hasKeyboard` throws a `NullPointerException`.

Fixes react-native-device-info/react-native-device-info#1722
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist


andy-j:
Re: testing on a device/simulator—I'm not certain how to reproduce this. We've seen it exclusively on Android 13 & 14 devices in the wild.

mikehardy (approved):
seems reasonable, thanks

PR #1730, opened 2025-10-14
feat: add appsetid for android

walterholohan:
## Description

Added `getAppSetId()` that allows you to get the Android AppSetId (A unique, per-device, per developer-account user-resettable ID for non-monetizing advertising usecases.). Full docs here http://developer.android.com/design-for-safety/privacy-sandbox/reference/adservices/appsetid/AppSetId

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -25,7 +25,6 @@ import {
   getUniqueId,
   getUniqueIdSync,
   getAppSetId,
-  getAppSetIdSync,
   useBatteryLevel,
   useBatteryLevelIsLow,
   usePowerState,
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
This seems like a good feature thanks! Wasn't even aware it existed.
Between the reflection and the generic exception catch, I can't see how the implementation here can do much harm - it shouldn't ever crash in other words

I'm going to rebase it on to our current set of commits and if CI passes, this will be a test case of our new publishing system

PR #1735, opened 2025-11-12
build: replace semantic-release with release-please

schie:
## Description
release-please is simpler to maintain (next to nothing), so migrating over is good choice.

This _does_ cause a slight workflow change. all pushes into master will trigger the new workflow. Ghe workflow will create / update a PR with the version bump. Once that PR is merged in, the workflow will run again, but instead of a PR, the release notes, tags, etc.. will be created, then a build will me made and pushed to NPM.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    N/A     |
| Windows |    N/A     |

## Checklist


mikehardy:
Irony - I've just switched my simpler repos to release-it in order to publish them from workflows and do the oidc thing. But as long as its easy sure. I'll take a look

schie:
Alright, should be good now

PR #1736, opened 2025-11-12
refactor: use release-it for publish / update CI workflow deps

mikehardy:
@schie what do you think of this one?

This is the same config that I've used in packages `@invertase/react-native-apple-authentication` and `nodejs-google-java-format`, and I was probably going to use it for react-native-netinfo and notifee and jet as well because it's just so simple and I've got it working in 2 spots already...

Whether this is good or you truly prefer a PR for releases + release-please like in #1735 - I've configured release.yml to be trusted on npmjs.org already, so that step is done - just need to merge one of these and hit the release workflow (or...I guess since it triggers on push, it'll go!)

Screenshot 2025-11-12 at 1 06 59 PM

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -56,7 +56,7 @@ jobs:
         id: yarn-cache
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/package.json') }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('package.json') }}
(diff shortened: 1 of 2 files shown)


mikehardy:
yeah - on the windows side CI is not going to upgrade cleanly to 2022, so I'm leaving it at the disabled 2019 runner which just cancels the action

I think my last push has the current example app building at least in CI

moving to react-native-test-app is the thing to do though, for sure. And then probably redoing the core, but new arch style

I don't have the appetite for that though, alas. I can likely do the RNTestApp change in a while though, I'm getting pretty good at driving that package

PR #1738, opened 2025-11-12
feat(android)!: getAppSetId - a specific type of android id / requires compileSdk 34+

mikehardy:
## Description

This comes from and adds to the work from @walterholohan on #1730 

I took the liberty of making it a semver major / breaking change by requiring compileSdk 34, that isn't a big ask for people these days (since we already have to *target* it with app updates anyway 🤷 ) and then we don't have to do all the reflection stuff - which wasn't working since the method isn't a GMS Task, it's a special "OutcomeReceiver" and needs an Executor too.

- Obsoletes #1730 

It is async only
It is android only
It does not seem to work on emulator (`Service is not available`) but the call does go through at least

Needs a squash merge once CI is done chewing on it

Added `getAppSetId()` that allows ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |
| Windows |    ❌     |

## Checklist

### DIFF
--- a/__tests__/getters.test.ts
+++ b/__tests__/getters.test.ts
@@ -759,23 +759,17 @@ describe('array getters', () => {
 
 describe('Object Getters', () => {
   describe('getAppSetId*', () => {
-    const [, asyncGetter, syncGetter, asyncNativeGetter, syncNativeGetter] = makeTable(
-      'getAppSetId'
-    );
+    const asyncGetter = RNDeviceInfo.getAppSetId;
+    const asyncNativeGetter = mockNativeModule.getAppSetId;
     const supportedPlatforms = ['android'];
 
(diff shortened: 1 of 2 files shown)


mikehardy:
My main goal is actually to test the release system from #1736 - but I needed a feature to release

mikehardy:
@schie if this looks good to you, could be the first release-it release? I'll do a squash merge with the appropriate `BREAKING CHANGES: requires compileSdk 34+` block in it and...should work ?

schie (approved):

schie:
@mikehardy 
deployment seems to have partially failed

Looks like v15.0.0 is in NPM, but github tag / release wasn't created.

PR #1798, opened 2026-08-26
fix(android): guard emulator keyboard probe

### LINKED ISSUE #1797: isEmulatorSync() crashes on Android 15/16 with compileSdk 36 — getEnabledInputMethodList() returns InputMethodInfoSafeList
### Summary

`isEmulatorSync()` throws on Android 15/16 devices when the library is built with `compileSdk 36`. Because it is declared `isBlockingSynchronousMethod = true`, the exception propagates synchronously into JS:

```
Exception in HostFunction: Couldn't convert result of type java.util.ArrayList
to com.android.internal.inputmethod.InputMethodInfoSafeList
```

The failure is in the `hasKeyboard("memuime")` branch, which calls `InputMethodManager.getEnabledInputMethodList()`. API 36 changed the declared return type of that method to `InputMethodInfoSafeList`, and the bridge conversion to `List<InputMethodInfo>` fails at runtime.

This affects **real devices, not just emulators** — see [Why this always fires](#why-this-always-fires-on-real-hardware) below.

### Environment

- `react-native-device-info` 15.0.2 (current latest; `master` is unchanged at the time of writing)
- React Native 0.86.2, New Architecture
- **`compileSdk = 36`** ← required to reproduce
- Physical hardware, Android 15, arm64 (reproduced on a Samsung foldable)

`compileSdk` is the key variable. `getEnabledInputMethodList()` resolves to a different return type depending on the SDK the library is *compiled* against, so a project on `compileSdk 35` will not see this even when running on the same Android 15/16 device. That is likely why this has gone unreported so far — it is gated on build configuration, not on device population. As projects move to `compileSdk 36` for Play's target-API requirements, the affected share will grow.

Note that the library picks this up automatically: `android/build.gradle` line 25 reads

```groovy
compileSdkVersion safeExtGet('compileSdkVersion', 34)
```

so any app that sets `compileSdkVersion = 36` in its root project compiles this module at 36 too. No unusual Gradle setup is involved.

### Why this always fires on real hardware

[`RNDeviceModule.java:320-321`](https://github.com/react-native-device-info/react-native-device-info/blob/master/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L320-L321):

```java
        || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic")
        || this.hasKeyboard("memuime"));
```

The grouping here parses as `(A && B) || C`, so `hasKeyboard(...)` is **not** guarded by the `Build.BRAND`/`Build.DEVICE` test. The indentation suggests the intent was for all three to form one emulator heuristic, but as written the inner `||` escapes that grouping.

The practical consequence: on a real device every preceding `Build.*` check is false, so the `||` chain never short-circuits and `hasKeyboard()` is always reached. Every `compileSdk 36` build calling `isEmulatorSync()` on Android 15/16 throws, regardless of whether it is an emulator.

[`RNDeviceModule.java:1115`](https://github.com/react-native-device-info/react-native-device-info/blob/master/android/src/main/java/com/learnium/RNDeviceInfo/RNDeviceModule.java#L1115) then calls the enumeration unguarded:

```java
private boolean hasKeyboard(String name) {
  List<InputMethodInfo> inputMethodList = this.inputMethodManager.getEnabledInputMethodList();
  ...
```

### Impact

Because `isEmulatorSync()` is a blocking synchronous method, callers cannot always defend against it. If the call sits in a module-level constant initialiser, it throws during `import` evaluation — before any caller's `try`/`catch` has been established — and takes down cold launch. In our case two separate layers of app-side error handling around the calling function were both unreachable for that reason.

### Suggested fix

Two independent changes, either of which stops the crash; together they also fix the latent logic bug.

**1. Guard the enumeration.** It is a best-effort heuristic for one emulator brand, so failing closed is safe:

```diff
 private boolean hasKeyboard(String name) {
-    List<InputMethodInfo> inputMethodList = this.inputMethodManager.getEnabledInputMethodList();
-    if (inputMethodList != null && !inputMethodList.isEmpty()) {
-      for (InputMethodInfo inputMethodInfo : inputMethodList) {
-        String serviceName = inputMethodInfo.getServiceName() != null ? inputMethodInfo.getServiceName().toLowerCase() : "";
-        String id = inputMethodInfo.getId() != null ? inputMethodInfo.getId().toLowerCase() : "";
-        if (serviceName.contains(name.toLowerCase()) || id.contains(name.toLowerCase())) {
-          return true;
+    try {
+      List<InputMethodInfo> inputMethodList = this.inputMethodManager.getEnabledInputMethodList();
+      if (inputMethodList != null && !inputMethodList.isEmpty()) {
+        for (InputMethodInfo inputMethodInfo : inputMethodList) {
+          String serviceName = inputMethodInfo.getServiceName() != null ? inputMethodInfo.getServiceName().toLowerCase() : "";
+          String id = inputMethodInfo.getId() != null ? inputMethodInfo.getId().toLowerCase() : "";
+          if (serviceName.contains(name.toLowerCase()) || id.contains(name.toLowerCase())) {
+            return true;
+          }
         }
       }
+    } catch (Throwable t) {
+      // Best-effort emulator heuristic — a platform change here must not take down the caller.
+      return false;
     }
     return false;
   }
```

**2. Fix the parenthesisation** so the input-method probe only runs for candidates that already look like an emulator:

```diff
-        || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic")
-        || this.hasKeyboard("memuime"));
+        || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic"))
+        || this.hasKeyboard("memuime");
```

This changes behaviour only for devices where exactly one of `BRAND`/`DEVICE` starts with `generic` — but it is worth doing regardless, since the current form reads as a mistake.

Worth considering beyond this specific fix: any future platform change to a `Build.*` or system-service call in the `isEmulatorSync()` chain will surface the same way — synchronously, potentially during module evaluation. Hardening the method as a whole would prevent the next instance of this.

Related: #1722 was a `NullPointerException` in this same loop, fixed by null-checking inside the iteration rather than guarding the call itself.

### Workaround

For anyone hitting this before a fix lands: wrap the call so a failing probe degrades to a fallback instead of throwing, and make sure it is not invoked from a module-level initialiser. That stops the crash, at the cost of `isEmulator` reading permanently `false` on affected devices.

Happy to send the above as a PR if that is useful.

huytdps13400:
## Description

Fixes #1797.

`isEmulatorSync()` treats the enabled-input-method scan as one best-effort emulator signal. Android framework failures from `getEnabledInputMethodList()` now fail closed to `false` instead of escaping synchronously through the React Native HostFunction and crashing real devices. The guard covers both `RuntimeException` failures such as a cast mismatch and `LinkageError` failures such as `NoSuchMethodError` from an API/runtime signature mismatch.

The successful input-method matching path and all other emulator heuristics remain unchanged. The probe is exposed package-private only for focused Android unit coverage; the public API is unchanged.

## Compatibility

| OS | Implemented |
| --- | :---: |
| iOS | N/A |
| Android | ✅ |
| Windows | N/A |

## Test plan

- focused Android regressions: 2/2 passed (`RuntimeException` and `NoSuchMethodError`)
- Android Java source and tests compile against compileSdk 34
- `yarn test --runInBand` — 434/434 tests passed
- TypeScript build check passed
- Babel CommonJS and module builds passed
- `git diff --check`

The repository declares JUnit Jupiter tests but does not configure a Jupiter engine for Gradle, so focused native verification uses a task-local Java test harness over the repository source and test files. The complete legacy Android test set also exposes three pre-existing `DeviceIdResolverTest` failures under Java 21/Mockito 3.6; both keyboard-probe regressions are green.

No dependency, lockfile, typing, documentation, or generated-output changes.

## Checklist

- [ ] Tested on a device/simulator — not available locally; the framework exception boundary is covered by focused Android tests
- [x] Documentation not required; no public API change
- [x] Typings not required; no public API change
- [x] Example change not required; no new API or user-facing option


huytdps13400:
CI status: the only workflow, `PR Title is Conventional`, was cancelled after about 15 minutes without starting any steps. The run is attached to the current head `18d5c546`, so this is a runner/queue cancellation rather than a title or code failure. I do not have upstream Actions permissions to rerun it; could a maintainer please rerun the cancelled workflow when convenient?

waveo-wangxiao:
Thanks for picking this up. Narrowing the `try` to just the `getEnabledInputMethodList()` call is better than what I suggested in the issue — my version wrapped the whole method body, which would also have swallowed genuine bugs in the matching loop. Keeping the loop outside the guard is the right call.

One concern about the catch clause, though.

### `catch (RuntimeException)` may not catch the actual failure

The reported failure is a return-type mismatch at the JNI/bridge boundary:

```
Couldn't convert result of type java.util.ArrayList
to com.android.internal.inputmethod.InputMethodInfoSafeList
```

When the compile-time signature (API 36's `InputMethodInfoSafeList`) does not match the method the runtime actually resolves, ART typically raises a `LinkageError` subclass — `NoSuchMethodError` or `IncompatibleClassChangeError` — rather than a `RuntimeException`. Those sit on a separate branch of the hierarchy:

```
Throwable
├── Error
│   └── LinkageError
│       └── IncompatibleClassChangeError
│           └── NoSuchMethodError          ← not caught by catch (RuntimeException)
└── Exception
    └── RuntimeException
        └── ClassCastException             ← caught
```

If the failure surfaces as `ClassCastException` the current guard works. If it surfaces as `NoSuchMethodError`, the exception escapes `hasKeyboard()`, propagates through the synchronous `isEmulatorSync()` HostFunction, and the crash is unchanged.

The test doesn't distinguish these cases, since `doThrow(new RuntimeException(...))` asserts against a hand-constructed exception rather than the real failure mode. It will pass either way.

### Suggestion

Widening the catch would cover both, and stays consistent with the "best-effort signal" framing already in the comment:

```diff
-    } catch (RuntimeException ignored) {
+    } catch (RuntimeException | LinkageError ignored) {
```

`catch (Throwable)` would also work. Either is defensible here — the probe is one heuristic for a single emulator brand, so failing closed costs nothing, while an escaping throwable takes down the process.

It'd also be worth adding a second test case that throws `NoSuchMethodError`, so the regression covers the linkage path and not only the runtime-exception path.

### Separately: the parenthesisation

Point 2 from the issue is still open:

```java
        || (Build.BRAND.startsWith("generic") && Build.DEVICE.startsWith("generic")
        || this.hasKeyboard("memuime"));
```

This parses as `(A && B) || C`, so `hasKeyboard()` still runs on every real device rather than only on emulator candidates. Not a blocker for this PR — with the guard in place it fails closed quietly — but the probe is still doing needless work on every call, and `isEmulator` stays `false` on affected devices instead of being correctly determined. Reasonable to leave for a follow-up if you'd rather keep this PR minimal.

I can't verify the linkage-vs-runtime question on hardware myself either, which is exactly why widening the catch seems worth doing: it makes the fix correct without needing to know which of the two the platform actually throws.

huytdps13400:
Thanks — you are right about the throwable hierarchy. Updated in `d454489`:

- the best-effort probe now catches `RuntimeException | LinkageError`, covering both cast-style failures and ART linkage failures such as `NoSuchMethodError`
- added a dedicated `NoSuchMethodError` regression alongside the existing `RuntimeException` case
- kept the matching loop outside the guard, so unrelated loop defects are still visible

Both focused Android regressions pass (2/2). The actual Android source/test files compile in the task-local harness, all 434 Jest tests pass, TypeScript passes, Babel CommonJS/module builds pass, and `git diff --check` is clean.

I left the parenthesization cleanup for a separate follow-up as suggested, so this PR remains focused on preventing the synchronous framework-boundary crash.

